PR #6, opened 2019-03-30
Add missing import statement to getting started guide

BalestraPatrick:
`Label` is defined in `BlueprintUICommonControls`, so it should be imported too along with `BlueprintUI`.


timdonnelly:
Thanks!

PR #14, opened 2019-07-02
Add font and configuration block to textfield

dostrander:
* Adds the font property to the textfield for font customization
* Adds a configuration block similar to how `Label` takes one on `init`


dostrander:
@davidapgar @timdonnelly @bencochran

PR #15, opened 2019-07-31
Add Swift 5 support

watt:
There were no required code changes for this, it's just a podspec and Xcode project update to signal compatibility.

### DIFF
--- a/BlueprintUI.podspec
+++ b/BlueprintUI.podspec
@@ -7,7 +7,7 @@ Pod::Spec.new do |s|
   s.author       = 'Square'
   s.source       = { :git => 'https://github.com/square/blueprint.git', :tag => s.version }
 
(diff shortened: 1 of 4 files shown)


BlueprintUI.podspec
```diff
@@ -7,7 +7,7 @@ Pod::Spec.new do |s|
   s.author       = 'Square'
   s.source       = { :git => 'https://github.com/square/blueprint.git', :tag => s.version }
 
-  s.swift_version = '4.2'
+  s.swift_versions = ['4.2', '5.0']
```
davidapgar: Up to you, but I think it'd be fine to make this only be `5.0` instead of both versions (we did that recently with workflow, and since swift 4.2 can build with swift 5.0 pods, it probably better to just switch over)

PR #16, opened 2019-08-25
[WIP] Add 'none' to possible stack underflows

kyleve:
There's currently no easy way to have stacks that are sized by their container (versus sized by their content) leave all their space at the end of the stack – adding this underflow value makes that doable without needing a growing spacer, or without needing a custom list layout.

- [ ] Figure out the right name
- [ ] Add tests
- [ ] Do we want more customizability? Eg, able to also specify if the extra space should be left at the top, bottom, or if the content should be centered.

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -289,7 +289,7 @@ extension StackLayout {
 
         switch underflow {
         case .none:
-            space = 0.0
+            space = minimumSpacing
         case .growProportionally:
             space = minimumSpacing
         case .growUniformly:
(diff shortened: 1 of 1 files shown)


BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -137,6 +137,9 @@ extension StackLayout {
 
     /// Determines the on-axis layout when there is extra free space available.
     public enum UnderflowDistribution {
+        
+        /// Additional space will appear after all items are laid out (bottom for vertical stacks, left for horizontal stacks).
```
kyleve: Should be right for horizontal stacks.

kyleve:
Naren is working on a fuller version of this, closing out.

PR #18, opened 2019-10-08
Screen scale rounding

watt:
This PR changes the rounding behavior when measuring text-containing elements to round up to the nearest screen pixel instead of the nearest point. This is accomplished through a new `roundingScale` property on affected elements, which defaults to `UIScreen.main.scale`. It can be set explicitly to other values but is not an initializer argument (since it will rarely be set outside of testing).

For tests that measure affected elements the scale is explicitly set to 1 to match current behavior. I also added one new test to make sure the rounding works on a known font/text combination.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -5,6 +5,8 @@ public struct AttributedLabel: Element {
 
     public var attributedText: NSAttributedString
     public var numberOfLines: Int = 0
+    /// The scale to which pixel measurements will be rounded. Defaults to `UIScreen.main.scale`.
+    public var roundingScale: CGFloat = UIScreen.main.scale
 
     public init(attributedText: NSAttributedString) {
         self.attributedText = attributedText
@@ -14,6 +16,7 @@ public struct AttributedLabel: Element {
         struct Measurer: Measurable {
 
             var attributedText: NSAttributedString
+            var roundingScale: CGFloat
 
             func measure(in constraint: SizeConstraint) -> CGSize {
                 var size = attributedText.boundingRect(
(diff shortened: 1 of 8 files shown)


BlueprintUICommonControls/Sources/FloatingPoint+ScaleRounding.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension FloatingPoint {
+    mutating func round(_ rule: FloatingPointRoundingRule, by scale: Self) {
+        self = self.rounded(rule, by: scale)
+    }
+
+    func rounded(_ rule: FloatingPointRoundingRule, by scale: Self) -> Self {
+        return (self * scale).rounded(rule) / scale
+    }
```
bencochran: Maybe some unit tests for these?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -5,6 +5,8 @@ public struct AttributedLabel: Element {
 
     public var attributedText: NSAttributedString
     public var numberOfLines: Int = 0
+    /// The scale to which pixel measurements will be rounded. Defaults to `UIScreen.main.scale`.
+    public var roundingScale: CGFloat = UIScreen.main.scale
```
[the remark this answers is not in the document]
watt: Maybe... even if you were transforming this element, I'd think you'd want the measured size to match the intrinsic size of the underlying UILabel, which is what this behavior does? I'm going to leave it for now, and perhaps we can add that later if someone with a specific need requests it.

BlueprintUICommonControls/Sources/FloatingPoint+ScaleRounding.swift
watt: Done!

PR #22, opened 2019-10-10
Rule element

watt:
A new `Rule` element that is a solid horizontal or vertical line. You specify the orientation, the color, and the thickness.

```swift
Rule(orientation: .horizontal, color: .black, thickness: .hairline)
```

The thickness is an enum with a special value for "hairline" (1 screen pixel). This is a little more magical than I'd like, but it makes hairline thickness extremely discoverable as an autocomplete suggestion.

Considered adding a static var to `CGFloat` instead, but that seems like namespace pollution. Open to alternative ideas of how to represent hairline width.


BlueprintUICommonControls/Sources/Rule.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// Represents whether the rule is parallel to the x or y axis.
+    public enum Orientation {
+        /// Indicates that the rule is parallel to the x axis.
+        case horizontal
+        /// Indicates that the rule is parallel to the y axis.
+        case vertical
+    }
+
+    /// Represents the thickness of a rule in the direction perpendicular to its orientation.
+    public enum Thickness {
```
AquaGeek: Should we conform this to `ExpressibleByIntegerLiteral` and `ExpressibleByFloatLiteral`? Then consumers could write:

```swift
let rule = Rule(orientation: .horizontal, color: .red, thickness: 2)
```

Doing so loses some specificity (i.e. is it 2 points or pixels?), but it might be worth that trade-off.

AquaGeek:
I agree that we probably don't want to pollute `CGFloat` with `.hairline`. Allowing the enum to be instantiated with number literals is a potential workaround.

BlueprintUICommonControls/Sources/Rule.swift
```diff
@@ -0,0 +1,77 @@
...
+        /// Indicates that the rule should be exactly 1 screen pixel thick,
+        /// the thinnest possible line that can be drawn.
+        case hairline
+        /// An exact thickness in points.
+        case points(CGFloat)
+
+        var value: CGFloat {
+            switch self {
+            case .hairline:
+                return 1.0 / UIScreen.main.scale
```
watt: My concern with magic behavior is mostly around this static reference to `UIScreen.main.scale`. Ideally we could set this explicitly but have a convenient default. See the way I implemented screen scale rounding in #18 for context.

One option is to add a screen scale property to `Rule`, but it doesn't make sense in cases where the thickness is a specific value.

Another option is to rewrite Thickness like this:
```swift
    public struct Thickness {
        public static func hairline(scale: CGFloat = UIScreen.main.scale) -> Thickness {
            return Thickness(value: 1.0 / scale)
        }

        public static func points(_ value: CGFloat) -> Thickness {
            return Thickness(value: value)
        }

        var value: CGFloat

        private init(value: CGFloat) {
            self.value = value
        }
    }
```

That makes it require parens at the point of use:

```swift
Rule(orientation: .horizontal, color: .black, thickness: .hairline())
```

But the autocomplete is pretty good and makes the default reasonably discoverable.

Screen Shot 2019-10-10 at 4 27 46 PM

Maybe that's fine?

BlueprintUICommonControls/Sources/Rule.swift
watt: This is an alright convenience, but I don't think it removes any of the magic from `hairline`.

kyleve: Fwiw, I prefer the enum route, so it's easy for us to find where custom sizes are used.

BlueprintUICommonControls/Sources/Rule.swift
timdonnelly: At the time when a developer is building a UI from elements they really should not have to worry about screen scale. If this information is required somewhere in the layout pass, it would be worth exploring how we expose this information (the current device/environment context) to the element implementation.

In the meantime, I'm ok using `UIScreen.main` as it will always be correct in practice.

BlueprintUICommonControls/Sources/Rule.swift
timdonnelly: I expect that `.hairline` will be the most commonly used option. If we want to improve the ergonomics of a custom line width I would try adding a convenience initializer before adding an `EspressibleByIntegerLiteral` conformance.

1. It makes it really easy to find code that calls that initializer in the future.
2. I'm very hesitant to add `ExpressibleBy[Foo]Literal` conformance to types that are not... literally a `Foo` (an integer/number in this case); it just makes for non-obvious code that someone/you will have to dig into in the future. In practice I've always ended up removing that stuff down the road for the sake of clarity.

PR #23, opened 2019-10-10
ConstrainedAspectRatio element

watt:
A new `ConstrainedAspectRatio` element that constrains the size of its child to — you guessed it — an aspect ratio.

You can choose to enforce this constraint by expanding or shrinking the child. Open to alternative suggestions for the name of "expand or shrink" behavior.

### DIFF
--- a/BlueprintUI/Sources/Layout/AspectRatio.swift
+++ b/BlueprintUI/Sources/Layout/AspectRatio.swift
@@ -1,0 +1,29 @@
+/// Represents an a proportional relationship between width and height.
+public struct AspectRatio {
+    /// A 1:1 aspect ratio.
+    public static let square = AspectRatio(x: 1, y: 1)
+
+    /// The horizontal component of this ratio.
+    public var x: CGFloat
+    /// The vertical component of this ratio.
+    public var y: CGFloat
+
+    /// Initializes with a horizontal and vertical components.
+    ///
+    /// - Parameter x: The horizontal comonent.
+    /// - Parameter y: The vertical component.
+    public init(x: CGFloat, y: CGFloat) {
+        self.x = x
+        self.y = y
+    }
+
+    func height(forWidth width: CGFloat) -> CGFloat {
+        // TODO: round to screen scale when that lands
+        return (width * y / x).rounded()
+    }
+
+    func width(forHeight height: CGFloat) -> CGFloat {
+        // TODO: round to screen scale when that lands
+        return (height * x / y).rounded()
+    }
+}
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/AspectRatio.swift
```diff
@@ -0,0 +1,29 @@
+/// Represents an a proportional relationship between width and height.
+public struct AspectRatio {
+    /// A 1:1 aspect ratio.
+    public static let square = AspectRatio(x: 1, y: 1)
+
+    /// The horizontal component of this ratio.
+    public var x: CGFloat
```
timdonnelly: The property names `x`/`y` imply a position in a coordinate system – if this structure persists I would almost certainly rename them to `width`/`height`.

Stepping back, I find it surprising that we are storing two values here. Can we simply store the aspect ratio (derived from `width/height`) as the backing storage for this type? Convenience initializers that take `width`/`height` or a `CGSize` seem useful, but it seems as though this type should encode the most fundamental representation of an aspect ratio as its storage: a single `Double` value.

Thoughts?

timdonnelly: For API reference, SwiftUI has the following:

```swift
extension View {
    public func aspectRatio(_ aspectRatio: CGFloat? = nil, contentMode: ContentMode) -> some View
    public func aspectRatio(_ aspectRatio: CGSize, contentMode: ContentMode) -> some View
}

public enum ContentMode {
    case fit
    case fill
}
```

watt: Agree `width`/`height` is better, I kind of punted on that (I did check Wikipedia to see if they had proper names first, but they don't seem to).

Here's an argument for storing two values:
- You get better diagnostics. `AspectRatio(width: 16, height: 9)` can have a debug string of `16:9` instead of `1.7777777777778`.
- A bit more resilient to transforms. If you have an existing ratio and you want to double its height, `aspectRatio.height *= 2` doesn't lose precision to normalization.

I recognize that's a pretty weak argument, and I'd rather match SwiftUI, so I'll change this.

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -0,0 +1,77 @@
+/// Constrains the size of the content element to an aspect ratio.
+public struct ConstrainedAspectRatio: Element {
```
kyleve: Non-blocking: Don't love this name but don't have a better suggestion right now.

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// Represents whether the content element's size should be expanded to fill its parent
+    /// or shrunk to fit it.
+    public enum ContentMode {
```
kyleve: Doesnt matter within Blueprint, but I tend to make these enums equatable when I can.

PR #26, opened 2019-10-23
Add EqualStack element

connorcimowsky:
stack


BlueprintUI/Sources/Layout/EqualSizeStack.swift
```diff
@@ -0,0 +1,134 @@
...
+                    height: (size.height - (gutter * CGFloat(items.count - 1))) / CGFloat(items.count))
+            }
+
+            return Array(repeating: LayoutAttributes(size: baseSize), count: items.count)
+                .enumerated()
+                .map { index, attributes in
+                    var attributes = attributes
+                    switch direction {
+                    case .horizontal:
+                        attributes.frame.origin.x = (baseSize.width + gutter) * CGFloat(index)
```
watt: How do you feel about rounding this value to the nearest pixel?

BlueprintUI/Sources/Layout/EqualSizeStack.swift
```diff
@@ -0,0 +1,134 @@
+/// An element that sizes all of its children equally, stacking them according to
+/// the specified `direction` and spacing them according to the specified `gutter`.
```
watt: It looks like the cross-axis sizing behavior is to justify all children to the full width/height of the stack. Can you add that to the docs?

timdonnelly: I would also mention the measuring behavior: even though it sizes all children equally, its measuring implementation returns the sum of all child sizes along the primary axis.

I'm actually not sure that the current measuring behavior is particularly helpful, as the layout implementation will not take the child sizes into account when assigning layout attributes. I think we may want this implementation to simply take up all available space? I'd have to play with this to form a stronger opinion, which I have not done.

BlueprintUI/Sources/Layout/EqualSizeStack.swift
```diff
@@ -0,0 +1,134 @@
...
+public struct EqualSizeStack: Element {
+
+    /// The direction in which this element will stack its children.
+    public var direction: Direction
+
+    /// The amount of space between children in this element.
+    public var gutter: CGFloat
```
timdonnelly: For consistency with `Stack`, I would call this `spacing` (`Stack` uses `minimumSpacing` as it can expand the spacing if needed)

BlueprintUI/Sources/Layout/EqualSizeStack.swift
```diff
@@ -0,0 +1,134 @@
...
+    /// - Parameters:
+    ///     - direction: The direction in which this element will stack its children.
+    ///     - gutter: The amount of space between children in this element. Defaults to 0.
+    ///     - children: The child elements to be laid out. Defaults to an empty array.
+    ///     - configure: A closure allowing the element to be further customized. Defaults to a no-op.
+    public init(
+        direction: Direction,
+        gutter: CGFloat = 0,
+        children: [Element] = [],
+        configure: (inout EqualSizeStack) -> Void = { _ in })
```
kyleve: Is the configure block useful if the initializer already provides all values to be customized?

BlueprintUI/Sources/Layout/EqualSizeStack.swift
kyleve: +1

kyleve: Same for all the things in here, likely.

kyleve:
Would EqualStack also work from a naming perspective?

BlueprintUI/Sources/Layout/EqualSizeStack.swift
connorcimowsky: not really; can remove.

BlueprintUI/Sources/Layout/EqualSizeStack.swift
bencochran: I could also see an argument that it could take the size of the largest child and multiply that by the number of children, since that would be the smallest size that would layout to fit all children. But measuring the be the maximum works too.

BlueprintUI/Sources/Layout/EqualSizeStack.swift
bencochran: It’s useful for adding children, just like how `Row` and `Column` do:

```swift
column.add(
    child: EqualSizeStack(
        direction: .horizontal,
        spacing: 16,
        configure: { stack in
            stack.add(child: Box(backgroundColor: .red))
            stack.add(child: Box(backgroundColor: .blue))
        }))
```

kyleve: Ah yeah, true. I guess having both the `children` and this threw me off

kyleve: (I had posted this, but my comment disappeared?)

Ah yeah. I guess having both the `children` argument and this threw me off, but that makes total sense.

bencochran: Oh I totally missed the `children` parameter 🤦‍♂. I guess maybe this doesn’t as much need the `configure` since it has no traits. Maybe just keep give one way or the other (array or `configure`).

watt: If picking one, I would probably lean towards the `configure` route for consistency with Row & Column. The `add(child:)` API also provides an opportunity for keyed children.

kyleve: +1

BlueprintUI/Sources/Internal/Extensions/FloatingPoint+ScaleRounding.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension FloatingPoint {
```
connorcimowsky: (copied this from BlueprintUICommonControls so that it can be internal to this module)

connorcimowsky:
Ready for another review! 🙏 /cc @kyleve @bencochran @watt @timdonnelly

BlueprintUI/Sources/Internal/Extensions/FloatingPoint+ScaleRounding.swift
connorcimowsky: just noticed that there's also a `ScaleRoundingTests` that I didn't copy over; anyone have opinions on whether I should move it (rather than copy it) into BlueprintUI and make it public?

kyleve: I'm fine with either, though it's nice to expose this in BlueprintUI so others can use it when writing their own elements, IMO.

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -0,0 +1,149 @@
...
+                    height: maximumItemHeight * CGFloat(items.count) + spacing * CGFloat(items.count - 1))
+            }
+
+            return totalSize
+        }
+
+        func layout(size: CGSize, items: [(traits: (), content: Measurable)]) -> [LayoutAttributes] {
+            guard items.count > 0 else { return [] }
+
+            var result: [LayoutAttributes] = []
```
connorcimowsky: note to self: move this to be next to `cumulativeSize`

BlueprintUI/Sources/Internal/Extensions/FloatingPoint+ScaleRounding.swift
watt: I've floated this idea before, but I think we want to avoid polluting `FloatingPoint` publicly. We could wrap the behavior up in a separate type to avoid that, but it's outside the scope of your change.

For now, I think copying this is fine. I've opened #27 to track the followup.

connorcimowsky: @watt sounds good, thanks!

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -0,0 +1,149 @@
...
+                    cumulativeSize += itemSize.width + spacing
+                case .vertical:
+                    attributes.frame.origin.y = cumulativeSize
+                    cumulativeSize += itemSize.height + spacing
+                }
+
+                result.append(attributes)
+            }
+
+            // Assign the remaining space to the nth item
```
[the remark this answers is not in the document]
connorcimowsky: as per our discussion offline, I removed the `roundingScale` code until we decide what to do about rounding on an infrastructural level 🙏

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -0,0 +1,118 @@
+/// An element that sizes all of its children equally, stacking them according to
+/// the specified `direction` and spacing them according to the specified `spacing`.
```
watt: What's the cross-axis layout story now?

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -0,0 +1,118 @@
...
+///
+/// - Note: A stack is measured to accommodate its largest child in each axis.
```
watt: This could be a little clearer — the measurement is the size required for *all* children to be sized to match the largest child, right?

BlueprintUI/Sources/Layout/EqualStack.swift
bencochran: Always fill

BlueprintUI/Sources/Layout/EqualStack.swift
connorcimowsky: agreed. will clarify!

BlueprintUI/Sources/Layout/EqualStack.swift
watt: Sorry, should have been clearer: can we add the cross-axis layout behavior to the documentation?

connorcimowsky: I already updated the note to (try to) document the behavior for both the primary/secondary axis:

```swift
/// - Note: A stack measures itself by determining its largest child in each axis, and
///         in the case of the primary axis, multiplying by `children.count` (accounting
///         for `spacing` as necessary).
```

if this isn't explicit enough, I could also add something along the lines of:

```swift
/// In the secondary axis, the stack will fill the space occupied by its largest child.
```

let me know!

watt: I see. I guess the way that reads to me, it only explains the measuring. I don't think it would hurt to be explicit about how children will be justified to the full size in the cross-axis.

connorcimowsky: here's what I went with:

```swift
/// An element that sizes its children equally, stacking them in the primary axis according to
/// the specified `direction` and spacing them according to the specified `spacing`. In the
/// secondary axis, children are justified to fill the space of the largest child.
///
/// - Note: A stack measures itself by determining its largest child in each axis, and
///         in the case of the primary axis, multiplying by `children.count` (accounting
///         for `spacing` as necessary).
```

PR #31, opened 2019-11-08
Encapsulate scale rounding into public type

### LINKED ISSUE #27: Scale rounding should be exposed publicly
We ought to be able to use the same behavior in both `BlueprintUI` and `BlueprintUICommonControls`, but ideally without publicly polluting `FloatingPoint`.

watt:
This PR wraps up the scale rounding logic into a public `RoundingScale` type, which allows it to be shared between modules without polluting the `FloatingPoint` protocol.

This type only operates on `CGFloat`, but in UI calculations that seems like a reasonable restriction.

This fixes #27


BlueprintUI/Sources/Measuring/RoundingScale.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+/// Represents a scale rounding behavior for `CGFloat`s.
+public enum RoundingScale: Equatable {
+    /// Do not round
+    case none
```
timdonnelly: Why is the `none` case included?

timdonnelly:
What is the concern with polluting the `FloatingPoint` protocol? 

I would lean toward an extension for ergonomic reasons:

```swift
let leftOrigin = (itemWidth*2/3).pixelAligned(screenScale: myScreenScale)
```

as opposed to

```swift
let leftOrigin =  RoundingScale.scale(2.0).round(itemWidth*2/3)
```

The former seems more idiomatic in Swift.

BlueprintUI/Sources/Measuring/RoundingScale.swift
watt: See @bencochran's [comment suggesting this previously](https://github.com/square/Blueprint/pull/18#discussion_r334164836)

watt:
> What is the concern with polluting the `FloatingPoint` protocol?

Personally, I think that in general it is poor form to publicly extend types from a library (unless those extensions are the library's main purpose). There's no way that I am aware of for a consumer to selectively import or avoid importing an extension in the way they can with top-level types, and it can be confusing to see those new methods pop up on existing types without an obvious source.

That said, this extension is pretty small risk; I'm willing to publish it publicly if there's a general consensus that it's OK.

> I would lean toward an extension for ergonomic reasons:

I agree that the extension is much more ergonomic, which is why it was written that way originally. :)

BlueprintUI/Sources/Measuring/RoundingScale.swift
timdonnelly: Ah – I see where that is coming from.

The name of the type (`RoundingScale`) currently implies that it represents a screen scale for use when rounding. In that case, `.none` seems unexpected.  I could instead just..... not call the rounding method.

It seems based on the discussion that this type is meant to model **the desired rounding behavior of an element**: so I would name it as such. `PixelRoundingBehavior` or something. `.none` is a very reasonable *behavior*, but it isn't a very useful *screen scale*.

timdonnelly: I would also add the use of this type to the documentation: it isn't a general purpose rounding tool, it's a way to specify a rounding behavior for an element

watt:
Superseded by #64

PR #32, opened 2019-11-08
Enforce subview order

watt:
The goal of this change is to ensure that subview arrays are ordered the same as their associated elements.

This will allow elements like `Overlay` to guarantee a z-ordering of elements from back to front.


BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -207,6 +210,9 @@ extension BlueprintView {
                     }
                     layoutTransition.perform {
                         child.layoutAttributes.apply(to: controller.view)
+
+                        contentView.insertSubview(controller.view, at: index)
```
kyleve: This just has the effect of moving the existing subview to the right index, right?

watt: That's the goal, yes.

kyleve: One thought I had after talking to Ben: Is `index` here guaranteed to be correct, even for elements that do not have backing views? Eg if i have 2 children that dont have backing views and one that does, is this index correct?

timdonnelly: It should be (but double check me here). Everything gets flattened before the update is applied.

If we had the hierarchy:

```
A
B {
  C
}
```

where A and C are view-backed, it will be flattened into:

[A, C] (in that order). C will still appear above A. That should work regardless of tree depth or complexity.

watt: This is the index of a `NativeViewController`, not any element. Every child here is view-backed.

PR #34, opened 2019-11-13
[WIP] Round all view frames to pixel boundaries

bencochran:
Proof of concept infrastructural fix for aligning views to pixel boundaries. This removes the need for layouts to do their own pixel-alignment.

This rounds each rect edge to the closest pixel (sometimes shrinking a frame, sometimes growing it, sometimes just moving it).

It’s still the case that certain elements may want to measure themselves rounded specifically (e.g. `Label` will want to round its measurement up to avoid truncation). For those cases, it may be worth passing the screen scale into `Measurable.measure` (with my implementation, if the main screen is 3x but the `BlueprintView` is on a screen that’s 1x, a `Label` won’t round up sufficiently to always avoid truncation).

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -30,6 +30,7 @@ public final class BlueprintView: UIView {
     private var needsViewHierarchyUpdate: Bool = true
     private var hasUpdatedViewHierarchy: Bool = false
     private var lastViewHierarchyUpdateBounds: CGRect = .zero
+    private var lastScreenScale: CGFloat = 1
 
     /// Used to detect reentrant updates
     private var isInsideUpdate: Bool = false
@@ -84,6 +85,14 @@ public final class BlueprintView: UIView {
         return element.content.measure(in: constraint)
     }
     
+    public override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        if let window = window, window.screen.scale != lastScreenScale {
+            setNeedsViewHierarchyUpdate()
+        }
+    }
+
     override public func layoutSubviews() {
         super.layoutSubviews()
         performUpdate()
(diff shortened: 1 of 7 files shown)


watt:
Thanks for putting this together!

I've put a little bit of thought into how this might work previously. I think it's worth noting that the way it's implemented here, it won't work the same as how AutoLayout and SwiftUI round to pixel boundaries. This implementation rounds each view’s frame within the coordinate system of its parent. AutoLayout & SwiftUI round based on the global, window-relative coordinates of each view.

This affects the rounding in subviews. If a view’s origin is at 0.4, and one of its subviews is at 0.3, this implementation will snap them both 0. AutoLayout & SwiftUI will snap the outer view to 0, but since the inner view is globally at 0.7, it would be rounded to 1.0. The effect becomes greater the deeper your view hierarchy gets and the more rounding you do.

I’m all for doing the rounding within BlueprintView, but I think we should consider mimicking the behavior of AutoLayout & SwiftUI.

timdonnelly:
Something like this might do the trick: it passes down the coordinate space origin relative to the screen space origin to allow each subview to round itself.

I think doing the rounding operation on the `NativeViewNode` tree is a pretty clean place to do all of this, versus including it in the `update` pass. Any thoughts there?

https://github.com/square/Blueprint/commit/55761962996c6a1b9731b8ea315f2e86064f4378

watt:
> Something like this might do the trick: it passes down the coordinate space origin relative to the screen space origin to allow each subview to round itself.
> 
> I think doing the rounding operation on the `NativeViewNode` tree is a pretty clean place to do all of this, versus including it in the `update` pass. Any thoughts there?

Yeah, I like this direction, although I was imagining passing down the parent's rounding offset to each child as just a "rounding offset" rather than "local origin in screen space". Have not actually tried it to see what has the best ergonomics though.

kyleve:
I also wonder if peer elements need to have some knowledge of each other.. eg, can we make this truly generic? For example if two elements are suppose to bump up against each other with no space, I could see this rounding causing a space.

PR #35, opened 2019-11-13
Make tests that do transforms be a little more lenient.

bencochran:
32-bit devices were failing tests because the resulting floats were slightly
off from expectations.


BlueprintUI/Tests/GeometryAssertions.swift
```diff
@@ -0,0 +1,35 @@
+import XCTest
+
+/// Asserts that the two floats are within a threshold of each other. Useful
+/// for testing the results of float math that may not result in exact values.
+func XCTAssertAlmostEqual(_ first: CGFloat, _ second: CGFloat, _ message: String = "", threshold: CGFloat = 0.0001, file: StaticString = #file, line: UInt = #line) {
```
bencochran: jk I just learned that there’s an XCTAssertEqual for floating points that takes an accuracy. will rework to use that.

bencochran: Updated! Also @watt taught me float things to use for the accuracy value rather than my arbitrary `0.0001`.

PR #36, opened 2019-11-13
Add a configure closure to the TextField initializer

kyleve:
- Add a configure closure
- Support changing text color and font

### DIFF
--- a/BlueprintUICommonControls/Sources/TextField.swift
+++ b/BlueprintUICommonControls/Sources/TextField.swift
@@ -28,8 +28,9 @@ public struct TextField: Element {
     public var becomeActiveTrigger: Trigger?
     public var resignActiveTrigger: Trigger?
 
-    public init(text: String) {
+    public init(text: String, configure: (inout TextField) -> Void = { _ in }) {
         self.text = text
+        configure(&self)
     }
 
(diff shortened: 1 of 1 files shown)


BlueprintUICommonControls/Sources/TextField.swift
```diff
@@ -10,6 +10,9 @@ public struct TextField: Element {
     public var onChange: ((String) -> Void)? = nil
     public var secure: Bool = false
     public var isEnabled: Bool = true
+    
+    public var font : UIFont? = nil
```
kyleve: These are nil so we can preserve the default color and font, and matches the properties themselves on UITextField.

kyleve:
Looks like this broke snapshots, will take a look

BlueprintUICommonControls/Sources/TextField.swift
```diff
@@ -41,6 +45,9 @@ public struct TextField: Element {
             configuration[\.onChange] = onChange
             configuration[\.isSecureTextEntry] = secure
             configuration[\.isEnabled] = isEnabled
+            
+            configuration[\.font] = font
```
kyleve: Need to switch to an `apply` block for this, so we can test `if self.font`...

PR #37, opened 2019-11-14
Swift Package Manager support

timdonnelly:


timdonnelly:
- Adds `Package.swift` for SPM support.
- Moves snapshot tests to Xcode 11.2/iOS 13.2 via SPM.
- Keeps a `BlueprintUI` test pass via Cocoapods for completeness.
- Adds a bunch of missing imports that are required when building via SPM.
- Moves vector code to modern Swift `SIMD` types to avoid warnings.

Podfile.lock
```diff
@@ -1,27 +0,0 @@
-PODS:
```
kyleve: Don't we want to keep this if the podfile is staying?

kyleve: Never mind... See there's one in the SampleApp dir.

timdonnelly: There were two, including an old unused version in `SampleApp`. `Podfile` + `.lock` now live in `SampleApp`.

watt:
Any idea why all these snapshots changed? Seems that they fail when running in Xcode 10.x now.

watt:
> Any idea why all these snapshots changed? Seems that they fail when running in Xcode 10.x now.

Actually it seems to be the result of running in iOS 13 instead of iOS 12.

timdonnelly:
Yeah, I scanned them for visual differences and they seemed ok.

watt:
My concern is that this makes `pod lib lint` fail, because it does not choose an iOS 13 simulator to run the snapshot tests.

timdonnelly:
Which simulator does it use? It isn't specified in the `podspec`.

timdonnelly:
Another option is to run all tests via SPM, and only perform a build step with Cocoapods.

PR #39, opened 2019-11-15
Change travis config to use iOS 9.3 for unit tests

watt:
In order to catch test failures that only happen on 32-bit devices (see #35).

I had to downgrade Xcode back to 10.2 in order for the Travis config to have an iOS 9.3 simulator available.

Open to discussion on whether we should run these tests twice on both 32-bit and 64-bit platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,11 @@ os: osx
 matrix:
   include:
     - language: swift
-      name: "Cocoapods BlueprintUI (Xcode 10.3)"
+      name: "Cocoapods BlueprintUI (Xcode 10.2)"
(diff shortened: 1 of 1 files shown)


timdonnelly:
> Open to discussion on whether we should run these tests twice on both 32-bit and 64-bit platforms.

We are already running them on 64-bit via SPM

watt:
> > Open to discussion on whether we should run these tests twice on both 32-bit and 64-bit platforms.
> 
> We are already running them on 64-bit via SPM

Oh, perfect, I thought it was only running snapshots. 👌

PR #42, opened 2019-12-16
Aligned Additions

kyleve:
- Add 'fill' options to the 'Aligned' element, to support content taking up the full width or height in each dimension.
- Add the UI tests to the 'SampleApp' workspace to allow running the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ matrix:
       xcode_destination: platform=iOS Simulator,OS=9.3,name=iPhone 4s
       before_install:
         - cd SampleApp
-        - gem update --system
-        - gem install bundler
         - bundle install
         - bundle exec pod repo update
         - bundle exec pod install
--- a/BlueprintUI/Sources/Layout/Aligned.swift
+++ b/BlueprintUI/Sources/Layout/Aligned.swift
@@ -2,8 +2,10 @@ import UIKit
 
 /// Aligns a content element within itself. The vertical and horizontal alignment may be set independently.
 ///
-/// The size of the content element is determined by calling `measure(in:)` on
-/// the content element – even if that size is larger than the wrapping element.
+/// When using alignment mode `.fill`, the content is scaled to the width or height of the `Aligned` element.
+///
+/// For other modes, the size of the content element is determined by calling `measure(in:)`
+/// on the content element – even if that size is larger than the wrapping element.
 ///
 public struct Aligned: Element {
     /// The possible vertical alignment values.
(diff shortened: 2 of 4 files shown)


kyleve:
Hmm, Travis is stuck, looking into it:

```
Installing RubyGems 3.1.1
  Successfully built RubyGem
  Name: bundler
  Version: 2.1.0
  File: bundler-2.1.0.gem
bundler's executable "bundle" conflicts with /Users/travis/.rvm/rubies/ruby-2.6.4/bin/bundle
Overwrite the executable? [yN]  
```

.travis.yml
```diff
@@ -15,8 +15,6 @@ matrix:
       xcode_destination: platform=iOS Simulator,OS=9.3,name=iPhone 4s
       before_install:
         - cd SampleApp
-        - gem update --system
-        - gem install bundler
```
kyleve: These lines aren't needed – they're already on Travis.

BlueprintUI/Sources/Layout/Aligned.swift
```diff
@@ -14,6 +14,9 @@ public struct Aligned: Element {
         case center
         /// Aligns the content to the bottom edge of the containing element.
         case bottom
+        
+        /// The content fills the full vertical height of the containing element.
+        case fill
```
watt: We should update the docs on this type if we add this, to something like:

> **For alignments other than `fill`,** the size of the content element is determined by calling `measure(in:)` on the content element – even if that size is larger than the wrapping element.

kyleve: Done

BlueprintUI/Sources/Layout/Aligned.swift
```diff
@@ -2,8 +2,10 @@ import UIKit
 
 /// Aligns a content element within itself. The vertical and horizontal alignment may be set independently.
 ///
-/// The size of the content element is determined by calling `measure(in:)` on
-/// the content element – even if that size is larger than the wrapping element.
+/// When using alignment mode `.fill`, the content is scaled to the width or height of the `Aligned` element.
+///
+/// For other modes, the size of the content element is determined by calling `measure(in:)`
+/// on the content element – even if that size is larger than the wrapping element.
```
kyleve: @watt not going to change it here, but this bit "even if that size is larger than the wrapping element." seems a bit weird to me – is there a use case for this, or should we change the implementation to cap the size?

watt: I copied this from `Centered`. I think it's reasonable behavior, as it limits the responsibility of this element and favors composition (also a reason that I have mixed feelings about adding `fill` here instead of introducing a `Fill` element). If you want to cap the size of an element, composition with `ConstrainedSize` seems like a more scalable solution.

PR #43, opened 2019-12-16
[WIP DNR] Respect numberOfLines for label sizing

kyleve:
Need to respect the `numberOfLines` on a `Label` or `AttributedLabel` when measuring elements. Otherwise, the size returned from the `measure` function ends up being wrong.

Ended up doing this by keeping around a static `UILabel`, which we can then use via `sizeThatFits` to get the correct size of the label. This seemed easier than dropping down to the CoreText or TextKit APIs... thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ matrix:
       xcode_destination: platform=iOS Simulator,OS=9.3,name=iPhone 4s
       before_install:
         - cd SampleApp
-        - gem update --system
-        - gem install bundler
         - bundle install
         - bundle exec pod repo update
         - bundle exec pod install
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,13 @@ matrix:
         - bundle install
         - bundle exec pod repo update
         - bundle exec pod install
+
     - language: swift
       name: "SPM (Xcode 11.2)"
       os: osx
       osx_image: xcode11.2
       script:
-        - xcodebuild -scheme "BlueprintUI-Package" test -destination "name=iPad Pro (9.7-inch)"
+        # Split into two commands so that we can pass 'quiet' during build to only see errors and warnings.
+        - xcodebuild build-for-testing -quiet -scheme "BlueprintUI-Package" -destination "name=iPad Pro (9.7-inch)"
+        - xcodebuild test-without-building -scheme "BlueprintUI-Package" -destination "name=iPad Pro (9.7-inch)"
 
--- a/BlueprintUICommonControls.podspec
+++ b/BlueprintUICommonControls.podspec
@@ -14,4 +14,9 @@ Pod::Spec.new do |s|
   s.source_files = 'BlueprintUICommonControls/Sources/**/*.swift'
 
   s.dependency 'BlueprintUI'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.source_files = 'BlueprintUICommonControls/Tests/**/*.swift'
+    test_spec.framework = 'XCTest'
+  end
 end
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -16,10 +16,13 @@ public struct AttributedLabel: Element {
     /// Creates a new label with the provided text and number of lines to display.
     public init(
         text: NSAttributedString,
-        numberOfLines: Int = 0
+        numberOfLines: Int = 0,
+        configure : (inout AttributedLabel) -> () = { _ in }
     ) {
         self.text = text
         self.numberOfLines = numberOfLines
+        
+        configure(&self)
     }
     
     /**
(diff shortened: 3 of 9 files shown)


kyleve:
Looking for feedback on this approach... Thoughts @watt @bencochran @timdonnelly ?

.travis.yml
```diff
@@ -15,8 +15,6 @@ matrix:
       xcode_destination: platform=iOS Simulator,OS=9.3,name=iPhone 4s
       before_install:
         - cd SampleApp
-        - gem update --system
-        - gem install bundler
```
kyleve: Fixes travis getting stuck

BlueprintUICommonControls.podspec
```diff
@@ -14,4 +14,9 @@ Pod::Spec.new do |s|
   s.source_files = 'BlueprintUICommonControls/Sources/**/*.swift'
 
   s.dependency 'BlueprintUI'
+
+  s.test_spec 'Tests' do |test_spec|
```
kyleve: Gets these tests showing up in Cocoapods.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -4,34 +4,43 @@ import UIKit
...
+    
+    /**
+     Keep around a label to use for measurement and sizing.
+     
+     We would usually do this using NSString or NSAttributedString's boundingRect family of methods,
+     but these do not let you specify a `numberOfLines` parameter, which is critical to correct sizing.
+     
+     As such, we will allocate this label once and then use it to measure by setting its text and attributes.
+     */
+    static let measurementLabel = UILabel()
```
kyleve: Verified that doing this does not have (m)any performance implications – measuring text 10,000 times with the label vs. the old boundingRect is about the same, +/- 5%.

kyleve:
@timdonnelly @watt @bencochran new year bump on this? matches what we ended up doing within posUI, due to the particulars of UILabel rendering.

BlueprintUICommonControls.podspec
kyleve: Yeah – but now you can't run them locally with Cocoapods. Also seems wise to run the tests in the Cocoapods environment in case there are resource, compilation issues, etc.

watt: The reason they were removed from the spec is that they require iOS 13, but `pod lib lint` and friends will execute them using a simulator for the minimum deployment target. So they will always cause linting to fail.

If you want to run them from Xcode, I recommend opening the project with the Package.swift file. It's nice!

kyleve: I've been doing that, but I really want them running on CI as well, since we do only consume this stuff via Cocoapods... Hmmm...

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Oops, yep!

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -4,34 +4,43 @@ import UIKit
...
-
-                return size
-            }
+        
+        return ElementContent { constraint in
+            let label = AttributedLabel.measurementLabel
+            let description = self.backingViewDescription(bounds: CGRect(origin: .zero, size: constraint.maximum), subtreeExtent: nil)!
+            
+            description.apply(to: label)
+            label.attributedText = self.attributedText
```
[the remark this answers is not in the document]
kyleve: Yep, my bad. I need to do a self-review pass on this.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -4,34 +4,43 @@ import UIKit
...
+            
+            var size = label.sizeThatFits(constraint.maximum)
+            
+            size.width = size.width.rounded(.up, by: self.roundingScale)
+            size.height = size.height.rounded(.up, by: self.roundingScale)
```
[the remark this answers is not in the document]
kyleve: Confirmed that it rounds to the screen scale:

```
(lldb) po size
▿ (153.5, 38.5)
  - width : 153.5
  - height : 38.5
```

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -4,34 +4,43 @@ import UIKit
...
+            
+            size.width = min(size.width, constraint.maximum.width, size.width)
```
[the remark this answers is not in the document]
kyleve: Going to constrain it in both, the recent find from @bizzemfrog shows this is necessary https://square.slack.com/archives/CKSU9GBFS/p1582219700026900, because of https://square.slack.com/archives/CKSU9GBFS/p1582226997037900

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -18,8 +18,26 @@ class AttributedLabelTests: XCTestCase {
         compareSnapshot(of: element)
         
     }
+    
+    func test_numberOfLines_sizing() {
```
kyleve: Definitely need to expand these tests for all the various cases we've found within SPOS.

BlueprintUICommonControls/Sources/AttributedLabel.swift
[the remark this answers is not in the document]
kyleve: Fwiw, not rounding to full px values does make the snapshots look a bit odd (new on right):

Screen Shot 2020-02-22 at 9 31 11 PM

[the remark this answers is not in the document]
kyleve: I remembered why we needed this, now: Because the label has `roundingScale`, we need to do this round up in order to conform to it. Otherwise the rounding returned is always the current device's scale, which is only right if you don't change `roundingScale`.

PR #47, opened 2020-01-16
Appearance Transition Refactor

kyleve:
### Overview
This PR significantly re-works and updates how `TransitionContainer` works to handle appearance and removal transitions. It's pretty much a complete redo of the API.

### Change Overview

- `TransitionContainer` is now `Transition`. Did this because most Blueprint elements are Containers, but do not include "Container" in their name.
- `AnimationAttributes` is now `AnimationOptions`. This more closely matches what the object does WRT the in-progress animation.
- `VisibilityTransition` is now `TransitionAnimation`. This is because it is an animation!
- Split animatable UIView properties out into `AnimatableViewProperties`, which is used for standard UIView animation types.
- We've introduced `CustomAnimationTransition`, a protocol for defining entirely custom animations (eg, backed by `CoreAnimation`). It has a simple API which is basically `func animate(..., completion : @escaping (Bool) -> ())`.
- `AnimationOptions` is now `TimingFunction`, and now has an option for defining the kind of animation, eg normal, spring, or keyframe when using UIView animations.
- Transitions by default are only performed when they are the outermost transition – inner transitions on inner views are not performed. You may override this if you wish. This prevents weird visual artifacts when nesting transitions within each other, which is almost never desired.
- `BlueprintView` gains the concept of explicitly animating updates (or not).
- The demo app is re-worked to provide a root list of demo items, so we can add them as we need them.

### Examples

Using the transition API remains very similar to before. Here is an example of creating a standard transition:

```
Transition(
     onAppear: .slideIn(),
     wrapping: Label(text: "Get 10 points for your purchase today.") { label in
         label.font = .systemFont(ofSize: 36.0, weight: .bold)
     }
)
```

And a custom transition:

```
Transition(
     onAppear: TransitionAnimation(custom: StatusProgressAnimation()),
     wrapping: StatusProgressElement(
         model: .init(initialPoints: 1, earnedPoints: 10, lowestTierPoints: 15),
         config: .init(radius: 100, lineWidth: 5, duration: 1.0)
     )
 )
```

And, an example of creating a `TransitionAnimation` leveraging spring animations:

```
 static func slideIn(
     from distance : CGFloat = -50.0,
     after delay : TimeInterval = 0.0,
     for duration : TimeInterval = 0.5
 ) -> Self {

     return TransitionAnimation(
         options: AnimationOptions(
             animationKind: .spring(dampingRatio: 0.5, velocity: 0.3),
             delay: delay,
             duration: duration,
             curve: .easeOut
         ),
         animations: AnimatableViewProperties(
             alpha: 0.0,
             transform: CGAffineTransform(translationX: 0.0, y: distance)
         )
     )
 }
```

### Remaining Work

- [ ] Still need to write tests – held off on this until we discuss the API a bit.


BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -28,33 +28,52 @@ import UIKit
...
+        }
+    }
+    
+    private var _element : Element?
+    
+    public func set(animated : Bool = BlueprintView.shouldAnimateChanges, element: Element?)
+    {
+        _element = element
+        
+        self.nextViewHierarchyUpdateEnablesAppearanceTransitions = animated
```
kyleve: Consider merging this with `needsViewHierarchyUpdate` via an enum instead of two bools

SampleApp/Podfile
```diff
@@ -1,6 +1,4 @@
 platform :ios, '12.0'
-inhibit_all_warnings!
-use_frameworks!
```
kyleve: This fixes a weird thing with Cocoapods I was seeing where the updated framework wasn't getting linked into the SampleApp, causing crashes. Couldn't debug any further so gave up and went with this...

SampleApp/Sources/DemoListScreen.swift
```diff
@@ -0,0 +1,107 @@
...
+        box.shadowStyle = .simple(radius: 6.0, opacity: 0.2, offset: .init(width: 0, height: 3.0), color: .black)
+        
+        return Button(onTap: self.onTap, wrapping: box)
+    }
+}
+
+
+fileprivate extension VisibilityTransition
+{
+    static func slideIn(after delay : TimeInterval, for duration : TimeInterval) -> VisibilityTransition
```
kyleve: Move this into the `VisibilityTransition` as one of the default options?

SampleApp/Sources/ExampleLoyaltyScreen.swift
```diff
@@ -0,0 +1,146 @@
...
+            }))
+            
+            return BlueprintUICommonControls.Button(wrapping: Box(backgroundColor: .init(white: 0.95, alpha: 1.0), wrapping: content))
+        }
+    }
+}
+
+fileprivate extension VisibilityTransition
+{
+    static func slideIn(after delay : TimeInterval, for duration : TimeInterval) -> VisibilityTransition
```
kyleve: ...And then we can get rid of this.

BlueprintUI/Sources/View Description/AppearanceAnimation.swift
```diff
@@ -0,0 +1,140 @@
+import UIKit
+
+/// The transition used when a view is inserted or removed during an update cycle.
+public struct AppearanceAnimation {
```
kyleve: Consider going back to `TransitionAnimation`

BlueprintUI/Sources/Animation/CustomTransitionAnimation.swift
```diff
@@ -0,0 +1,58 @@
...
+
+import UIKit
+
+
+/**
+ Allows you to create a custom animation for Blueprint's appearance transitions.
+ 
+ See individual methods for more information.
+ */
+public protocol CustomAnimation {
```
kyleve: This needs to be renamed to match the file

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -28,33 +28,53 @@ import UIKit
 public final class BlueprintView: UIView {
```
kyleve: In general this file needs a bit more of a look.

BlueprintUI/Sources/View Description/AnimationAttributes.swift
```diff
@@ -1,44 +0,0 @@
-import UIKit
-
-/// UIView animation configuration values.
-public struct AnimationAttributes {
```
kyleve: This is now `AnimationOptions`.

BlueprintUI/Sources/View Description/VisibilityTransition.swift
```diff
@@ -1,77 +0,0 @@
-import UIKit
-
-/// The transition used when a view is inserted or removed during an update cycle.
-public struct VisibilityTransition {
```
kyleve: This is now (loosely) `TransitionAnimation`.

BlueprintUICommonControls/Sources/TransitionContainer.swift
```diff
@@ -1,31 +0,0 @@
-import BlueprintUI
-import UIKit
-
-
-/// Wraps a content element and adds transitions when the element appears,
-/// disappears, or changes layout.
-public struct TransitionContainer: Element {
```
kyleve: This is now `Transition`.

SampleApp/Sources/DemoListScreen.swift
```diff
@@ -0,0 +1,89 @@
+//
+//  DemoListScreen.swift
```
kyleve: Rename this file?

SampleApp/Sources/ExampleLoyaltyScreen.swift
```diff
@@ -0,0 +1,124 @@
...
+//  Created by Kyle Van Essen on 1/16/20.
+//  Copyright © 2020 Square. All rights reserved.
+//
+
+import UIKit
+import BlueprintUI
+import BlueprintUICommonControls
+
+
+final class ExampleLoyaltyViewController : UIViewController
```
kyleve: Clean this up or remove?

SampleApp/Sources/ViewController.swift
```diff
@@ -28,12 +28,15 @@ let posts = [
 
 final class ViewController: UIViewController {
```
kyleve: Rename this

SampleApp/Sources/ViewController.swift
```diff
@@ -28,12 +28,15 @@ let posts = [
...
 
-    private let blueprintView = BlueprintView(element: List(posts: posts))
+    private let blueprintView = BlueprintView()
 
     override func loadView() {
         self.view = blueprintView
+        
+        Timer.scheduledTimer(withTimeInterval: 1.0, repeats: false) { _ in
```
kyleve: Remove

BlueprintUI/Sources/Animation/AnimationOptions.swift
```diff
@@ -0,0 +1,157 @@
...
+        self.allowUserInteraction = allowUserInteraction
+        
+        configure(&self)
+    }
+}
+
+
+extension AnimationOptions {
+    
+    public enum AnimationKind {
```
kyleve: Comment on the type

BlueprintUI/Sources/Animation/CustomTransitionAnimation.swift
```diff
@@ -0,0 +1,98 @@
...
+     Use this method to clean up any state your animation may have required during execution.
+     
+     You do not need to implement this method.
+     A default implementation of this method is provided, which does nothing.
+     */
+    func didEnd()
+}
+
+
+public protocol AnyCustomTransitionAnimation {
```
kyleve: Comment on what this is

BlueprintUI/Sources/Animation/CustomTransitionAnimation.swift
```diff
@@ -0,0 +1,98 @@
...
+        direction : TransitionAnimation.Direction,
+        with anyView : UIView,
+        currentProperties : AnimatableViewProperties,
+        completion : @escaping (Bool) -> ()
+    ) {
+        guard let containerView = anyView as? TransitionContainerView else {
+            fatalError()
+        }
+        
+        let contentView = containerView.subviews.first
```
kyleve: This is kinda gross... Clean up somehow? Or at least test it.

BlueprintUI/Sources/Animation/Transition.swift
```diff
@@ -0,0 +1,84 @@
+import UIKit
+
+
+/**
+ Wraps a content element and adds transitions when the element appears, disappears, or changes layout.
+ */
+public struct Transition<Wrapped:Element> : Element {
```
kyleve: More generally... How do we feel about moving towards making `wrapped` generic? Nice for instantiation and then configuration, eg:

```
var transition = Transition(onAppear: ..., wrapping: foo)

transition.wrapping.thing = bar
```

kyleve:
👋 @timdonnelly @watt @bencochran @kylebshr: Opening this up for your initial thoughts. Still some work and clean up to do, but the overall shape is here.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -245,7 +309,25 @@ extension BlueprintView {
...
+        mutating func add(appearanceTransition: TransitionAnimation?) {
+            
+            guard let transition = appearanceTransition else {
+                return
+            }
+            
+            self.parentAppearanceTransitions.append(transition)
+        }
+        
+        func animate(_ transition : TransitionAnimation) -> Bool {
```
kyleve: shouldAnimate

SampleApp/Sources/ExampleCustomAnimation.swift
```diff
@@ -0,0 +1,360 @@
+//
+//  StatusProgressView.swift
```
kyleve: Moved this over from SPOS since this is an actual thing we need to support, so figured may as well go right to the source for a real example...

BlueprintUI/Sources/Animation/Transition.swift
```diff
@@ -0,0 +1,84 @@
...
+    
+    //
+    // MARK: Initialization
+    //
+    
+    /// Creates a new `Transition` with both `onAppear` and `onDisappear` set to the provided `animation`.
+    public init(
+        animation: TransitionAnimation,
+        layout: LayoutTransition = .specific(AnimationOptions()),
+        wrapping element: Wrapped
```
kyleve: Consider making this a block that returns an element instead of a param? Generally this is more SwiftUI like, and allows inline config, and makes the API visually nicer when reading code.

BlueprintUI/Sources/Animation/Transition.swift
```diff
@@ -0,0 +1,84 @@
...
+        )
+    }
+
+    /**
+     Creates a new `Transition` with the provided options.
+    
+     If you only provide one of `onAppear` or `onDisappear`, only that direction of appearance
+     transition will be animated. The other one will be performed with no animation.
+     */
+    public init(
```
kyleve: We need to assert in here that the wrapped element is view backed (and maybe it’s type matches the animation?) for custom animations. Otherwise there could be multiple child views in the transition view.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
kyleve: This is good to go now.

BlueprintUI/Sources/Animation/CustomTransitionAnimation.swift
kyleve: Probably just test it... I think this is all we can do given how BlueprintView sets up its child views.

BlueprintUI/Sources/Animation/Transition.swift
kyleve: Or – likely smarter, wrap this in an element that _is_ view backed, which then drives animations for all children.

BlueprintUI/Sources/Animation/Transition.swift
```diff
@@ -0,0 +1,84 @@
...
+
+
+/**
+ A view class used to wrap elements that are within a `TransitionContainer`.
+ The animations of the `TransitionContainer` are applied to this view.
+ 
+ This is used instead of a plain `UIView` so when examining the view hierarchy,
+ there's an indication of where the view is coming from.
+ */
+public final class TransitionContainerView : UIView { }
```
kyleve: Add a debug description for this so it's clear what this in the debugger.

kyleve:
This is good for review now

BlueprintUI/Sources/Animation/TransitionAnimation+Predefined.swift
```diff
@@ -0,0 +1,67 @@
...
+     
+     The default options provide a slide animation that drops in from the top over 0.5 seconds.
+     
+     Provide a negative `distance` to slide in from the top, or a positive `distance` to slide in
+     from the bottom.
+     */
+    static func slideIn(
+        from distance : CGFloat = -50.0,
+        after delay : TimeInterval = 0.0,
+        for duration : TimeInterval = 0.5
```
watt: This API reads a little rough. Hold off on the new predefined animations for now?

SampleApp/Podfile
watt: Suspect this is a problem with your local environment, and shouldn't need to be changed as part of this PR.

kyleve: I've seen this same problem across Blueprint and Listable, on 3 separate computers and OS versions (some on Catalina, some not) for several months. Not sure how this could be a local thing given that...

kyleve: (I believe `use_frameworks!` fixes the underlying issue – but we also shouldn't be hiding warnings, so also removed that.)

BlueprintUICommonControls/Sources/TransitionContainer.swift
kyleve: I removed `Container` because basically every type that acts as a wrapper is a container, but no others have `container` in their name – eg `Inset` isn't `InsetContainer`, etc. I like `Transitioning`, I think

BlueprintUI/Sources/Animation/AnimationOptions.swift
kyleve: I like it

BlueprintUI/Sources/Animation/Transition.swift
kyleve: I marked this as public since I believe this affects how the type gets exported when viewing within stack traces, etc. Need to verify.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -245,7 +336,25 @@ extension BlueprintView {
             
             children = newChildren
         }
-        
     }
     
+    fileprivate struct UpdateContext {
+        private var parentAppearanceTransitions : [TransitionAnimation] = []
```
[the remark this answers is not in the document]
kyleve: True – was using this for debugging, which was helpful, but probably not useful for the average developer.

BlueprintUI/Sources/Internal/Extensions/UIView.swift
```diff
@@ -1,8 +1,9 @@
 import UIKit
 
+
 extension UIView {
     
-    final class var isInAnimationBlock: Bool {
+    public final class var isInAnimationBlock: Bool {
```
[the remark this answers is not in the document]
kyleve: Needed because this is now used within the default arg for public functions – if that ends up going away, can go back to internal.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -74,40 +114,36 @@ public final class BlueprintView: UIView {
...
     
     private func updateViewHierarchyIfNeeded() {
-        guard needsViewHierarchyUpdate || bounds != lastViewHierarchyUpdateBounds else { return }
+        
+        guard neededUpdateType != .none || bounds != lastViewHierarchyUpdateBounds else {
+            return
+        }
+                        
+        guard isInsideUpdate == false else {
+            fatalError("Reentrant updates are not supported in BlueprintView. Ensure that view events from within the hierarchy are not synchronously triggering additional updates.")
```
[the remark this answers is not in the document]
kyleve: This was intentional – seems weird to only fail in debug, but then expose us to who knows what in prod – would rather we always fail out.

BlueprintUI/Sources/Animation/TransitionAnimation.swift
```diff
@@ -0,0 +1,177 @@
...
+extension TransitionAnimation {
+    
+    /// When the animation should be performed relative to other animations
+    /// further up the tree (eg, `TransitionAnimations` for parent elements).
+    public enum PerformRule {
+        /// The animation will always be performed, even if it is nested in other transition animations.
+        case always
+        
+        /// The animation will only be performed if not nested in other transition animations.
+        case ifNotNested
```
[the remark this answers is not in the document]
kyleve: This is actually the primary change I wanted – visually it looks super weird to transition children with every possible nested transition, versus transitioning the outermost with the children already positioned (which is what `ifNotNested` allows). Eg, if every child's appearance is an alpha fade, you end up with visually weird timing curves because you've multiplying 70% of 70% as things fade in, etc.

Fixing this was one of the primary concerns when discussing things with @bencochran, so ccing him in if he has a better example.

BlueprintUI/Sources/Animation/Transition.swift
kyleve: Are they? Generics are pretty table stakes in Swift – and this doesn't affect the ability to use the value as an `Element`. If it were up to me, we'd be going back through all elements and making them generic around `Element`.

BlueprintUICommonControls/Sources/TransitionContainer.swift
kyleve: Or maybe `OnTransition(onAppear:onDisappear:)`?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -27,44 +27,84 @@ import UIKit
...
+            self.set(element: newValue)
+        }
+        get {
+            return self._element
+        }
+    }
+    
+    private var _element : Element?
+    
+    public func set(animated : Bool = UIView.isInAnimationBlock, element: Element?)
```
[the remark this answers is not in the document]
kyleve: Correct – since this is the "seam" between declarative and non-declarative land – you may be using a blueprint view within your normal UIKit views, which is what this is. Maybe I should remove the param and just rely on if we're in an animation block?

BlueprintUI/Sources/View Description/ViewDescription.swift
```diff
@@ -45,9 +45,9 @@ public struct ViewDescription {
     private let _apply: (UIView) -> Void
     private let _contentView: (UIView) -> UIView
     
+    private let _onAppear: TransitionAnimation?
+    private let _onDisappear: TransitionAnimation?
     private let _layoutTransition: LayoutTransition
-    private let _appearingTransition: VisibilityTransition?
-    private let _disappearingTransition: VisibilityTransition?
```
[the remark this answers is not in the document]
kyleve: Swift style generally avoids leaving off the type name from the variable (eg, UIScreen.main), since the type system guides you, which is why I changed this. Would rather not encode the entire type name in its own var – thoughts?

BlueprintUI/Sources/Animation/Transition.swift
kyleve: Going to remove this for now, but I think it makes sense to convert these to generics eventually.

BlueprintUI/Sources/Animation/Transition.swift
kyleve: _seems_ to be fine without the public, removing.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -245,7 +336,25 @@ extension BlueprintView {
             
             children = newChildren
         }
-        
     }
     
+    fileprivate struct UpdateContext {
```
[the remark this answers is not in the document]
kyleve: Keeping as `UpdateContext`, since it's the context passed through the `update(...` call tree.

BlueprintUI/Sources/View Description/ViewDescription.swift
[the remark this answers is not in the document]
kyleve: Still noodling on these..

BlueprintUI/Sources/Animation/TransitionAnimation.swift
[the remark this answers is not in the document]
kyleve: Adding docs to clear this behaviour up.

BlueprintUI/Sources/Animation/TransitionAnimation.swift
```diff
@@ -0,0 +1,177 @@
...
+        /**
+         A custom animation which can be backed by CoreAnimation, CADisplayLink, etc.
+         
+         When performing an animation, Blueprint will call your provided factory to create a new
+         `CustomTransitionAnimation` for each animation instance. This allows you to hold onto state
+         within your `CustomTransitionAnimation` object to track and manage the animation.
+         
+         The `CustomTransitionAnimation` object is retained until the completion block passed to it is called.
+         */
+        case custom(() -> AnyCustomTransitionAnimation)
```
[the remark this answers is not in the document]
kyleve: Sure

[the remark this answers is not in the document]
kyleve: - [ ] Remove custom animations into a follow up PR

BlueprintUI/Sources/Animation/TransitionAnimation+Predefined.swift
kyleve: Sure – though may even rip out the existing ones? Not sure anyone uses them and I dont think they're particularly useful.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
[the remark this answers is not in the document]
kyleve: Going to add an `Animated` element as well, but also leave this – both serve their purpose depending on the layer of interaction (UIView layer vs. within the element tree).

BlueprintUI/Sources/View Description/ViewDescription.swift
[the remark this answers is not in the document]
kyleve: I havent come up with anything much better here... I still like these names more than the old ones. Thoughts?

BlueprintUI/Sources/Animation/Animated.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Animated.swift
+//  BlueprintUI
+//
+//  Created by Kyle Van Essen on 4/6/20.
+//
+
+import Foundation
+
+
+/**
```
kyleve, line 11: Missing comments on this type

BlueprintUI/Sources/Animation/CustomTransitionAnimation.swift
kyleve: - [ ] comment

BlueprintUI/Sources/Animation/Transition.swift
kyleve: - [ ] wrap in view backed element

kyleve: Wait, nvm.. this eventually will reduce down to a single view backed element.. I think? Need to verify this.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -138,11 +164,22 @@ public final class BlueprintView: UIView {
...
-    
 }
 
 extension BlueprintView {
     
+    private enum UpdateType : Equatable {
+        case none
+        case needsViewHierarchyUpdate(animated : Bool)
+        
+        var animated : Bool {
```
kyleve, line 175: Needed still?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -191,52 +237,106 @@ extension BlueprintView {
                 let (path, child) = node.children[index]
 
                 guard usedKeys.contains(path) == false else {
-                    fatalError("Duplicate view identifier")
+                    fatalError("Blueprint Error: Duplicate view identifier: \(path).")
                 }
+                
                 usedKeys.insert(path)
+                
+                var childContext = context
```
kyleve, line 245: Comment; also rename `context` to `parentContext`?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -191,52 +237,106 @@ extension BlueprintView {
...
+                let transition = controller.viewDescription.onDisappear
+                
+                var childContext = context
+                
+                childContext.setAllowsAnimations(
+                    with: node.viewDescription.allowsAnimations,
+                    updateRequestedAnimations: animated
+                )
+                
+                if let transition = transition, context.allowsAnimations {
```
kyleve, line 334: `context` should be `childContext`

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -249,3 +349,43 @@ extension BlueprintView {
     }
 }
 
+
+fileprivate struct UpdateContext {
+
+    // MARK: Animations Enabled
+    
+    var allowsAnimations : Bool = true
```
kyleve, line 357: `private(set) var`

SampleApp/SampleApp.xcodeproj/project.pbxproj
```diff
@@ -739,6 +755,7 @@
 			baseConfigurationReference = 0DA29F056002872418F7D2C9 /* Pods-SampleApp.debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				CLANG_ENABLE_MODULES = YES;
```
kyleve, line 758: See if I can remov

SampleApp/Sources/DemoListScreen.swift
kyleve: - [ ] Rename

BlueprintUI/Sources/View Description/ViewDescription.swift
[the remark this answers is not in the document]
kyleve: AppearanceTransition seems reasonable, but a bit wordy; going with that for now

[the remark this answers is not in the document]
kyleve: (Though, it also contains a `LayoutTransition`, which isnt necessarily related to appearance.. hm)

BlueprintUI/Sources/Animation/Animated.swift
```diff
@@ -0,0 +1,78 @@
...
+    // MARK: Element
+    //
+    
+    public var content: ElementContent {
+        ElementContent(child: self.wrapped)
+    }
+    
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        ViewDescription(AnimatedView.self) {
+            $0.builder = { AnimatedView(frame: bounds) }
```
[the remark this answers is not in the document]
kyleve: Will do, but IMO this

```
        ViewDescription(AnimatedView.self) {
            $0.builder = { AnimatedView(frame: bounds) }
            $0.allowsAnimations = self.animated
        }
```

Is really no more or less clear than this:

```
        ViewDescription(AnimatedView.self) { config in
            config.builder = { AnimatedView(frame: bounds) }
            config.allowsAnimations = self.animated
        }
```

BlueprintUI/Sources/View Description/ViewDescription.swift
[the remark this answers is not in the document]
kyleve: (Going with `AppearanceTransition`)

PR #50, opened 2020-01-23
Bump podspec versions to 0.5

kyleve:
Need to ship a 0.5 to get some recent updates

### DIFF
--- a/BlueprintUI.podspec
+++ b/BlueprintUI.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'BlueprintUI'
-  s.version      = '0.4.0'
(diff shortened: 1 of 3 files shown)


watt:
In previous releases we've done this directly on master. Make sure you tag the merge commit before publishing. See RELEASING.md for more info.

kyleve:
@watt Saw that in the releasing doc – though I tend to prefer to do it in PR and then publish once merged, to get another paid of eyes?

watt:
@kyleve I'm fine with doing it this way, but we should update the docs and try to consistently follow them, so that any maintainer can release the same way. If we're not following our docs then there's no point in having them.

PR #53, opened 2020-01-29
Add alignment property to TextField

meherkasam:


BlueprintUICommonControls.podspec
```diff
@@ -14,4 +14,9 @@ Pod::Spec.new do |s|
   s.source_files = 'BlueprintUICommonControls/Sources/**/*.swift'
 
   s.dependency 'BlueprintUI'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.source_files = 'BlueprintUICommonControls/Tests/**/*.swift'
+    test_spec.framework = 'XCTest'
+  end
```
meherkasam: I couldn't figure out how else to run the snapshot tests for `BlueprintUICommonControls`, so I added this. Does this belong here?

BlueprintUICommonControls/Sources/TextField.swift
```diff
@@ -10,6 +10,7 @@ public struct TextField: Element {
     public var onChange: ((String) -> Void)? = nil
     public var secure: Bool = false
     public var isEnabled: Bool = true
+    public var textAlignment: NSTextAlignment = .natural
```
kyleve: The default value on UITextField is `.left` – is it worth matching that? https://developer.apple.com/documentation/uikit/uitextfield/1619643-textalignment

PR #54, opened 2020-02-04
Fix ScrollView layout and vertical bounce behavior

lechristian:
When UIRefreshControl is added to UIScrollView, it follows the
UIScrollView contentInset. UIRefreshControl's frame's width is
always the scroll view's frame's width. This causes the
refresh control to look off center. Subtract the scroll view's
contentInset.left to account for this.

Adding UIRefreshControl to a UIScrollView with disabled
`alwaysBounceVertical` whose content size is less than the
scroll view's height will not allow pulling on the scroll view
to refresh. If the refresh behavior is not `disabled` then
set `alwaysBounceVertical` to `true`.

### DIFF
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -221,12 +221,13 @@ fileprivate final class ScrollerWrapperView: UIView {
     /// The current `ScrollView` state we represent.
     private var representedElement : ScrollView
 
-    private var refreshControl: UIRefreshControl? = nil {
-
-        didSet {
-            scrollView.refreshControl = refreshControl
+    private var refreshControl: UIRefreshControl? {
+        get {
+            return scrollView.refreshControl
         }
-
+        set {
+            scrollView.refreshControl = newValue
+        }
     }
 
     private var refreshAction: () -> Void = { }
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -221,7 +221,7 @@ fileprivate final class ScrollerWrapperView: UIView {
     /// The current `ScrollView` state we represent.
     private var representedElement : ScrollView
 
-    private var refreshControl: UIRefreshControl? = nil {
+    private var refreshControl: RefreshControl? = nil {
 
         didSet {
             scrollView.refreshControl = refreshControl
@@ -272,7 +272,7 @@ fileprivate final class ScrollerWrapperView: UIView {
             refreshControl = nil
         case .enabled, .refreshing:
             if refreshControl == nil {
-                let control = UIRefreshControl()
+                let control = RefreshControl()
                 control.addTarget(self, action: #selector(didPullToRefresh), for: .valueChanged)
                 refreshControl = control
             }
@@ -298,9 +298,15 @@ fileprivate final class ScrollerWrapperView: UIView {
         if self.scrollView.alwaysBounceHorizontal != scrollView.alwaysBounceHorizontal {
             self.scrollView.alwaysBounceHorizontal = scrollView.alwaysBounceHorizontal
         }
-
-        if self.scrollView.alwaysBounceVertical != scrollView.alwaysBounceVertical {
+        
+        switch scrollView.pullToRefreshBehavior {
+        case .disabled:
             self.scrollView.alwaysBounceVertical = scrollView.alwaysBounceVertical
+        case .enabled, .refreshing:
+            // The refresh behavior will not work when `alwaysBounceVertical` is false
+            // and the content size is less than the parent view size (no scroll is needed).
+            // Set `alwaysBounceVertical` to true when the `refreshControl` is needed.
+            self.scrollView.alwaysBounceVertical = true
         }
 
         if self.scrollView.contentSize != contentSize {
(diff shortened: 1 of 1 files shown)


BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -267,3 +273,23 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+
+    override var frame: CGRect {
+        get {
+            return super.frame
+        }
+        set {
+            // UIRefreshControl frame does not account for UIScrollView's left inset. Offset the
+            // frame with negative inset so that the `refreshControl` is centered with respect to the content.
+            var newFrame = newValue
+            if let parentScrollView = superview as? UIScrollView {
```
timdonnelly: This is very surprising behavior for a property. Even if it's a private implementation detail, it's pretty fragile – it makes assumptions about the view hierarchy, and it assumes that it is the child view of a scroll view at the time frame is assigned. It's also surprising that the parent can assign one frame, but a different frame is actually assigned... This effectively means that the child is making layout decisions about itself. Parents should always be entirely responsible for the layout of their child views, and this breaks that contract. The following code seems to work well in a sample project. It accomplishes the same thing, but it keep layout decisions in the parent view:

```
final class CustomScrollView: UIScrollView {
    
    override func layoutSubviews() {
        super.layoutSubviews()
        refreshControl?.center.x = bounds.midX
    }
    
}
```

timdonnelly:
Thanks for fixing! I'm surprised that this is the standard `UIKit` behavior... Content insets shouldn't break refresh control alignment, but they certainly do.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: (Also, does this happen in a plain UIScrollView in a new project?)

timdonnelly: It does, I just tested.

kyleve: Yay...

lechristian: Thanks, will update. Originally had the refresh control layout in `layoutSubviews` but for some reason it wasn't laying out correctly.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -267,3 +266,37 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+private extension ScrollerWrapperView {
+
+    final class CustomScrollView: UIScrollView {
+
+        override var refreshControl: UIRefreshControl? {
+            get {
+                if #available(iOS 10.0, *) {
+                    return super.refreshControl
+                } else {
+                    return refreshControl
```
kyleve: I guess this isn't an recursive bomb since Swift knows to return the underlying value and not go through the getter it's already in?

timdonnelly: No, I think it's a 💣.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -143,19 +143,12 @@ extension ScrollView {
 
 fileprivate final class ScrollerWrapperView: UIView {
     
-    let scrollView = UIScrollView()
+    let scrollView = CustomScrollView()
 
     private var refreshControl: UIRefreshControl? = nil {
 
         didSet {
```
timdonnelly: Rather than duplicating storage, I'd just forward this property to the scroll view:

```
var refreshControl: UIRefreshControl? {
    get { scrollView.refreshControl }
    set { scrollView.refreshControl = newValue }
}
```

BlueprintUICommonControls/Sources/ScrollView.swift
lechristian: Yea it is, I realized this after I blew up my own feature.

Using a `private var _refreshControl` to manage this for < 10. Not sure if I should still set it on >= 10 though.

BlueprintUICommonControls/Sources/ScrollView.swift
timdonnelly: Or, actually – this is a private `var`. Should it be removed, and callers can just use `scrollView.refreshControl`?

BlueprintUICommonControls/Sources/ScrollView.swift
timdonnelly: Leaving it nil is ok, you can save a `objc_retain` call and make everything 0.00000000000001% faster.

BlueprintUICommonControls/Sources/ScrollView.swift
lechristian: Most just saves on typing `scrollView.refreshControl` inside the `apply`? 

Since it's not doing any additional logic anymore, probably more explicit to use `scrollView.refreshControl` though.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -267,3 +263,38 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+
+private extension ScrollerWrapperView {
+
+    final class CustomScrollView: UIScrollView {
+
+        private var _refreshControl: UIRefreshControl? = nil
+
+        override var refreshControl: UIRefreshControl? {
+            get {
+                if #available(iOS 10.0, *) {
```
kyleve: Thought: We've deprecated iOS 9 in SPOS, so we can support iOS 10+ here. Shall we bump the min deployment target and remove this special casing?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -222,8 +212,14 @@ fileprivate final class ScrollerWrapperView: UIView {
             self.scrollView.alwaysBounceHorizontal = scrollView.alwaysBounceHorizontal
         }
 
-        if self.scrollView.alwaysBounceVertical != scrollView.alwaysBounceVertical {
```
watt: May want to preserve the logic that prevents setting this property to the same value.

kyleve:
Bump on this

kyleve:
@lechristian ping, any plans to move this forward? if not we can take it over

PR #55, opened 2020-02-17
ScrollView observes and adjusts for keyboard position

kyleve:
This PR makes `ScrollView` smart about observing the position of the system keyboard. When the keyboard is displayed, updated, or dismissed, the `ScrollView` will adjust its `contentInset.bottom` to account for the updated inset. This is done through `KeyboardObserver`, which is an object to observe and manage keyboard positioning, which I've copied over from Listable. I've also added `keyboardDismissMode` to `ScrollView` in this PR, allowing management of keyboard dismissal in `ScrollView` instances.


BlueprintUICommonControls/Tests/Sources/Internal/KeyboardObserverTests.swift
```diff
@@ -0,0 +1,189 @@
+//
+//  KeyboardObserverTests.swift
+//  Listable-Unit-Tests
```
kyleve: This is wrong now

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -17,6 +17,7 @@ public struct ScrollView: Element {
     public var showsHorizontalScrollIndicator: Bool = true
     public var showsVerticalScrollIndicator: Bool = true
     public var pullToRefreshBehavior: PullToRefreshBehavior = .disabled
+    public var keyboardDismissMode: UIScrollView.KeyboardDismissMode = .none
```
kyleve: This matches the UIKit default, but it also is almost nearly wrong... Consider `.interactive`?

Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,168 @@
+//
+//  KeyboardObserver.swift
```
kyleve: Xcode strikes again... Put this into the BlueprintUICommonControls dir.

Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,168 @@
...
+    
+    private var latestNotification : NotificationInfo?
+    
+    //
+    // MARK: Handling Changes
+    //
+    
+    enum KeyboardFrame : Equatable
+    {
+        case notVisible
```
bizzemfrog: nit: would name this noOverlap, or notInView. notVisible makes it sound like its not on screen at all, when it might be, but just not overlapping the passed in view.

kyleve: I went back and forth on this, but found the other things awkward... Let me see if I can come up with something better.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Leaving as is.

Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,168 @@
...
+        self.animate(with: new) {
+            self.delegate?.keyboardFrameWillChange(observer: self)
+        }
+    }
+    
+    //
+    // MARK: Notification Listeners
+    //
+    
+    @objc func keyboardFrameChanged(_ notification : Notification)
```
kyleve: Clean this up

BlueprintUICommonControls.podspec
```diff
@@ -14,4 +14,9 @@ Pod::Spec.new do |s|
   s.source_files = 'BlueprintUICommonControls/Sources/**/*.swift'
 
   s.dependency 'BlueprintUI'
+
+  s.test_spec 'Tests' do |test_spec|
```
kyleve: Added to get my tests running, need to figure out if we can actually add this back since it breaks pod linting.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -237,6 +240,10 @@ fileprivate final class ScrollerWrapperView: UIView {
         if self.scrollView.showsHorizontalScrollIndicator != scrollView.showsHorizontalScrollIndicator {
             self.scrollView.showsHorizontalScrollIndicator = scrollView.showsHorizontalScrollIndicator
         }
+        
+        if self.scrollView.keyboardDismissMode != scrollView.keyboardDismissMode {
```
vocaro: Why is this conditional needed?

kyleve: Ah right, I need to get rid of this since your PR added this in.

As for the conditional, I _think_ this is a safety net to ensure the scroll view isn’t resetting any of its own state if the value didn’t change (even though in theory, it shouldn’t) - the other properties within this apply method follow the same convention.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -262,8 +269,53 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+        self.scrollView.contentInset.bottom = inset
+    }
+    
+    //
+    // MARK: KeyboardObserverDelegate
+    //
+    
+    func keyboardFrameWillChange(observer : KeyboardObserver)
+    {
+        self.setContentInsetWithKeyboardFrame()
```
watt: Can we have a property that enables this behavior, rather than hardcoding it in?

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ Implementation borrowed from Listable:
+ https://github.com/kyleve/Listable/blob/master/Listable/Sources/Internal/KeyboardObserver.swift
+ 
+ iOS Docs for keyboard management:
+ https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html
+ */
+final class KeyboardObserver
```
watt: Why not build this into the scroll view? It seems too tightly coupled to be reusable with any other view.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,172 @@
...
+        self.latestNotification = new
+        
+        if let old = old {
+            guard old.endingFrame != new.endingFrame else {
+                return
+            }
+        }
+        
+        self.animate(with: new) {
+            self.delegate?.keyboardFrameWillChange(observer: self)
```
watt: It seems weird to me that this class wraps delegate calls in animation block. That's a lot of responsibility for an "observer". The decision of "how to react to this keyboard frame change" could probably be left up to the delegate.

BlueprintUICommonControls/Tests/Sources/Internal/KeyboardObserverTests.swift
```diff
@@ -0,0 +1,182 @@
...
+
+
+extension XCTestCase
+{
+    func testcase(_ name : String = "", _ block : () -> ())
+    {
+        block()
+    }
+    
+    func assertThrowsError(test : () throws -> (), verify : (Error) -> ())
```
watt: XCTAssertThrowsError?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -262,8 +269,53 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+    private func setContentInsetWithKeyboardFrame() {
+        guard let frame = self.keyboardObserver.currentFrame(in: self) else {
+            return
+        }
+        
+        var inset : CGFloat
+        
+        switch frame {
+        case .notVisible: inset = 0.0
+        case .visible(let frame): inset = (self.bounds.size.height - frame.origin.y)
```
watt: I hate to nit on formatting but I find this line hard to read.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
timdonnelly: I don't mind keeping it separate – it's a chunk of logic that's pretty easy to contain. I wish `UIKit` exposed an API more like this, to be honest.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,172 @@
...
+    //
+    // MARK: Initialization
+    //
+    
+    init(center : NotificationCenter = .default)
+    {
+        self.center = center
+
+        self.center.addObserver(self, selector: #selector(keyboardFrameChanged(_:)), name: UIWindow.keyboardWillChangeFrameNotification, object: nil)
+        self.center.addObserver(self, selector: #selector(keyboardFrameChanged(_:)), name: UIWindow.keyboardDidChangeFrameNotification, object: nil)
```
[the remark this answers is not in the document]
timdonnelly: Those should probably trigger different behaviors if we want animations to work correctly...

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,172 @@
...
+//  BlueprintUICommonControls
+//
+//  Created by Kyle Van Essen on 2/16/20.
+//
+
+import UIKit
+
+
+protocol KeyboardObserverDelegate : AnyObject
+{
```
timdonnelly: Why are `{`s on a newline? That's (very) unusual for Swift, and breaks convention with the rest of the project.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
timdonnelly: I agree that this is unexpected. I've always modeled the animation attributes in a type that is vended to the delegate (`keyboardFrameWillChange(to:animationAttriubutes:)` or something).

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: This is the only right way to write swift and everyone else is wrong!

(It's copied from another repo, will fix... But I still hate it.)

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: One reason for this is because the animation curve for keyboards is a private value you need to get by shifting the curve by << 16. I'll fix by passing in the pre-shifted value for the delegate to use?

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Sure

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
timdonnelly: Yeah, I've kept the conversion internal and provided the final value. `KeyboardObserver` isn't public though, so these API quirks don't leak out.

BlueprintUICommonControls/Sources/ScrollView.swift
timdonnelly: I probably wrote most of the other conditionals – `UIScrollView` has a few properties (including `frame`, `contentSize`, and insets, IIRC) that reset any in-progress scrolling when they are set, regardless of whether the values have actually changed.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: Yeah, wanted to keep this separate because the nuts and bolts of it aren't related to UIScrollView at all – could see us using this in more elements, plus it makes it easier to test.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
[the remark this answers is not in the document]
kyleve: Now to remember why I did this... I think `will` is the one we want to use for animations, I forget why I included did..

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Thoughts on the name? `adjustsContentInsetForKeyboard`? Or could do an enum `KeyboardAdjustmentType` and have a `.none` and `.adjusts`, which reads a bit nicer.

kyleve: Going with `keyboardAdjustmentMode` for now.

kyleve:
Pushed some, but not all requested changes, will get to those later.

This update also includes changes to how we manage `contentInset` overall, because the provided `ScrollView.contentInset` and the keyboard's bottom inset needed to play nicer together. This updates ensures we continue using `ScrollView.contentInset.bottom` alongside the additional inset from the keyboard. Also centralized calculation of the final inset to one place.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -262,8 +304,120 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+        }
+    }
+}
 
 
+extension ScrollView
+{
+    // Calculates the correct content inset to apply for the given inputs.
+    
+    static func finalContentInset(
```
kyleve: Test me

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,166 @@
...
+            animationDuration: new.animationDuration,
+            options: animationOptions
+        )
+    }
+    
+    //
+    // MARK: Notification Listeners
+    //
+    
+    @objc func keyboardFrameChanged(_ notification : Notification) {
```
kylebshr: Any reason not to make this private?
```suggestion
    @objc private func keyboardFrameChanged(_ notification : Notification) {
```

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -262,8 +304,120 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+    //
+    // MARK: KeyboardObserverDelegate
+    //
+    
+    func keyboardFrameWillChange(
+        for observer : KeyboardObserver,
+        animationDuration : Double,
+        options : UIView.AnimationOptions
+    ) {
+        if animationDuration > 0.0 {
```
kylebshr: The docs for `UIView.animate` specifically state:

> If you specify a negative value or 0, the changes are made without animating them.

Is there a precedent for needing this check?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -262,8 +304,120 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+        case .adjustsWhenVisible:
+            guard let keyboardFrame = self.keyboardObserver.currentFrame(in: self) else {
+                return 0.0
+            }
+            
+            var bottomInset : CGFloat
+            
+            switch keyboardFrame {
+            case .notVisible: bottomInset = 0.0
+            case .visible(let frame): bottomInset = (self.bounds.size.height - frame.origin.y)
```
kylebshr: Are you sure `bounds` is the correct height to use for this calculation, and not `frame`? If the content is taller than the frame of the scroll view, won't this calculation be incorrect? Maybe I misunderstand what's being calculated though 😅

kylebshr:
Could you verify [this](https://github.com/kyleve/Listable/issues/100) isn't also an issue with the implementation here?

bizzemfrog:
any movement on this?

kyleve:
Nope... can try to look soon.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: TIL you can directly suggest changes in Github

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Huh, I could swear this never used to happen. Will remove!

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: Landing this locally

kyleve:
> Could you verify [this](https://github.com/kyleve/Listable/issues/100) isn't also an issue with the implementation here?

Doesn't seem to be an issue from testing

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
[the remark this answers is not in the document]
kyleve: Dug into this more: When there's a floating keyboard or split keyboard, will is not called for frame changes, only did is called once the keyboard position changes. Because we guard based on the last reported frame, animations still work as expected. Adding a comment inline.

BlueprintUICommonControls/Tests/Sources/Internal/KeyboardObserverTests.swift
kyleve: Good point. I've just been mindlessly copy/pasting this around since before that existed

SampleApp/Sources/AppDelegate.swift
```diff
@@ -8,7 +8,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
         window = UIWindow(frame: UIScreen.main.bounds)
-        window?.rootViewController = ViewController()
+        window?.rootViewController = ScrollViewKeyboardViewController()
```
kyleve: Revertme before merging.

kyleve:
This should be good for another review.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -37,6 +37,13 @@ final class KeyboardObserver {
     init(center : NotificationCenter = .default) {
         
         self.center = center
+        
+        /// We need to listen to both `will` and `keyboardWillChangeFrame` notifications. Why?
```
kylebshr:
```suggestion
        /// We need to listen to both `will` and `keyboardDidChangeFrame` notifications. Why?
```

I think you meant to make on of these `did`?

kyleve: Doh

BlueprintUICommonControls.podspec
kyleve: Still need to remove this.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -260,8 +300,137 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+}
 
+
+extension ScrollerWrapperView : KeyboardObserverDelegate {
+    
+    //
+    // MARK: Keyboard
+    //
+    
+    private func updateBottomContentInsetWithKeyboardFrame() {
```
watt: This is almost the same func as `applyContentInset`, can they merge?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -260,8 +300,137 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+        }
+    }
+}
 
 
+extension ScrollView
+{
+    // Calculates the correct content inset to apply for the given inputs.
+    
+    static func finalContentInset(
```
watt: Why is this static? What's "final" about it? If it's for calculating the content inset, maybe a name like `calculateContentInset()`?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -231,20 +262,29 @@ fileprivate final class ScrollerWrapperView: UIView {
...
-            contentInset.top += refreshControl.bounds.height
-        }
-
+        
+        self.applyContentInset(with: scrollView)
+    }
+    
+    private func applyContentInset(with scrollView : ScrollView)
+    {
+        // Keep a copy of the content inset as provided by the developer.
```
watt: What does this mean? Where are we keeping "a copy of the content inset as provided by the developer"?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -145,6 +164,10 @@ extension ScrollView {
 fileprivate final class ScrollerWrapperView: UIView {
     
     let scrollView = UIScrollView()
+    let keyboardObserver = KeyboardObserver()
+    
+    /// The current `ScrollView` state we represent.
+    private var representedElement : ScrollView
```
watt: Can we just hang onto the values we need rather than the whole element? This is really tightly coupled.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -29,22 +40,25 @@ public struct ScrollView: Element {
...
+            }
+            
             config.contentView = { $0.scrollView }
-            config.apply({ (view) in
-                view.apply(scrollView: self, contentFrame: subtreeExtent ?? .zero)
-            })
+            
+            config.apply {
+                $0.apply(scrollView: self, contentFrame: subtreeExtent ?? .zero)
+            }
```
watt: Intentional change?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -12,12 +12,23 @@ public struct ScrollView: Element {
     public var contentSize: ContentSize = .fittingHeight
     public var alwaysBounceVertical = false
     public var alwaysBounceHorizontal = false
+    
+    /**
+     How much the content of the `ScrollView` should be inset.
+     
+     Note: When `keyboardAdjustmentMode` is used, it will also adjust
+     the on-screen `UIScrollView`s `contentInset.bottom` to make space for the keyboard.
+     */
```
watt: 👍 for comments!

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,178 @@
...
+        
+        self.latestNotification = new
+        
+        /// Only communicate a frame change to the delegate if the frame actually changed.
+        
+        if let old = old {
+            guard old.endingFrame != new.endingFrame else {
+                return
+            }
+        }
```
watt: `guard` nested in `if` is a real brain-teaser. Suggest flattening to simplify:

```suggestion
        if let old = old, old.endingFrame == new.endingFrame {
            return
        }
```

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,178 @@
...
+    )
+}
+
+/**
+ Implementation borrowed from Listable:
+ https://github.com/kyleve/Listable/blob/master/Listable/Sources/Internal/KeyboardObserver.swift
+ 
+ iOS Docs for keyboard management:
+ https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html
+ */
```
watt: These are not the most helpful docs for understanding what this class does.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,178 @@
...
+    // MARK: Handling Changes
+    //
+    
+    enum KeyboardFrame : Equatable {
+        
+        /// The current frame does not overlap the current view at all.
+        case nonOverlapping
+        
+        /// The current frame does overlap the view, by the provided rect, in the view's coordinate space.
+        case visible(frame: CGRect)
```
watt: Should this case be named `overlapping` to match the other case?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -260,8 +300,137 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+            
+            return {
+                switch keyboardFrame {
+                case .nonOverlapping:
+                    return 0.0
+                    
+                case .visible(let frame):
+                    return self.bounds.size.height - frame.origin.y
+                }
+            }()
```
watt: I think you can remove this closure wrapper?

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,178 @@
...
+    //
+    // MARK: Initialization
+    //
+    
+    init(center : NotificationCenter = .default) {
+        
+        self.center = center
+        
+        /// We need to listen to both `will` and `keyboardDidChangeFrame` notifications. Why?
+        /// When dealing with an undocked or floating keyboard, moving the keyboard
```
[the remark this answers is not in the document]
kyleve: I went back and forth on this... there's four forms of keyboards:

1) Docked, takes up full width
2) Undocked, takes up full width
3) Also undocked, split, also semantically is full width
4) Floating on iPad, does not take up full width

I figured we generally want this for the first three cases, so kept it simple for now. I'll follow up with something for #4, to not inset in this case?

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,178 @@
...
+            self.receivedUpdatedKeyboardInfo(info)
+        } catch {
+            assertionFailure("Blueprint could not read system keyboard notification. This error needs to be fixed in Blueprint. Error: \(error)")
+        }
+    }
+}
+
+extension KeyboardObserver
+{
+    struct NotificationInfo : Equatable {
```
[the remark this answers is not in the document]
kyleve: Needs to be internal for testing access.

BlueprintUICommonControls/Tests/Sources/Internal/KeyboardObserverTests.swift
```diff
@@ -0,0 +1,182 @@
...
+    
+    func testcase(_ name : String = "", _ block : () throws -> ()) {
+        do {
+            try block()
+        } catch {
+            XCTFail("Unexpected error thrown: \(error)")
+        }
+    }
+    
+    func assertThrowsError(test : () throws -> (), verify : (Error) -> ()) {
```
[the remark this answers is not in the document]
kyleve: Whoops

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -260,8 +300,137 @@ fileprivate final class ScrollerWrapperView: UIView {
...
+        UIView.animate(withDuration: animationDuration, delay: 0.0, options: options, animations: {
+            self.updateBottomContentInsetWithKeyboardFrame()
+        })
+    }
+}
+
+
+private extension UIView {
+    
+    var bp_safeAreaInsets : UIEdgeInsets {
```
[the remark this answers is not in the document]
kyleve: I'd rather keep this as is... It's clear to developers from patterns what this means.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Static since it makes testing easier – just pass in the needed values and verify the output.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Old comment, removin

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Yep

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: I went back and forth on this... Started with the individual values, and then went back to this since these types seem inherently tightly coupled, and it make the code easier to read/write, IMO.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Doh, will do

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Oops, yep

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: Committing locally

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: 👍🏼

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
kyleve: Added some more color, but figured since this was an internal type, it didn't matter that much.

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Need to dig into this more... there's a bit of nuanced behaviour between the logic in `applyContentInset` handling changes only from the model updates, vs `updateBottomContentInsetWithKeyboardFrame` always updating on live views.

kyleve: Going to leave as is for now and come back to this one.

BlueprintUICommonControls.podspec
kyleve: Removed

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: Renamed to `calculateContentInset`.

BlueprintUICommonControls/Sources/Internal/KeyboardObserver.swift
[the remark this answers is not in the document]
kyleve: https://github.com/square/Blueprint/issues/71

BlueprintUICommonControls/Tests/ScrollViewTests.swift
```diff
@@ -0,0 +1,67 @@
...
+import XCTest
+
+import BlueprintUI
+
+@testable import BlueprintUICommonControls
+
+
+class ScrollViewTests : XCTestCase {
+    
+    func test_contentInset() {
```
kyleve: Remove this

BlueprintUICommonControls/Tests/ScrollViewTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        let view = BlueprintView()
+    }
+    
+    func test_finalContentInset()
```
kyleve: Name wrong now

PR #56, opened 2020-02-21
Element Diffing Updates

kyleve:
Note: This PR also bumps to iOS 10 as min deployment target, since SPOS has dropped it.

This PR updates how elements are identified, to make view reuse more resilient in many cases (eg, less tearing down and re-creation of an identical view).

We do this by identifying the count of a given _type and key_ of element in a hierarchy, not just indexing on the total number of elements in a hierarchy. For example, in the imaginary hierarchy below, we would have had the following `ElementIdentifiers` / `Components`.

```
ElementA -> ElementIdentifier(0 or [optional key])
ElementB -> ElementIdentifier(1 or [optional key])
ElementB -> ElementIdentifier(2 or [optional key])
ElementC -> ElementIdentifier(3 or [optional key])
```

In this PR, we now do:

```
ElementA -> ElementIdentifier(ElementA.self, 1,  [optional key])
ElementB -> ElementIdentifier(ElementB.self, 1, [optional key])
ElementB -> ElementIdentifier(ElementB.self, 2, [optional key])
ElementC -> ElementIdentifier(ElementC.self, 1, [optional key])
```

So if you were to remove A, or one of the Bs, the final `ElementIdentifier` instances for the latter elements do not change:

```
ElementB -> ElementIdentifier(ElementB.self, 1, [optional key])
ElementC -> ElementIdentifier(ElementC.self, 1, [optional key])
```

Alongside this, I was also able to simplify away `Storage.Component` in favor of only `ElementIdentifier`.

I also changed keying to `AnyHashable` instead of String to be more flexible. Swift automatically bridges any `Hashable` type to `AnyHashable` at the callsite, so no external code changes should be needed for anyone using the keying API.

What does this mean in practice?

Before this fix, if you had this view:

```
Row {
   $0.add(thing)
   $0.add(button)
}
```

And then removed `thing`

```
Row {
   $0.add(button)
}
```

`button` would be removed and re-created in the view hierarchy, because it did not have an explicit key. After this fix, it's just updated in place, even with no provided key. I have added a test to verify this.


SampleApp/Podfile
```diff
@@ -1,6 +1,4 @@
 platform :ios, '12.0'
-inhibit_all_warnings!
-use_frameworks!
```
kyleve: This fixes a weird Cocoapods thing I keep running into locally, where it's not putting the re-built framework into the simulator, so it crashes.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -134,19 +134,23 @@ extension ElementContent.Builder: ContentStorage {
 
         var result: [(identifier: ElementIdentifier, node: LayoutResultNode)] = []
         result.reserveCapacity(children.count)
+        
+        let identifierFactory = ElementIdentifier.Factory()
```
kyleve: Reserve capacity here?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -134,19 +134,23 @@ extension ElementContent.Builder: ContentStorage {
...
 
         for index in 0..<children.count {
-            let currentChildLayoutAttributes = childAttributes[childAttributes.startIndex.advanced(by: index)]
```
kyleve: I don't know why this was written this way before.. Changed...

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,49 @@
...
-    /// of children is used instead.
-    case index(Int)
-
-    internal init(key: String?, index: Int) {
-        if let key = key {
-            self = .key(key)
-        } else {
-            self = .index(index)
-        }
+/*
```
kyleve: Finish comment

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,49 @@
...
-    var description: String {
-        switch self {
-        case let .key(string):
-            return string
-        case let .index(index):
-            return "(\(index))"
+    
+    final class Factory
+    {
+        var countsByElementType : [ObjectIdentifier:Int] = [:]
```
kyleve: private

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,49 @@
...
-        }
+/*
+ */
+struct ElementIdentifier: Hashable, CustomDebugStringConvertible {
+    
+    let elementType : ObjectIdentifier
+    
+    let count : Int
+    
+    let key : AnyHashable?
```
kyleve: Is having this key still useful?

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,49 @@
...
-extension ElementIdentifier: CustomStringConvertible {
-
-    var description: String {
-        switch self {
-        case let .key(string):
-            return string
-        case let .index(index):
-            return "(\(index))"
+    
+    final class Factory
```
kyleve: Likely also want to group by key here.

kyleve: Done!

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -142,6 +143,8 @@ public final class BlueprintView: UIView {
 
 extension BlueprintView {
     
+    final class RootNativeNodeView : UIView { }
```
kyleve: Add a comment to this for people who find their way to it.

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,110 @@
...
-
-    var description: String {
-        switch self {
-        case let .key(string):
-            return string
-        case let .index(index):
-            return "(\(index))"
+    
+    var debugDescription: String {
+        "\(self.elementType).\(String(describing: self.key)).\(self.count)"
```
kyleve: Do an `if let` on the key to simplify this a bit when printed.

BlueprintUI/Tests/BlueprintViewAdditions.swift
```diff
@@ -0,0 +1,30 @@
...
+/**
+ Testing additions to help verify native view construction and management.
+ */
+extension BlueprintView
+{
+    internal var rootNativeElementViews : [UIView] {
+        self.rootNativeNodeView?.subviews ?? []
+    }
+    
+    private var rootNativeNodeView : RootNativeNodeView? {
```
kyleve: Make this hard fail – we control this, so it should not return a optional if we can't find the view.

kyleve: Actually, wait, no, this doesn't make sense – if the layout hasn't happened yet, there's no root note, so it could be nil. Leave as is.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -134,19 +134,23 @@ extension ElementContent.Builder: ContentStorage {
...
                 layoutAttributes: currentChildLayoutAttributes,
-                content: currentChild.content)
-
-            let identifier = ElementIdentifier(
-                key: currentChild.key,
-                index: index)
+                content: currentChild.content
+            )
+            
+            let identifier = identifierFactory.identifier(
```
kyleve: For readers: This is where the valuable change which actually improves view diffing happens.

BlueprintUI/Tests/ViewDiffingBehaviourTests.swift
```diff
@@ -0,0 +1,122 @@
...
+//
+//  Created by Kyle Van Essen on 2/20/20.
+//
+
+import Foundation
+import XCTest
+
+@testable import BlueprintUI
+
+class ViewDiffingBehaviourTests : XCTestCase
```
kyleve: Add something to this test that also verifies with applied keys. This is pretty nuanced behaviour, so we should test this.

kyleve: Done

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
+            let count = self.count(for: type, key: key)
+            
+            return ElementIdentifier(
+                elementType: type,
+                key: key,
+                count: count
+            )
+        }
+        
+        private var countsByElementType : [Key:Int]
```
kyleve: `countsByKey`?

.travis.yml
```diff
@@ -8,16 +8,17 @@ os: osx
...
   include:
     - language: swift
-      name: "Cocoapods BlueprintUI (Xcode 10.2)"
+      name: "Cocoapods BlueprintUI (Xcode 11.2)"
       xcode_scheme: BlueprintUI-Unit-Tests
-      osx_image: xcode10.2
+      osx_image: xcode11.2
       xcode_workspace: SampleApp.xcworkspace
-      xcode_destination: platform=iOS Simulator,OS=9.3,name=iPhone 4s
+      xcode_destination: platform=iOS Simulator,OS=10.3,name=iPhone 4s
```
kyleve: Something in here is wrong, fix.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: @timdonnelly: In case im missing something, any context on this?

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
-        switch self {
-        case .key(let key):
-            return key
-        default:
-            return nil
+    
+    /**
+     Internal type used to create `ElementIdentifier` instances during view hierarchy updates.
+     */
+    final class Factory
```
timdonnelly: Why is this a class?

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
-extension ElementIdentifier: CustomStringConvertible {
-
-    var description: String {
-        switch self {
-        case let .key(string):
-            return string
-        case let .index(index):
-            return "(\(index))"
+                
+        func identifier(for type : Element.Type, key : AnyHashable?) -> ElementIdentifier
```
timdonnelly: `identifier(for:key:)` does *not* communicate the fact that this method modifies the state of this factory. Maybe `makeIdentifier` / `nextIdentifier` instead?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -52,7 +52,7 @@ public final class BlueprintView: UIView {
         
         rootController = NativeViewController(
             node: NativeViewNode(
-                content: UIView.describe() { _ in },
+                content: RootNativeNodeView.describe() { _ in },
```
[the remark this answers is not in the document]
timdonnelly: Can we move this to a separate PR? I'm not sure I understand why we're adding a new class here, and it seems unrelated to the diffing update/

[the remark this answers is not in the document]
kyleve: I did this to make writing this testing utility safer: https://github.com/square/Blueprint/pull/56/files#diff-31b5eeee8999bbbffb63a561e49ed03b

But can also remove this and assume there's only ever one subview.

BlueprintUI/Sources/Internal/ElementIdentifier.swift
kyleve: Went back and forth on this. In Listable, I moved it to be a class because it actually ended up effecting performance meaningfully, but for many thousands of items in a diff. Since we don't have that case here, can switch back.

kyleve: Pushed an update.

BlueprintUI/Sources/Internal/ElementIdentifier.swift
kyleve: Done

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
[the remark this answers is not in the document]
kyleve: Removed for now.

BlueprintUI/Sources/Element/ElementContent.swift
kylebshr: If `childAttributes` were a slice or something then using `startIndex` would make sense, but `Layout.layout(..)` always returns an array, so looks equivalent to me

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
-extension ElementIdentifier: CustomStringConvertible {
-
-    var description: String {
-        switch self {
-        case let .key(string):
-            return string
-        case let .index(index):
-            return "(\(index))"
+                
+        mutating func nextIdentifier(for type : Element.Type, key : AnyHashable?) -> ElementIdentifier
```
kylebshr: This type has a lot of stylistic differences from the rest of the project, i.e., spaces before colons and the return before brackets. Should we set up Swiftlint or something for Blueprint?

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
+    /**
+     Internal type used to create `ElementIdentifier` instances during view hierarchy updates.
+     */
+    struct Factory {
+        
+        init(reserveCapacity capacity: Int? = nil) {
+            
+            self.countsByKey = [:]
+            
+            if let capacity = capacity {
```
kylebshr: Dictionary has an init that takes in a minimum capacity, this init could be simplified to just `self.countsByKey = .init(minimumCapacity: capacity ?? 0)` (and you could also make `capacity` non-optional with a default of 0 to avoid the `??`)

kylebshr: Pretty much just style though, feel free to leave ¯\\\_(ツ)_/¯

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -1,43 +1,114 @@
...
+        mutating private func nextCount(for type : Element.Type, key : AnyHashable?) -> Int {
+            
+            let key = Key(
+                elementType: ObjectIdentifier(type),
+                key: key
+            )
+            
+            let current = self.countsByKey[key, default: 1]
+            
+            self.countsByKey[key] = (current + 1)
```
kylebshr: If I understand correctly, key counts only go up right? If an element is removed, the count doesn't decrease? If that's true, just wondering if we should worry about overflow.

kyleve: This is per element tree diff, so unless a element has 2^32 children (or whatever), this won't be a problem in practice.

kylebshr:
Oops, wasn’t thinking about that! Seems fine ofc

BlueprintUI/Sources/Internal/ElementIdentifier.swift
kyleve: Didn't know about `Dictionary(minimumCapacity: capacity)`; switching over!

PR #57, opened 2020-02-22
Allow ScrollView to dismiss the keyboard on tap

vocaro:
Tapping in a text field brings up the keyboard, but (on iPhone) there is no built-in way to dismiss it. This is a problem because the keyboard may cover up other UI elements, making them inaccessible.

This fixes the problem by adding a new property, shouldEndEditingOnTap, which defaults to false. Setting it to true changes the underlying UIScrollView’s keyboardDismissMode to .onDrag.

This change also demonstrates the new behavior in the SampleApp by adding a couple of text fields to it.

### DIFF
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -17,7 +17,7 @@ public struct ScrollView: Element {
     public var showsHorizontalScrollIndicator: Bool = true
     public var showsVerticalScrollIndicator: Bool = true
     public var pullToRefreshBehavior: PullToRefreshBehavior = .disabled
-    public var shouldEndEditingOnTap = false
+    public var keyboardDismissMode = UIScrollView.KeyboardDismissMode.none
 
     public init(wrapping element: Element) {
         self.wrappedElement = element
@@ -265,7 +265,7 @@ fileprivate final class ScrollerWrapperView: UIView {
         }
 
 
-        self.scrollView.keyboardDismissMode = scrollView.shouldEndEditingOnTap ? .onDrag : .none
+        self.scrollView.keyboardDismissMode = scrollView.keyboardDismissMode
     }
 
 }
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -17,6 +17,7 @@ public struct ScrollView: Element {
     public var showsHorizontalScrollIndicator: Bool = true
     public var showsVerticalScrollIndicator: Bool = true
     public var pullToRefreshBehavior: PullToRefreshBehavior = .disabled
+    public var shouldEndEditingOnTap = false
 
     public init(wrapping element: Element) {
         self.wrappedElement = element
@@ -264,6 +265,7 @@ fileprivate final class ScrollerWrapperView: UIView {
         }
 
 
+        self.scrollView.keyboardDismissMode = scrollView.shouldEndEditingOnTap ? .onDrag : .none
     }
 
 }
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -239,6 +239,10 @@ fileprivate final class ScrollerWrapperView: UIView {
             self.scrollView.showsHorizontalScrollIndicator = scrollView.showsHorizontalScrollIndicator
         }
 
+        if self.scrollView.keyboardDismissMode != scrollView.keyboardDismissMode {
+            self.scrollView.keyboardDismissMode = scrollView.keyboardDismissMode
+        }
+
         var contentInset = scrollView.contentInset
 
         if case .refreshing = scrollView.pullToRefreshBehavior, let refreshControl = refreshControl {
@@ -265,7 +269,6 @@ fileprivate final class ScrollerWrapperView: UIView {
         }
 
 
-        self.scrollView.keyboardDismissMode = scrollView.keyboardDismissMode
     }
 
 }
--- a/SampleApp/Sources/ViewController.swift
+++ b/SampleApp/Sources/ViewController.swift
@@ -28,7 +28,7 @@ let posts = [
 
 final class ViewController: UIViewController {
 
-    private let blueprintView = BlueprintView(element: List(posts: posts))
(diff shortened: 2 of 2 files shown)


vocaro:
Screenshot of SampleApp changes
Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-22 at 12 18 09

timdonnelly:
Why should scroll views have anything to do with keyboard/first responder status?

kyleve:
^ The above is why I originally recommended doing this through a wrapper element – it could get used by any element, including scroll view, or just a stack that has a text field in it, etc. Put this in Slack, but posting here as well for discussion:


```
struct TapToDismissKeyboard : Element
{
    init(onDismiss : @escaping () -> () = {}, wrapping : Element)
    {
        self.onDismiss = onDismiss
        self.wrapping = wrapping
    }
    var wrapping : Element
    var onDismiss : () -> ()
    var content: ElementContent {
        return ElementContent(child: self.wrapping)
    }
        ViewDescription(View.self) { config in
            config.builder = {
                View(frame: bounds)
            }
            config.apply {
                $0.onDismiss = self.onDismiss
            }
        }
    }
    final private class View : UIView, UIGestureRecognizerDelegate
    {
        var onDismiss : () -> () = {}
        let gestureRecognizer = UITapGestureRecognizer()
        override init(frame: CGRect)
        {
            super.init(frame: frame)
            self.gestureRecognizer.delegate = self
            self.gestureRecognizer.addTarget(self, action: #selector(didTapToDismiss))
            self.addGestureRecognizer(self.gestureRecognizer)
        }
        required init?(coder: NSCoder) {
            fatalError("init(coder:) has not been implemented")
        }
        @objc func didTapToDismiss()
        {
            if self.endEditing(true) {
                self.onDismiss()
            }
        }
        override func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool
        {
            guard gestureRecognizer == self.gestureRecognizer else {
                return super.gestureRecognizerShouldBegin(gestureRecognizer)
            }
            // Eventually may want to check for an active first responder here...
            return true
        }
    }
}
```

vocaro:
> Why should scroll views have anything to do with keyboard/first responder status?

They are intimately related. In UIKit, the usual way of handling the keyboard overlap problem is to add a gesture recognizer to UIScrollView, so exposing a property to do this automatically is simply a convenience for a common use case.

In fact, Apple added a property to do just that in iOS 7: [keyboardDismissMode](https://developer.apple.com/documentation/uikit/uiscrollview/1619437-keyboarddismissmode). I updated the pull request to use this property, and it is much simpler now.

If there is strong preference of doing it with a wrapper because it is more pure or whatever, I'm fine with that too. However, I think that approach is not as discoverable as this property is. I also have trouble thinking of a case where the wrapper would be wrapping anything other than a scroll view in practice.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -17,6 +17,7 @@ public struct ScrollView: Element {
     public var showsHorizontalScrollIndicator: Bool = true
     public var showsVerticalScrollIndicator: Bool = true
     public var pullToRefreshBehavior: PullToRefreshBehavior = .disabled
+    public var shouldEndEditingOnTap = false
```
kyleve: Would it make more sense to expose the keyboardDismissMode here? That would also allow callers to utilize `interactive` if they wish.

vocaro: I originally was going to do that but I figured Blueprint is trying to abstract away UIKit as much as possible. But then again that is just my assumption.

I tried out `.interactive` and it doesn't seem very useful.

But I'm open to either way.

kyleve: I think in this case, it makes sense to expose the entire enum: If someone needs or wants interactive dismissal (which makes sense – eg to mirror a UI similar to messages), we'd need to introduce a second property which overlaps with this one and change how `shouldEndEditingOnTap` is applied to check if interactive dismissal is enabled – which isnt as clean.

In terms of abstraction: Blueprint does generally abstract away the complexities of UIKit when it makes sense (eg, see `PullToRefreshBehavior` as a good example – reducing the complexities of defining PTR down to one enum, versus dealing with views and target/action callbacks). In this case, because the enum is a 1:1 passthrough, exposing it is fine.

timdonnelly:
> > Why should scroll views have anything to do with keyboard/first responder status?
> 
> They are intimately related. In UIKit, the usual way of handling the keyboard overlap problem is to add a gesture recognizer to UIScrollView, so exposing a property to do this automatically is simply a convenience for a common use case.
> 
> In fact, Apple added a property to do just that in iOS 7: [keyboardDismissMode](https://developer.apple.com/documentation/uikit/uiscrollview/1619437-keyboarddismissmode). I updated the pull request to use this property, and it is much simpler now.
> 
> If there is strong preference of doing it with a wrapper because it is more pure or whatever, I'm fine with that too. However, I think that approach is not as discoverable as this property is. I also have trouble thinking of a case where the wrapper would be wrapping anything other than a scroll view in practice.

Happen to use them together frequently != intimately related.

`keyboardDismissMode` involves the gestural behavior of the scroll view, which makes sense, but a scroll view should absolutely not add complexity like a tap recognizer to dismiss the keyboard for subviews.

Text fields are not always displayed within a scroll view. Scroll views don’t always contain text fields.

Can we rename the property to match the UIKit name? This is no longer related to a tap gesture.

BlueprintUICommonControls/Sources/ScrollView.swift
vocaro: Changed property type to UIScrollView.KeyboardDismissMode. It is now passing through to the underlying UIScrollView.

vocaro:
> Can we rename the property to match the UIKit name?

Yeah, done

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -264,6 +265,7 @@ fileprivate final class ScrollerWrapperView: UIView {
         }
 
 
+        self.scrollView.keyboardDismissMode = scrollView.keyboardDismissMode
```
kyleve: @timdonnelly: Do you remember why all the applications in this method do:

```
if self.scrollView.property != scrollView.property {
   self.scrollView.property = scrollView.property
}
```

?

PR #59, opened 2020-02-27
Implement intrinsicContentSize on BlueprintView

bencochran:
I mostly want then when making SwiftUI previews for BlueprintElements, providing a UIViewRepresentable around a BlueprintView. But it would also be useful for people putting Blueprint elements into a UI that uses auto layout.

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -39,8 +39,8 @@ public final class BlueprintView: UIView {
     /// The root element that is displayed within the view.
     public var element: Element? {
         didSet {
-            invalidateIntrinsicContentSize()
             setNeedsViewHierarchyUpdate()
+            invalidateIntrinsicContentSize()
         }
     }
 
@@ -92,7 +92,14 @@ public final class BlueprintView: UIView {
     public override var intrinsicContentSize: CGSize {
         guard let element = element else { return .zero }
         let constraint: SizeConstraint
-        if bounds.width == 0 {
+
+        // Use unconstrained when
+        // a) we need a view hierarchy update to force a loop through an
+        //    unconstrained width so we don’t end up “caching” the previous
+        //    element’s width
+        // b) the current width is zero, since constraining by zero is
+        //    nonsensical
+        if bounds.width == 0 || needsViewHierarchyUpdate {
             constraint = .unconstrained
         } else {
             constraint = SizeConstraint(width: bounds.width)
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -39,6 +39,7 @@ public final class BlueprintView: UIView {
     /// The root element that is displayed within the view.
     public var element: Element? {
         didSet {
+            invalidateIntrinsicContentSize()
             setNeedsViewHierarchyUpdate()
         }
     }
@@ -60,6 +61,8 @@ public final class BlueprintView: UIView {
         
         self.backgroundColor = .white
         addSubview(rootController.view)
+        setContentHuggingPriority(.defaultHigh, for: .horizontal)
+        setContentHuggingPriority(.defaultHigh, for: .vertical)
     }
 
     public override convenience init(frame: CGRect) {
@@ -83,9 +86,23 @@ public final class BlueprintView: UIView {
         }
         return element.content.measure(in: constraint)
     }
+
+    /// Returns the size of the element bound to the current width (mimicking
+    /// UILabel’s `intrinsicContentSize` behavior)
+    public override var intrinsicContentSize: CGSize {
+        guard let element = element else { return .zero }
+        let constraint: SizeConstraint
+        if bounds.width == 0 {
+            constraint = .unconstrained
+        } else {
+            constraint = SizeConstraint(width: bounds.width)
+        }
+        return element.content.measure(in: constraint)
+    }
     
     override public func layoutSubviews() {
         super.layoutSubviews()
+        invalidateIntrinsicContentSize()
         performUpdate()
     }
     


BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -83,6 +86,11 @@ public final class BlueprintView: UIView {
         }
         return element.content.measure(in: constraint)
     }
+
+    /// Returns the unconstrained element size
+    public override var intrinsicContentSize: CGSize {
+        return sizeThatFits(.zero)
```
kyleve: I always forget – is this suppose to be zero or infinite?

bencochran: BlueprintView [special-cases zero to be an unconstrained measure](https://github.com/square/Blueprint/blob/master/BlueprintUI/Sources/Blueprint%20View/BlueprintView.swift#L79-L80).

I wouldn’t personally trust throwing infinity at UI-ish things (I don’t see any special handling of that in Blueprint beyond assertions that we end up with finite results. So I’m guessing things would actually break if we did that.

kyleve: Replied to this before but then I fat fingered resolving this, so it got dropped.

I'm missing something, probably dumb: This means any measurement is going to be unconstrained in both directions, I think, which means eg if we're measuring a Stack, nothing will properly flow because the initial width measurement is going to end up being infinite. Is that OK? 

(This is a weird API anyway...)

bencochran: Yeah, that’s basically how `intrinsicContentSize` works. And I hate it.

kyleve: Wondering if it makes sense to implement this on `Element`, with a default value. I could see the default being `noIntrinsicMetric` for one dimension, with elements like stack deferring to children, or returning a default? Agree there's no right thing here.

kylebshr: This is definitely tricky - I did verify that this implementation doesn't work for Label (and probably stacks):

| Layout using Blueprint | Layout with Auto Layout |
|-|-|
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-02 at 12 05 12 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-02 at 12 02 52 |

Here's how I'm laying out the blueprint view:

```swift
    override func viewDidLoad() {
        super.viewDidLoad()

        view.backgroundColor = .white
        view.addSubview(blueprintView)
        blueprintView.translatesAutoresizingMaskIntoConstraints = false

        NSLayoutConstraint.activate([
            blueprintView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
            blueprintView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
            blueprintView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
        ])
    }
```

A halfway solution is to assume width is always constrained, `invalidateIntrinsicContentSize` in `layoutSubviews`, and use the views width to calculate the elements size with an unconstrained height. Ideally though, we would be able to tell which dimension (if not both) were fully constrained, but I don't know how to do that off the top of my head. This implementation at least works as expected for the above layout:

```swift
/// Returns the element size, constraining the width to the frame of the view.
public override var intrinsicContentSize: CGSize {
    guard let element = element else { return .zero }
    let constraint = SizeConstraint(width: frame.width)
    return element.content.measure(in: constraint)
}

override public func layoutSubviews() {
    super.layoutSubviews()
    performUpdate()
}
```

Not sure it's the best solution here, but might be more useful for prototyping?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -60,6 +61,8 @@ public final class BlueprintView: UIView {
         
         self.backgroundColor = .white
         addSubview(rootController.view)
+        setContentHuggingPriority(.defaultHigh, for: .horizontal)
+        setContentHuggingPriority(.defaultHigh, for: .vertical)
```
watt: I've never seen a view set this on itself. Isn't the CHP something that should come from outside as part of the layout that includes this view (i.e. the storyboard or code that applies constraints to this view)? Why change the defaults?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
watt: This looks like the right implementation to me. The contract of `intrinsicContentSize` is that it measures the unconstrained size of the content.

kylebshr: That is what the docs say, but `UILabel` et. al adjust their height based on their width when working with constraints, and do so by returning an `intrinsicContentSize` based on their width. Do you see another way to make a multiline blueprint label work with Auto Layout?

kylebshr: Verified with this snippet: 

```
view.addSubview(label)

label.text = "Hello World, this is some really long text! I hope it's okay."
label.numberOfLines = 0
label.translatesAutoresizingMaskIntoConstraints = false

print(label.intrinsicContentSize) // prints (430.6666666666667, 20.333333333333332)

NSLayoutConstraint.activate([
    label.widthAnchor.constraint(equalToConstant: 30),
    label.leadingAnchor.constraint(equalTo: view.leadingAnchor),
    label.topAnchor.constraint(equalTo: view.topAnchor),
])

view.layoutIfNeeded()

print(label.intrinsicContentSize) // prints (30.0, 385.6666666666667)
```

Classic case of Apple breaking their own rules. Not sure what the right decision is here though.

watt: I seem to recall that AutoLayout sets [preferredMaxLayoutWidth](https://developer.apple.com/documentation/uikit/uilabel/1620534-preferredmaxlayoutwidth) to get that behavior?

At the end of the day, I'm not sure it matters if a BlueprintView containing a UILabel has a different intrinsic size than a UILabel under AutoLayout, as long as we return an intrinsic size that makes sense for the content.

kylebshr:
Oh interesting, I didn’t realize that property was set automatically. 

> I'm not sure it matters if a BlueprintView containing a UILabel has a different intrinsic size than a UILabel under AutoLayout

It would matter in a case like self-sizing table view cells, or using a BlueprintView in a UIStackView, right?

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
kyleve: Love me some mushy UIKit (re preferredMaxLayoutWidth). That does sound right though.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
bencochran: Per the documentation:

> Custom views should set default values for both orientations on creation, based on their content, typically to `UILayoutPriorityDefaultLow` or `UILayoutPriorityDefaultHigh`.

Mostly though, without this, when used in a SwiftUI preview the `.layout(.sizeThatFits)` option gives you a screen-sized preview. And it makes sense to me that a BlueprintView would quite prefer to be the correct size for its contents.

bencochran:
Updated this with the suggestion from @kylebshr to mimic UILabel’s behavior here.

Issue #61, opened 2020-03-06
Proposal: SwiftUI-like EnvironmentValues


watt:
I've prototyped a potential implementation of an environmental context like SwiftUI's `@Environment` functionality. I'd like to get some feedback on this approach before opening a PR.

Like SwiftUI, it features an extensible, strongly-typed dictionary, where keys are types, and each key has a default value. Environment modifications in outer elements cascade to the contained elements at layout time without the two having any knowledge of each other.

Unlike SwiftUI, it does not use property wrappers, and is less magical. Instead there is a new protocol `ContextElement` that works just like `ProxyElement`, but instead of an `elementRepresentation` property it has a function:

```swift
public protocol ContextElement: Element {
    func elementRepresentation(in environment: Environment) -> Element
}
```

To accomplish this I added a new `ContentStorage` implementation for `ElementContent` that holds a closure, and defers building its child elements until layout time. By deferring until layout time, we can use the layout pass to cascade an `Environment` instance through the entire tree.

This strategy requires building children twice: once for measuring and once for layout. Leaf nodes might be built numerous times if there are elements upstream in the tree that measure their children for layout.

## Prototype

Because the environment is required to build children, and building children is required for measuring, this change affects all `Measurable`s. I tried two implementations.

https://github.com/square/Blueprint/tree/watt/environment-api-measurable

This branch changes `Measurable` and adds an `Environment` argument.

https://github.com/square/Blueprint/tree/watt/environment-api-nonmeasurable

This branch does not change `Measurable`. Some types simply no longer conform. `ElementContent` has a function like this instead:

```swift
func measurable(in environment: Environment) -> Measurable 
```

The second approach had a smaller impact, but I could see doing it either way.

## Example

[This commit](https://github.com/square/Blueprint/commit/755e242c57f9efec44b22b900593d7a9bcbecbad) contains an example use in the sample app.

kyleve:
Also similar: ViewEnvironment in Workflow: https://github.com/square/workflow/pull/999/files

watt:
Yeah, I believe they are similar ideas. ViewEnvironment cascades down the workflow tree, and this cascades down an element tree. You could use ViewEnvironment as the source of values to a Blueprint Environment.

kyleve:
Can we an open a PR in this repo for review + comment? I find this the easiest way to absorb and provide feedback – include WIP in the title so it's more obvious what it's for?

watt:
[...]

PR #63, opened 2020-03-06
Environment

watt:
## Overview

This is an implementation of a SwiftUI-like Environment for #61.

It introduces a way for elements to read from the environment, and a way to write changes that will cascade to child elements.

## Reading the environment

In SwiftUI, environment values are exposed through the `@Environment` property wrapper.

```swift
struct DaylightSavingLabel: View {
    @Environment(\.timeZone) var timeZone

    var body: some View {
        Text(timeZone.isDaylightSavingTime() ? "daylight saving time" : "standard time")
    }
}
```

SwiftUI makes this possible with lifecycle guarantees around the `body` property. The `@Environment` property is only guaranteed to be correct while `body` is being evaluated. It's a bit magical.

In Blueprint, it is difficult to make a similar lifecycle guarantee, because it is common to perform logic in element initializers, or to mutate an element before `elementRepresentation` (Blueprint's equivalent of `body`) is called.

Consider this pseudocode, which attempts to eagerly resolve the label text:

```swift
struct DaylightSavingLabel: ProxyElement {
    var text: String

    @Environment(\.timeZone) var timeZone

    init() {
        self.text = timeZone.isDaylightSavingTime() ? "daylight saving time" : "standard time"
    }

        return Label(text: text)
    }
}
```

This wouldn't work as expected, because the `timeZone` property cannot be guaranteed to be correct during `init`.

Instead of providing the environment through "magic" properties with lifecycle requirements, this implementation propagates the environment explicitly. To use it, you use a new protocol, `DynamicElement`.

`DynamicElement` is similar to `ProxyElement`, but the environment is explicitly provided as a parameter:

```swift
struct DaylightSavingLabel: DynamicElement {

    func elementRepresentation(in environment: Environment) -> Element {

        let text = environment.timeZone.isDaylightSavingTime()
            ? "daylight saving time"
            : "standard time"

        return Label(text: text)
    }
}
```

## Writing to the environment

SwiftUI uses view modifiers to "wrap" views with environment settings.

```swift
struct CaliforniaView: View {
    var body: some View {
        DaylightSavingLabel()
            .environment(\.timeZone, TimeZone(identifier: "America/Los_Angeles")!)
    }
}
```

This implementation follows a similar approach, with a wrapper named `AdaptedEnvironment`.

There is an initializer to apply a single key-value pair:

```swift
struct CaliforniaElement: ProxyElement {
        return AdaptedEnvironment(
            key: TimeZoneKey.self,
            value: TimeZone(identifier: "America/Los_Angeles")!
            wrapping: DaylightSavingLabel()
        )
    }
}
```

Or with a closure that lets you directly mutate the environment:

```swift
struct MyRootElement: ProxyElement {
    var safeAreaInsets: UIEdgeInsets
    var screenScale: CGFloat
    var timeZone: TimeZone
    var wrappedElement: Element

        return AdaptedEnvironment(
            by: { (environment) in
                environment.safeAreaInsets = self.safeAreaInsets
                environment.screenScale = self.screenScale
                environment.timeZone = self.timeZone
            },
            wrapping: wrappedElement
        )
    }
}
```

## Implementation details

A default `Environment` is created by `BlueprintView` before a render pass. It cascades through the tree during layout and measuring.

`DynamicElement` uses a new type of `ElementContent` storage to accomplish this. `LazyStorage` is a single-child storage that defers creation of the child until the environment is available.

Rather than holding a child Element, this storage has a closure of type `(Environment) -> Element`. When asked to perform layout, it calls the closure to create an element with the environment for this layout pass.

```swift
extension ElementContent {
    public init(build builder: @escaping (Environment) -> Element) {
        self.storage = LazyStorage(
            layout: PassthroughLayout(),
            builder: builder)
    }
}

extension DynamicElement {
        return ElementContent(build: self.elementRepresentation(in:))
    }
}
```

Another new storage type, `EnvironmentAdaptingStorage`, allows us to mutate the environment and propagate those changes to its child. It has 1 child and a closure of type `(inout Environment) -> Void`.

`AdaptedEnvironment` uses this storage:

```swift
extension ElementContent {
    public init(child: Element, environment environmentAdapter: @escaping (inout Environment) -> Void) {
        self.storage = EnvironmentAdaptingStorage(
            layout: PassthroughLayout(),
            adapter: environmentAdapter,
            child: child)
    }
}

public struct AdaptedEnvironment: Element {
    var wrappedElement: Element
    var environmentAdapter: (inout Environment) -> Void

    public init(
        by environmentAdapter: @escaping (inout Environment) -> Void,
        wrapping wrappedElement: Element)
    {
        self.wrappedElement = wrappedElement
        self.environmentAdapter = environmentAdapter
    }

    public init<K>(child: Element, key: K.Type, value: K.Value) where K: EnvironmentKey {
        self.init(child: child) { environment in
            environment[key] = value
        }
    }

        return ElementContent(child: wrappedElement, environment: environmentAdapter)
    }
}
```

### Repeated child creation

During a render pass `LazyStorage` may end up creating its descendent trees several times (likely at least twice, for measuring & layout), depending on how many times its parent asks it to measure. This could be a performance problem in very large, deep element hierarchies. There are some optimizations we could make to mitigate it if the need becomes apparent.

### 2-layer composition for environment in view-backed elements

To get environment values into a view-backed element, it must be wrapped in a `DynamicElement`. For example, this pair of elements propagates a property `foo` from the environment to a native view. The inner element cannot vary by itself — the outer element must actually create a different instance of `InnerElement` from the environment values.

```swift
struct OuterElement: DynamicElement {
    func elementRepresentation(in environment: Environment) -> Element {
        return InnerElement(foo: environment.foo)
    }
}

struct InnerElement {
    var foo: Foo

    init(foo: Foo) {
        self.foo
    }

        return NativeFooView.describe { config
            config[\.foo] = foo
        }
    }
}
```

This may be a little inconvenient, but it is fully compatible with existing elements. To use the environment directly in a view-backed element would require breaking changes to `backingViewDescription` and `Layout`.

## Example usage

There's an example in this PR in the sample app.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -154,30 +200,116 @@ extension ElementContent.Builder: ContentStorage {
...
+        environment: Environment)
+        -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
+    {
+        let environment = adapted(environment: environment)
+
+        let childAttributes = layout.layout(
+            size: attributes.bounds.size,
+            child: child.content.measurable(in: environment))
+
+        let identifier = ElementIdentifier(key: nil, index: 0)
```
kylebshr: Still not super familiar with identifiers, but are you sure there's no repercussions of this always being nil/0?

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -0,0 +1,30 @@
...
+public struct Environment {
+    public static let empty = Environment()
+
+    private init() { }
+
+    private var values: [ObjectIdentifier: Any] = [:]
+
+    public subscript<K>(key: K.Type) -> K.Value where K: EnvironmentKey {
+        get {
+            let objectId = ObjectIdentifier(key)
```
kylebshr: TIL about `ObjectIdentifier`, pretty neat

kylebshr:
LGTM with some minor questions! Great work on this, seems like it'll be super useful

BlueprintUI/Sources/Element/ComposedElement.swift
```diff
@@ -0,0 +1,22 @@
...
+/// Similar in purpose to `ProxyElement`, but the contents may change depending on the `Environment`.
+///
+/// - seealso: ProxyElement
+public protocol ComposedElement: Element {
+
+    /// Return the contents of this element in the given environment.
+    func elementRepresentation(in environment: Environment) -> Element
+}
+
+extension ComposedElement {
```
kyleve: I know it'd be a painful refactor, but maybe fine once: Rather than a new element, should we refactor existing base types to always take in the environment?

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -0,0 +1,30 @@
...
+
+            if let value = values[objectId] as? K.Value {
```
kyleve: I think should be

```
if let value = values[objectId] {
   return value as! K.Value
}
```

Since this somehow ending up as a the wrong type should be impossible from how this is implemented, so we'd want to fail out.

BlueprintUI/Sources/Environment/Adapted.swift
```diff
@@ -0,0 +1,41 @@
+import CoreGraphics
+
+/// Wraps an element tree with a modified environment.
+///
+/// By specifying environmental values with this element, all child elements nested
+/// will automatically inherit those values automatically. Values can be changed
+/// anywhere in a sub-tree by inserting another `Adapted` element.
+public struct Adapted: Element {
```
kyleve: `AdaptedEnvironmentElement`? Not super clear from the name what this does or why

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -154,30 +200,116 @@ extension ElementContent.Builder: ContentStorage {
...
+        let node = LayoutResultNode(
+            element: child,
+            layoutAttributes: childAttributes,
+            content: child.content,
+            environment: environment)
+
+        return [(identifier, node)]
     }
 
+    func measure(in constraint: SizeConstraint, environment: Environment) -> CGSize {
```
kyleve: Unrelated to this PR, but I wonder if it's worth merging layout and measurement... Since we can derive measurement from layout, seems like the system could be made more efficient by merging them.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -154,30 +200,116 @@ extension ElementContent.Builder: ContentStorage {
...
+        attributes: LayoutAttributes,
+        environment: Environment)
+        -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
+    {
+        let child = buildChild(in: environment)
+        let childAttributes = layout.layout(
+            size: attributes.bounds.size,
+            child: child.content.measurable(in: environment))
+
+        let identifier = ElementIdentifier(key: nil, index: 0)
```
kyleve: Note; will require changes one we land https://github.com/square/Blueprint/pull/56

BlueprintUI/Sources/Element/ComposedElement.swift
kyleve: (Eg, in 6 months from now; more people will be using Blueprint – do we want them to have to deal with the separate types? Does it let us simplify anything in here to do this annoying refactor now?)

BlueprintUI/Sources/Element/ComposedElement.swift
```diff
@@ -0,0 +1,22 @@
...
+    public var content: ElementContent {
+        return ElementContent(build: self.elementRepresentation(in:))
+    }
+
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
```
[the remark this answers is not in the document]
kyleve: Can we file this off as a GH issue to come back to later? It's definitely an interesting problem / thought.

BlueprintUI/Sources/Element/ComposedElement.swift
```diff
@@ -0,0 +1,22 @@
+import CoreGraphics
+
+/// An element built out of other elements.
+///
+/// Similar in purpose to `ProxyElement`, but the contents may change depending on the `Environment`.
+///
+/// - seealso: ProxyElement
+public protocol ComposedElement: Element {
```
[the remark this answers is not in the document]
kyleve: +1

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -15,20 +15,43 @@ public struct ElementContent: Measurable {
...
-        return storage.performLayout(attributes: attributes)
+    func performLayout(attributes: LayoutAttributes, environment: Environment) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
+        return storage.performLayout(attributes: attributes, environment: environment)
     }
 
 }
 
+extension ElementContent {
+    func measurable(in environment: Environment) -> Measurable {
+        struct EnvironmentMeasurable: Measurable {
```
kyleve: Is the idea here that we have some measurables which do not conform to or have an environment? Assume this is to avoid having to do downstream refactoring?

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: This is fine because there's only ever one child, I believe.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -154,30 +200,116 @@ extension ElementContent.Builder: ContentStorage {
...
         var element: Element
 
-        func measure(in constraint: SizeConstraint) -> CGSize {
-            return content.measure(in: constraint)
-        }
+    }
+
+}
 
+private struct EnvironmentAdaptingStorage: ContentStorage {
```
kyleve: If we're going to make envs a core part of blueprint, I would prefer we pay the tax of moving this stuff onto ContentStorage itself, I think.

BlueprintUI/Sources/Element/ComposedElement.swift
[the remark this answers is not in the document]
kyleve: (A bit more context: With Workflow, we haven't been afraid to do large breaking refactors, which has been valuable to keep us moving forward – think we should take the same approach.)

BlueprintUI/Sources/Element/ComposedElement.swift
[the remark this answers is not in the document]
watt: Yeah, naming is hard. This has a very similar role to ProxyElement, so I tried to come up with a similar name, but maybe that's just confusing. If ProxyElement didn't exist, I think `ComposedElement` would be a good name.

I avoided `EnvironmentElement` because I think it's unclear whether that means it does something to the environment or consumes it or something else.

The main difference between this and ProxyElement is that this lets you dynamically create an element, while ProxyElement is static. Perhaps `DynamicProxyElement` is better?

BlueprintUI/Sources/Element/ComposedElement.swift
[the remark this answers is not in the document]
watt: This would basically mean reversing `Element` and `ProxyElement`. Might be an improvement, but IMO hard to justify unless it's accompanied by some functional improvements in addition to the naming.

BlueprintUI/Sources/Element/ElementContent.swift
watt: Yep, it's scoped to this element's single child, so literally any value would be fine.

BlueprintUI/Sources/Element/ComposedElement.swift
watt: If you don't need the environment, ProxyElement is potentially much more efficient. I don't think we should get rid of it until we can guarantee that ComposedElement is safe as a universal replacement. I would love to only have one eventually though.

BlueprintUI/Sources/Element/ElementContent.swift
watt: I've thought about that a bunch actually, and I think it would be great! But I haven't had time to experiment and see if there are any gotchas.

BlueprintUI/Sources/Element/ElementContent.swift
watt: Yep, since ElementContent cannot be measured without an environment now, this is an adapter for functions that still want a Measurable.

I had an alternative PR where I tried refactoring Measurable instead, and the result was (IMO) worse.

BlueprintUI/Sources/Element/ElementContent.swift
watt: Do you mean moving the environment adapter onto `ElementContent.Builder` (the existing `ContentStorage`)? I actually had it that way in an earlier version, and moved it onto this new type to clean things up.

BlueprintUI/Sources/Environment/Adapted.swift
watt: This is the element you use to actually put values in the environment. So the idea was that it _adapts_ the environment for all of its children.

BlueprintUI/Sources/Environment/Environment.swift
watt: Sure, shouldn't matter either way.

- [ ] force cast K.Value

kyleve:
I started writing up a long / confusing comment yesterday about another thought I had for how to go about implementing environment, and figured it'd be better if I just wrote it as actual code instead. I did that here, https://github.com/square/Blueprint/pull/69/files, with context in a PR description. Thoughts?

watt:
I rewrote the description of this PR to focus more on how it's used and less on the implementation details.

BlueprintUI/Sources/Element/ComposedElement.swift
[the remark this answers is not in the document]
watt: I wound up renaming this to `DynamicElement`, which I think distinguishes it better?

BlueprintUI/Sources/Environment/Adapted.swift
watt: Renamed to `AdaptedEnvironment`!

kyleve:
Finally taking a look at this..

BlueprintUI/Sources/Element/ElementContent.swift
kyleve, line 34: Might want to move this out of the func? Eg, not sure how this type appears in the debugger or stacks readability-wise.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -72,17 +95,34 @@ extension ElementContent {
...
+        self.storage = EnvironmentAdaptingStorage(
+            layout: PassthroughLayout(),
+            adapter: environmentAdapter,
+            child: child)
+    }
 
-fileprivate protocol ContentStorage: Measurable {
-    var childCount: Int { get }
-    func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
+    public init<K>(child: Element, key: K.Type, value: K.Value) where K: EnvironmentKey {
```
kyleve, line 108: Nit: Don't love the single letter generics. `Key`?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -158,30 +204,116 @@ extension ElementContent.Builder: ContentStorage {
...
+    }
+
+    private func adapted(environment: Environment) -> Environment {
+        var environment = environment
+        adapter(&environment)
+        return environment
+    }
 }
 
+/// Content storage that defers creation of its child until measurement or layout time.
```
kyleve, line 276: Thought: Do we just want to make this be the default? I guess because of double-creation?

kyleve: oh yeah, guess so – another reason merging these will be rad

BlueprintUI/Sources/Environment/AdaptedEnvironment.swift
```diff
@@ -0,0 +1,41 @@
+import CoreGraphics
+
+/// Wraps an element tree with a modified environment.
+///
+/// By specifying environmental values with this element, all child elements nested
+/// will automatically inherit those values automatically. Values can be changed
+/// anywhere in a sub-tree by inserting another `AdaptedEnvironment` element.
+public struct AdaptedEnvironment: Element {
+    var wrappedElement: Element
```
kyleve, line 9: Nit: Just "wrapped"?

BlueprintUI/Sources/Environment/AdaptedEnvironment.swift
```diff
@@ -0,0 +1,41 @@
+import CoreGraphics
+
+/// Wraps an element tree with a modified environment.
+///
+/// By specifying environmental values with this element, all child elements nested
+/// will automatically inherit those values automatically. Values can be changed
+/// anywhere in a sub-tree by inserting another `AdaptedEnvironment` element.
+public struct AdaptedEnvironment: Element {
```
kyleve, line 8: Should add an `extension Element` to match the SwiftUI-style stuff: (eg, https://github.com/square/Blueprint/blob/master/BlueprintUI/Sources/Layout/Aligned.swift#L118)

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+public protocol EnvironmentKey {
```
kyleve, line 3: Docs on these before merge, please!

BlueprintUI/Tests/Element+Testing.swift
```diff
@@ -0,0 +1,14 @@
...
+import QuartzCore
+
+extension Element {
+    /// Build a fully laid out element tree with complete layout attributes
+    /// for each element.
+    ///
+    /// - Parameter frame: The frame to assign to the root element.
+    ///
+    /// - Returns: A layout result
+    func layout(frame: CGRect) -> LayoutResultNode {
```
kyleve, line 11: Worth taking in an env here with a default of empty for other testing usages?

kyleve:
Some thoughts:

1. I don’t love the name DynamicElement, but don’t have a better idea right now. It makes me think it's stateful or will be updated without re-updating the element tree. Maybe EnvironmentElement?

1. Not being able to access within backingViewDescription kinda sucks, but I get that it happens due to measure/layout/view creation being separate passes. I would if we could cache the `Environments` for each `Element` by their `ElementPath`, and then use that `ElementPath` during view creation to fetch the correct environment for the element?

1. Would be nice to create some sensible defaults for the environment, plus expose the root env on the blueprint view. In my PR I’d added some useful defaults to the Environment, plus made the environment configurable through BlueprintView, as an example of this: https://github.com/square/Blueprint/pull/69/files#r426765496, https://github.com/square/Blueprint/pull/69/files#diff-467b658cffa263aa6a0c4bd277350b18

1. Names of the new ContentStorage types... Trying to think of ways to make the three a bit clearer. Maybe {Dynamic/Environment}ElementStorage, MutateEnvironmentStorage, StaticElementStorage?

1. Should add a SwiftUI-style Element extension for AdaptedEnvironment, as we did for other wrapper elements.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -15,20 +15,43 @@ public struct ElementContent: Measurable {
...
 
-    func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
-        return storage.performLayout(attributes: attributes)
+    func performLayout(attributes: LayoutAttributes, environment: Environment) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
+        return storage.performLayout(attributes: attributes, environment: environment)
     }
 
 }
 
+extension ElementContent {
```
kyleve, line 32: Should we also add a companion to some of the ElementContent extensions, like `public init(measureFunction: @escaping (SizeConstraint) -> CGSize)`, but that also pass the env? (Sorry if I missed that in here)

BlueprintUI/Sources/Element/ElementContent.swift
watt: Like how it demangles? Not sure. It looks fine in Xcode, though.

We use this pattern in the `measureFunction` init too. Seems like the right amount of encapsulation if it's not a type that's being used anywhere else.

BlueprintUI/Sources/Element/ElementContent.swift
watt: Sure

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: Yeah, or in the debugger / crash logs – that's why I've avoided this before, but maybe its better now (probably).

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: (Sorry, by "these", I meant merging layout and measurement)

BlueprintUI/Sources/Environment/AdaptedEnvironment.swift
kyleve: The general swift convention is to leave the type name off of the var (eg people instead of peopleArray), but this is also fine.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: Thinking of a case where you may want to vary the measurement based on something in the environment, but I guess we can add that later if it's needed.

kyleve:
> To use an environment value from a backing view description, you can put the value you want in a regular property, and wrap that element in a DynamicElement.

> The element where you extract from the environment is always at least 1 level above the element where you actually use the value.

Yes – realize this is how it's set up now, but I'm questioning if this is the ergonomics / hoops we want people to have to jump through – it definitely isnt obvious at first glance.

> Defaults probably shouldn't be set by BlueprintView unless they are values coming off the view itself. That's what the defaultValue of the key is for.

I am thinking there are a bunch of different sources you may have for environment values
1) A value set in the environment further up the element tree
1) A value set by where you are in a UIView tree – safe area insets, trait collections, or by the system itself – locale, screen scale, etc.
1) The default fallback value when no other value is available.
1) Values pushed into the blueprint view from the thing that's hosting it in a UIView tree – a view controller, etc. This is where I see the `environment` on the BlueprintView itself being useful.

> DynamicStorage, StaticStorage, EnvironmentMutatingStorage? (Might be better to bikeshed in a follow-up?)

Sure

PR #64, opened 2020-03-13
Round view frames to pixel boundaries

watt:
Another attempt at snapping views to pixel boundaries.

This approach is similar to what @timdonnelly talked about in #34, demo'd here: https://github.com/square/Blueprint/commit/55761962996c6a1b9731b8ea315f2e86064f4378

The trick has two parts:

1. Apply a correction to children to offset the amount the parent was rounded. This prevents "drift". For example, if an element is inset by 0.3 it rounds to 0.0, but if its child is also inset by 0.3, it is first corrected to 0.6 before rounding upwards to 1.0 (at scale 1.0).
2. Offset by the origin relative to the BlueprintView before rounding, and revert after rounding. This is to account for negative offsets from part 1 (otherwise a child at 0.0 corrected to -0.5 rounds to -1.0 rather than up to 0.0).

## Example renders

These were rendered on a screen scale of 2.0 (one pixel per 0.5 points) and then zoomed for detail.

### Nested boxes

These boxes have integral vertical insets and fractional horizontal insets, which you can see snapping together.

```swift
struct NestedBoxes: ProxyElement {
    let colors: [UIColor] = [
        .cyan,
        .red,
        .green,
        .yellow,
        .blue,
        .magenta,
        .purple,
        .orange
    ]
    var depth: Int = 10

        guard depth > 0 else {
            return Box(backgroundColor: .white, wrapping: Spacer(size: CGSize(width: 10, height: 10)))
        }

        let color = colors[depth % colors.count]
        return Box(
            backgroundColor: color,
            wrapping: Inset(
                top: 2.0,
                bottom: 2.0,
                left: 0.3,
                right: 0.3,
                wrapping: NestedBoxes(depth: depth - 1)
            )
        )
    }
}
```

Screen Shot 2020-03-12 at 4 34 31 PM

### Row spacing

In this example 5 boxes of 2 points each are set in a row 13 points wide, with the `spaceEvenly` underflow distribution. That leaves 3 points of gutter space must be split into 4 parts. This results in gutters of 1.0, 0.5, 1.0, and 0.5.

```swift
struct FractionRow: ProxyElement {
        return Box(
            backgroundColor: .yellow,
            wrapping: ConstrainedSize(
                width: .absolute(13.0),
                wrapping: Row { row in
                    row.horizontalUnderflow = .spaceEvenly
                    for _ in 1...5 {
                        row.add(
                            child: Box(
                                backgroundColor: .red,
                                wrapping: Spacer(size: CGSize(width: 2.0, height: 5.0))
                            )
                        )
                    }
                }
            )
        )
    }
}
```

Screen Shot 2020-03-12 at 4 49 36 PM


BlueprintUI/Sources/Internal/NativeViewNode.swift
```diff
@@ -39,5 +39,29 @@ struct NativeViewNode {
...
+    /// Recursively rounds this node's layout frame and all its children to snap to pixel boundaries.
+    ///
+    /// - Parameters:
+    ///   - origin: The global origin to offset the frame by before rounding. This offset is used to ensure that
+    ///     positive and negative frame coordinates both round away from zero.
+    ///   - correction: The amount of rounding correction to apply to the origin before rounding, to account for the
+    ///     rounding applied to this node's parent.
+    ///   - scale: The screen scale to use when rounding.
+    mutating func round(from origin: CGPoint, correction: CGPoint, scale: CGFloat) {
+        guard CATransform3DIsIdentity(layoutAttributes.transform) else {
```
kylebshr: We should document why this check exists

watt:
Fixes #70

kyleve:
Getting to this today, sorry for delay

BlueprintUI/Sources/Layout/LayoutAttributes.swift
```diff
@@ -153,7 +153,34 @@ public struct LayoutAttributes {
...
+
+    /// Performs rounding on the frame to snap to pixel boundaries.
+    ///
+    /// - Parameters:
+    ///   - origin: The global origin to offset the frame by before rounding. This offset is used to ensure that
+    ///     positive and negative frame coordinates both round away from zero.
+    ///   - correction: The amount of rounding correction to apply to the origin before rounding, to account for the
+    ///     rounding applied to this node's parent.
+    ///   - scale: The screen scale to use when rounding.
+    mutating func round(from origin: CGPoint, correction: CGPoint, scale: CGFloat) -> CGPoint {
```
kyleve, line 165: Is it worth making `RoundingCorrection` its own type that wraps a `CGPoint`? Seeing all these CGPoints and CGFloats makes it a bit hard to follow what's going on, IMO.

watt: I could see that if it were publicly exposed, but as an internal thing I think the documentation is probably sufficient.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -141,11 +141,14 @@ public final class BlueprintView: UIView {
...
         rootController.view.frame = bounds
         
-        let rootNode = NativeViewNode(
+        var rootNode = NativeViewNode(
             content: UIView.describe() { _ in },
             layoutAttributes: LayoutAttributes(frame: bounds),
             children: viewNodes
         )
+
+        let scale = window?.screen.scale ?? UIScreen.main.scale
```
bencochran, line 150: Should we do `setNeedsViewHierarchyUpdate` in `didMoveToWindow`? (a view moving from a window at one scale to a window at another is _entirely_ an edge case, but…)

watt: Yeah, I suppose. (If that did happen we would also have problems with elements that do measuring with prototype view instances.)

PR #67, opened 2020-03-18
Xcode 11 and Swift 5.1 support

watt:
This updates the Travis config to build with Xcode 11, which means Swift 5.1 features are now supported.

Unfortunately, this required a kludge to get tests working on 32-bit simulators. For some reason, the builtin `swift-stdlib-tool` in Xcode 11 does not copy all the swift stdlibs into the xctest bundle. This PR contains a workaround that just adds the stdlib directory to the runtime search paths instead of copying them.


watt:
Update: I found a more elegant solution for the missing dylibs that also passes pod linting. Still unclear why this became an issue in Xcode 11 though!

PR #68, opened 2020-03-20
Measure stack axes separately

### LINKED ISSUE #58: Stack layout should measure cross axis after layout axis has been allotted
Stack layout currently measures its children only once, using the stack's entire bounds as a constraint. These basis sizes are then used for the cross axis sizes, while the layout axis sizes are adjusted according to the child priorities.

For elements like a `Label` with text wrapping, where the height depends on the width, this can result in undesirable cross axis measurements.

Here's an example:

```swift
struct TestElement: ProxyElement {
        return Column { column in
            column.verticalUnderflow = .justifyToCenter
            column.horizontalAlignment = .fill

            column.add(child: row(withMargin: 0))
            column.add(child: row(withMargin: 50))
        }
    }

    func row(withMargin margin: CGFloat) -> Element {
        return Row { row in
            row.horizontalOverflow = .condenseProportionally

            row.add(
                growPriority: 0,
                shrinkPriority: 0,
                child: Spacer(size: CGSize(width: margin, height: 0)))

            row.add(
                growPriority: 1,
                shrinkPriority: 1,
                child: Label(text: "This is a long label for testing. It takes up 2 lines if there are no margins, but needs 3 if we add some margins."))

            row.add(
                growPriority: 0,
                shrinkPriority: 0,
                child: Spacer(size: CGSize(width: margin, height: 0)))
        }
    }
}
```

Screen Shot 2020-02-25 at 12 03 54 PM

In the second instance, the label should ideally be wrapped onto 3 lines. However, its height measurement is the same as the first instance, and does not take into account the adjusted width. It ends up truncated.

To fix this, I think we should re-measure children after determining the layout axis, and use those measurements for the cross axis sizes, rather than the basis sizes.

watt:
Fixes #58.

Note: Stack layout is generalized for rows & columns, and the terms "axis" and "cross" are used extensively to symbolically represent the axis under layout and the other axis. I have added a comment explaining this [term of art](https://swift.org/documentation/api-design-guidelines/#use-terminology-well).

This PR changes stack layout so that it measures children twice, to accommodate flexible elements (like text) that may change their cross axis measurement when the layout axis is constrained.

The old layout logic was something like this:
1. measure children constrained by the stack's bounds, call these the "basis sizes"
2. distribute space along axis according to basis sizes
3. align along cross axis according to basis sizes

The new logic defers calculation of cross axis values until the axis has been allocated:
1. measure basis sizes (same)
2. distribute space according to basis sizes (same)
3. measure again, constraining each child by the space it was allocated in step 2
4. align along cross axis according to size from step 3
5. merge axis & cross values into frame

Measuring a stack also uses this behavior now when the measurement is constrained.

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -345,7 +345,7 @@ extension StackLayout {
         let totalBasisSize: CGFloat = basisSizes.reduce(0.0, +)
         let totalSpacing = minimumSpacing * CGFloat(basisSizes.count-1)
 
-        /// The size that will be distributed among children (can be positive or negative)
+        /// The overflow size that will be distributed among children
         let extraSize: CGFloat = layoutSize - (totalBasisSize + totalSpacing)
 
         assert(extraSize <= 0.0)
@@ -397,6 +397,7 @@ extension StackLayout {
         let totalBasisSize: CGFloat = basisSizes.reduce(0.0, +)
 
         let minimumTotalSpace = minimumSpacing * CGFloat(basisSizes.count-1)
+        /// The underflow size that will be distributed among children
         let extraSize: CGFloat = layoutSize - (totalBasisSize + minimumTotalSpace)
         
         assert(extraSize >= 0.0)
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -179,69 +179,112 @@ extension StackLayout {
 
 }
 
+// MARK: - Layout logic
+
+// Stack layout is generalized to be independent of the stack's axis (which is `horizontal` for rows and `vertical`
+// for columns).
+//
+// This terminology is used to symbolically represent the two axes:
+//
+//   - The axis along which elements are being laid out is generally called "axis".
+//   - The other axis is the cross axis, or often just "cross".
+//
 extension StackLayout {
 
-    fileprivate func _layout(size: CGSize, items: [(traits: Traits, content: Measurable)]) -> [LayoutAttributes] {
-
+    private func _layout(size: CGSize, items: [(traits: Traits, content: Measurable)]) -> [LayoutAttributes] {
         guard items.count > 0 else { return [] }
 
-        let constraint = SizeConstraint(size)
+        let vectorConstraint = SizeConstraint(size).vectorConstraint(on: axis)
+        let frames = _frames(for: items, in: vectorConstraint)
 
-        let layoutSize = size.stackVector(axis: axis)
-
-        let basisSizes = _getBasisSizes(constraint: constraint, items: items.map { $0.content })
-
-        let totalMeasuredAxis: CGFloat = basisSizes.map({ $0.axis }).reduce(0.0, +)
-        let minimumTotalSpacing = CGFloat(items.count-1) * minimumSpacing
-
-        let frames: [Frame]
-
-        /// Determine if we are dealing with overflow or underflow
-        if totalMeasuredAxis + minimumTotalSpacing >= layoutSize.axis {
-            /// Overflow
-            frames = _layoutOverflow(basisSizes: basisSizes, traits: items.map { $0.traits }, layoutSize: layoutSize)
-        } else {
-            /// Underflow
-            frames = _layoutUnderflow(basisSizes: basisSizes, traits: items.map { $0.traits }, layoutSize: layoutSize)
+        return frames.map { frame in
+            return LayoutAttributes(frame: frame.rect(axis: axis))
         }
-
-        return frames.map({ (frame) -> LayoutAttributes in
-            let rect = frame.rect(axis: axis)
-            return LayoutAttributes(frame: rect)
-        })
     }
 
-    fileprivate func _measureIn(constraint: SizeConstraint, items: [(traits: Traits, content: Measurable)]) -> CGSize {
+    private func _measureIn(constraint: SizeConstraint, items: [(traits: Traits, content: Measurable)]) -> CGSize {
+        guard items.count > 0 else { return .zero }
 
-        guard items.count > 0 else {
-            return .zero
+        let vectorConstraint = constraint.vectorConstraint(on: axis)
+
+        let frames = _frames(for: items, in: vectorConstraint)
+
+        let vector = frames.reduce(Vector.zero) { (vector, frame) -> Vector in
+            return Vector(
+                axis: max(vector.axis, frame.maxAxis),
+                cross: max(vector.cross, frame.maxCross))
         }
 
-        var result = Vector.zero
+        return vector.size(axis: axis)
+    }
 
-        for item in items {
-            let measuredSize = item.content.measure(in: constraint).stackVector(axis: axis)
-            result.axis += measuredSize.axis
-            result.cross = max(result.cross, measuredSize.cross)
+    private func _frames(
+        for items: [(traits: Traits, content: Measurable)],
+        in vectorConstraint: VectorConstraint
+    ) -> [VectorFrame] {
+        // First allocate available space along the layout axis.
+        let axisSegments = _axisSegments(for: items, in: vectorConstraint)
+
+        // Then measure cross axis for each item based on the space it was allocated.
+        let crossSegments = _crossSegments(
+            for: items.map { $0.content},
+            axisConstraints: axisSegments.map { $0.magnitude },
+            crossConstraint: vectorConstraint.cross)
+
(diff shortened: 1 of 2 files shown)


BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -179,69 +179,112 @@ extension StackLayout {
...
 }
 
+// MARK: - Layout logic
+
+// Stack layout is generalized to be independent of the stack's axis (which is `horizontal` for rows and `vertical`
+// for columns).
+//
+// This terminology is used to symbolically represent the two axes:
+//
+//   - The axis along which elements are being laid out is generally called "axis".
```
kyleve: Minor, but it would be good to include examples of this in practice (eg, what is axis and cross in a row vs. column). I still have to think about it every damn time...

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -179,69 +179,112 @@ extension StackLayout {
...
+            for: items.map { $0.content},
+            axisConstraints: axisSegments.map { $0.magnitude },
+            crossConstraint: vectorConstraint.cross)
+
+        // Finally, merge axis and cross segments into frames.
+        return zip(axisSegments, crossSegments).map(VectorFrame.init(axis:cross:))
+    }
 
-        var result = Vector.zero
+    private func _axisSegments(
```
kylebshr:
```suggestion
    /// This function calculates the position and length of each element along the axis of the stack,
    /// based on the constraints and configuration of the stack. 
    /// 
    /// In the case of a Row stack, for example, the axis is measuring the height as illustrated below:
    ///
    /// +-------------------------------+
    /// |                               |
    /// |   Stack                       |
    /// |                               |
    /// |  +----------------+           |
    /// |  | Element        |           |
    /// |  |                |           |
    /// |  +----------------+           |
    /// |                               |
    /// |  +----------------------+     |
    /// |  | Element              |     |
    /// |  +----------------------+     |
    /// |                               |
    /// |  +------------------+         |
    /// |  | Element      ∧   |         |
    /// |  |              |   |         |
    /// |  |              |   |         |
    /// |  |            Axis  |         |
    /// |  |              |   |         |
    /// |  |              |   |         |
    /// |  |              ∨   |         |
    /// |  +------------------+         |
    /// |                               |
    /// +-------------------------------+
    /// 
    private func _axisSegments(
```

kylebshr: This looks better in an IDE, GH has a lot of space in between lines

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -368,49 +407,86 @@ extension StackLayout {
...
 
-        return frames
-    }
+        // Then pick the max cross value based on the constraint
+        let maxCross: CGFloat
+        switch crossConstraint {
+        case .unconstrained:
+            maxCross = crossMagnitudes.reduce(0, max)
+        case .atMost(let max):
+            maxCross = max
```
kyleve: Should this be a min of the max and the actual height?

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -368,49 +407,86 @@ extension StackLayout {
...
+            let magnitude: CGFloat
 
-    fileprivate func _calculateCross(basisSizes: [Vector], layoutSize: Vector) -> [Frame] {
-        return basisSizes.map { (measuredSize) -> Frame in
-            var result = Frame.zero
             switch alignment {
             case .center:
-                result.origin.cross = (layoutSize.cross - measuredSize.cross) / 2.0
-                result.size.cross = measuredSize.cross
+                origin = (maxCross - measuredCross) / 2.0
```
kyleve: It didnt have it before, but should this be rounded?

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -368,49 +407,86 @@ extension StackLayout {
...
-            let basis = basisSizes[index]
+            return Segment(origin: origin, magnitude: magnitude)
+        }
 
-            frames[index].size.axis = basis.axis + ((growPriorities[index] / totalPriority) * extraSize)
+        return axisSegments
+    }
 
-            axisOrigin = frames[index].maxAxis + space
+    private func _crossSegments(
```
kylebshr:
```suggestion
    /// This function calculates the position and length of each element along the cross of the stack,
    /// based on the given axis constraint and configuration of the stack.
    ///
    /// In the case of a Row stack, for example, the cross is measuring the width as illustrated below.
    /// 
    /// +-------------------------------+
    /// |                               |
    /// |   Stack                       |
    /// |                               |
    /// |  +----------------+           |
    /// |  | Element        |           |
    /// |  |                |           |
    /// |  +----------------+           |
    /// |                               |
    /// |  +----------------------+     |
    /// |  | Element              |     |
    /// |  +----------------------+     |
    /// |                               |
    /// |  +------------------+         |
    /// |  | Element          |         |
    /// |  |                  |         |
    /// |  |                  |         |
    /// |  |                  |         |
    /// |  |                  |         |
    /// |  |                  |         |
    /// |  |<---- Cross ----> |         |
    /// |  +------------------+         |
    /// |                               |
    /// +-------------------------------+
    /// 
    private func _crossSegments(
```

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -368,49 +407,86 @@ extension StackLayout {
...
 
-    fileprivate func _calculateCross(basisSizes: [Vector], layoutSize: Vector) -> [Frame] {
-        return basisSizes.map { (measuredSize) -> Frame in
-            var result = Frame.zero
             switch alignment {
             case .center:
-                result.origin.cross = (layoutSize.cross - measuredSize.cross) / 2.0
-                result.size.cross = measuredSize.cross
+                origin = (maxCross - measuredCross) / 2.0
+                magnitude = measuredCross
```
kyleve: Got to this line, will continue later.

BlueprintUI/Sources/Layout/Stack.swift
kylebshr: I just added some suggestions with ascii art, maybe those should be moved/copied up to here as well.

watt: Will do.
- [ ] add ascii art docs for axis/cross

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -179,69 +179,112 @@ extension StackLayout {
...
 
-        let totalBasisSize: CGFloat = basisSizes.map({ $0.axis }).reduce(0.0, +)
+        let totalBasisSize: CGFloat = basisSizes.reduce(0.0, +)
         let totalSpacing = minimumSpacing * CGFloat(basisSizes.count-1)
 
         /// The size that will be distributed among children (can be positive or negative)
-        let extraSize: CGFloat = layoutSize.axis - (totalBasisSize + totalSpacing)
+        let extraSize: CGFloat = layoutSize - (totalBasisSize + totalSpacing)
 
         assert(extraSize <= 0.0)
```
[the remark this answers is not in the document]
watt: Yeah, it does...
- [ ] fix comment about extraSize

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -271,30 +314,28 @@ extension StackLayout {
...
         assert(basisSizes.count > 0)
 
-        let totalBasisSize: CGFloat = basisSizes.map({ $0.axis }).reduce(0.0, +)
+        let totalBasisSize: CGFloat = basisSizes.reduce(0.0, +)
 
         let minimumTotalSpace = minimumSpacing * CGFloat(basisSizes.count-1)
-        let extraSize: CGFloat = layoutSize.axis - (totalBasisSize + minimumTotalSpace)
+        let extraSize: CGFloat = layoutSize - (totalBasisSize + minimumTotalSpace)
         
         assert(extraSize >= 0.0)
```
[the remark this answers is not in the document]
watt: That seems likely.

BlueprintUI/Sources/Layout/Stack.swift
watt: I could see it working that way, but I think it would break some existing behavior. Consider a Row with `fill` alignment inside a minimum height constraint — if we use the min here, it won't fill that height.

BlueprintUI/Sources/Layout/Stack.swift
watt: I don't think we have enough context to round correctly here. I'm hoping #64 will solve this generally.

PR #69, opened 2020-03-30
[Example DNR] Environment on layout idea

kyleve:
While reviewing https://github.com/square/Blueprint/pull/63, and I had a thought about another possible way to implement environment propagation. Rather than typing up a somewhat confusing comment, I figured it'd be easier to just spend some time writing up the actual code.

Quick overview:

You can get an `Environment` on any `Element` by making it conform to `EnvironmentElement`; which for most cases is adding the conformance and then adding `var environment : Environment = .empty` in their element. The system handles the propagation during layout. If you also need to propagate values down the tree, you can also implement `func editEnvironment(_ environment : inout Environment)` and do what you need to in there.

Added a way to provide standard Environment values when working with Blueprint view, via `internal static func `default`(with view : UIView) -> Environment`. Currently includes `safeAreaInsets`, `screenScale`, `locale`, and `traitCollection`.

What's missing: `Measurable` does not support environment propagation here, which is a large hole. I think if and when we need measurable changes, we'll need a refactor to allow propagating element-based environment edits alongside the measuring tree. Another option is to not allow elements to mutate the Environment from anywhere but the top level, but that seems restrictive – know Andrew hit this as well.


BlueprintUI/Tests/EnvironmentTests.swift
```diff
@@ -0,0 +1,98 @@
...
+//
+//  Created by Kyle Van Essen on 3/28/20.
+//
+
+import Foundation
+import XCTest
+@testable import BlueprintUI
+
+
+class EnvironmentTests : XCTestCase {
```
kyleve: This doesn't do a whole bunch right now, was mostly for me setting breakpoints while writing.

watt:
Can you show an example of how you'd use this to consume environment values? I'm not sure I follow the pattern here.

kyleve:
Definitely, will push an update to the tests to show something more realistic

kyleve:
Pushed a change w/ an example usage, but need to do more changes to make it actually function, since this has been missing more changes to make it work: https://github.com/square/Blueprint/pull/69/files#diff-47af1dcfa56e59d4b4326c7d03acd599

watt:
I think the main concern I have with this approach is that it relies on lifecycle guarantees to work. If you access the environment before layout (e.g. during `init`), it won't be correct. I added some new words about lifecycle guarantees vs explicit argument passing in my PR description.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -0,0 +1,159 @@
...
+    associatedtype Value
+
+    static var defaultValue: Self.Value { get }
+}
+
+
+public struct Environment {
+    public static let empty = Environment()
+    
+    internal static func `default`(with base : Environment? = nil, in view : UIView) -> Environment {
```
kyleve, line 21: Defaults

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -42,6 +42,9 @@ public final class BlueprintView: UIView {
             setNeedsViewHierarchyUpdate()
         }
     }
+    
+    /// The default environment passed down to the element tree.
+    public var environment : Environment = .empty
```
kyleve, line 47: Default env

PR #72, opened 2020-04-02
Stack underflow measurement fix

watt:
This is a follow-up to #68 to fix a case that was handled wrong: measurement with underflow.

## Background

Prior to #68, measurement was always unconstrained. This caused undesired layout when stacks contained flexible children, like text. After #68, measurement uses the same behavior as layout. This improves measurement when there is overflow, but in the case of underflow, it causes measurement to incorrectly use the entire available constraint size.

## Fix

This PR corrects the behavior by splitting out stack layout's internal size constraints from 2 cases to 3:

- `unconstrained`: Sizes are based on the natural (unconstrained) measurements of the children. Used during unconstrained measurement.
- `atMost`: When there is overflow, compress children. When there is underflow, use their natural sizes. Used during constrained measurement.
- `exactly`: When there is overflow, compress children. When there is underflow, expand them. Used during layout.

These 3 cases should allow measurement and layout to continue to share most of their logic.


BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -596,15 +624,30 @@ extension StackLayout {
 
     /// Represents a size constraint with symbolic axes
     struct VectorConstraint {
-        var axis: SizeConstraint.Axis
-        var cross: SizeConstraint.Axis
+        enum Axis {
+            case exactly(CGFloat)
+            case atMost(CGFloat)
+            case unconstrained
```
watt, line 630: New cases are here, replacing use of `SizeConstraint.Axis`.

PR #74, opened 2020-04-11
Rename Builder to ContentStorage

kyleve:
Builder was a weird name for this value, since it was always accessed through .storage. Rename to ContentStorage.

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -3,16 +3,16 @@ import UIKit
 /// Represents the content of an element.
 public struct ElementContent: Measurable {
 
-    private let storage: ContentStorage
+    private let storage: AnyContentStorage
 
     /// Initializes a new `ElementContent` with the given layout and children.
     ///
     /// - parameter layout: The layout to use.
     /// - parameter configure: A closure that configures the layout and adds children to the container.
-    public init<LayoutType: Layout>(layout: LayoutType, configure: (inout Builder<LayoutType>) -> Void = { _ in }) {
-        var builder = Builder(layout: layout)
-        configure(&builder)
-        self.storage = builder
+    public init<LayoutType: Layout>(layout: LayoutType, configure: (inout ContentStorage<LayoutType>) -> Void = { _ in }) {
+        var storage = ContentStorage(layout: layout)
+        configure(&storage)
+        self.storage = storage
     }
 
     public func measure(in constraint: SizeConstraint) -> CGSize {
@@ -26,7 +26,6 @@ public struct ElementContent: Measurable {
     func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
         return storage.performLayout(attributes: attributes)
     }
-
 }
 
 extension ElementContent {
@@ -76,17 +75,9 @@ extension ElementContent {
 }
 
 
-fileprivate protocol ContentStorage: Measurable {
-    var childCount: Int { get }
-    func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
-}
-
-
-
-
 extension ElementContent {
 
-    public struct Builder<LayoutType: Layout> {
+    public struct ContentStorage<LayoutType: Layout> {
 
         /// The layout object that is ultimately responsible for measuring
         /// and layout tasks.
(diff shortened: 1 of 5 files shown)


kyleve:
Bump

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -13,9 +13,13 @@ public protocol StackElement: Element {
 extension StackElement {
 
     public var content: ElementContent {
-        return ElementContent(layout: layout) {
+        ElementContent(layout: layout) {
```
bencochran, line 16: Our Package.swift specifies `.v5` as the Swift language version, which doesn’t include implicit `return`s. But there doesn’t seem to be a `.v5_1`. We should probably either bump to `.version("5.1")` or not use 5.1-only features?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -98,27 +89,41 @@ extension ElementContent {
         init(layout: LayoutType) {
             self.layout = layout
         }
+        
+        /// Adds the given child element.
+        public mutating func add(element: Element, traits: LayoutType.Traits = LayoutType.defaultTraits, key: AnyHashable? = nil) {
```
bencochran, line 94: I think we put `traits` before `element` to match the style for Row and Column, etc where you want to be able to have the nested element last (so that it can span multiple lines without the traits/properties being down at the end). That use-case is less relevant for `ContentStorage`, but it would be nice to keep the consistency I think.

bencochran, line 94: While I’m thinking about it: would be maybe nice to add a convenience `func add(elements: [Element])` that uses default traits and no keys, since that’s a common use-case for these from my quick code scan.

bencochran:
Only blocking thing I see is the Swift 5.1 issue.

BlueprintUI/Sources/Layout/Stack.swift
kyleve: Is "5.1" a real thing? AFAIK, Xcode only thinks in terms of major version, right?

bencochran: Not exactly sure TBH. But [there’s a `.v4_2`](https://github.com/apple/swift-package-manager/blob/master/Sources/PackageDescription/LanguageStandardSettings.swift#L94)

watt: We could set the `swift-tools-version` magic comment to 5.1, but from what I understand the 5.1 enum case isn't there because it's not really a thing you can enforce, and there are only going to be cases for major versions going forward. This also came up when I specified 5.1 in the podspecs, because apparently `swiftc` also won't accept anything above major version 5.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -98,27 +89,41 @@ extension ElementContent {
...
-    public mutating func add(traits: LayoutType.Traits = LayoutType.defaultTraits, key: AnyHashable? = nil, element: Element) {
-        let child = Child(
-            traits: traits,
-            key: key,
-            content: element.content,
-            element: element)
-        children.append(child)
-    }
 
+fileprivate protocol AnyContentStorage : Measurable {
```
watt, line 120: Usually when I see a type prefixed with `Any`, it's a type-erasing struct wrapper, like [AnyIterator, AnySequence, AnyHashable, etc](https://developer.apple.com/documentation/swift/swift_standard_library/collections/supporting_types). I would prefer to keep the protocol as-is and give the implementation a more expressive name.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -98,27 +89,41 @@ extension ElementContent {
...
+    var childCount: Int { get }
+    func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
 }
 
-extension ElementContent.Builder: ContentStorage {
+
+extension ElementContent.ContentStorage : AnyContentStorage {
```
watt, line 126: Hmm. To me, this implementation feels like its distinguishing feature is that you can use it to build the list of its children. How about `BuildingContentStorage`? Or maybe `DefaultContentStorage`, since it's currently the only type?

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -112,9 +112,9 @@ private struct TestContainer: Element {
     var children: [Element]
     
     var content: ElementContent {
-        return ElementContent(layout: TestLayout()) { (builder) in
```
watt, line 115: Not a fan of this style. Loss of useful information here.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: I took to this pattern after using it for type erasure in Listable, since i liked that it meant you didnt need to define yet another wrapper type to do erasure. I don't think the actual type of the Any* thing doesnt really matter, IMO

BlueprintUI/Tests/BlueprintViewTests.swift
kyleve: I've taken to using the implicit $0 for methods under < 4 lines long, since its pretty clear what you're referring to

BlueprintUI/Sources/Element/ElementContent.swift
kylebshr: I kind of agree that it makes more sense not to treat this as erasure when it's really just the base functionality for building storage. Maybe one of the names @watt suggested below makes sense? Not a super strong opinion, after all this is kind of like `AnyObject` right? (Types conform to it, aren't wrapped it).

kyleve: It's too bad objects and protocols cant have the same name like they do in ObjC

PR #75, opened 2020-04-13
[WIP DNR] Explore Merging Measurement & Layout

kyleve:
## Overview

This PR is an exploration of what it would look like to merge the measurement and layout passes, since they regularly depend on each other, which results in some oddly shaped code to reference internal layout from both places, and more importantly, potential performance implications for large or deep element trees – we now only have to do one pass to perform a layout, versus a measurement pass followed by a layout pass.

This makes implementing things like `GeometryReader`, etc, easier, because all layout information flows through a single tree.

## Major Changes

- `Measurable` protocol has been removed; `func measure(in constraint: SizeConstraint) -> CGSize` is folded into...
- The `Layout` protocol. All layout is now done via one method: `func layout(in constraint : SizeConstraint, items: [LayoutItem<Self>]) -> LayoutResult`.

### Layout

Digging into `Layout` a bit more:

```
public protocol Layout {
   associatedtype Traits = ()
   static var defaultTraits: Self.Traits { get }

   func layout(in constraint : SizeConstraint, items: [LayoutItem<Self>]) -> LayoutResult
}
```
The `layout` method is passed a set of items to lay out, which look like this:
```
public struct LayoutItem<LayoutType:Layout> {
    public var element: Element
    public var content: ElementContent
    public var traits: LayoutType.Traits
    public var key: AnyHashable?
}
```

And it returns a layout result, which is just its size, plus the layout attributes of its children:

```
public struct LayoutResult {

    public var size : CGSize
    public var layoutAttributes : [LayoutAttributes]
```
### SingleChildLayout

I've also updated `SingleChildLayout` to follow the new single pass layout pattern.

It now has one method:

```
func layout(in constraint : SizeConstraint, child : MeasurableChild) -> SingleChildLayoutResult
```

Which behaves similarly to a normal `Layout`, but simplified for a single child:

```
public struct SingleChildLayoutResult {
    public var size : CGSize
    public var layoutAttributes : LayoutAttributes
}
```

A remaining TODO is to migrate away from `MeasurableChild` here, and instead provide layout directly.

### Migration of existing code
I've kept the shape of existing measure + layout code in this change, and instead folded both bits into the single layout method. This does lose us some of the benefit of the idealistic implementation of only one layout pass, but thats fine: elements can be migrated in pieces to the new layout system in individual PRs. A good way to tell when code is still in this shape, is if it uses
```
public init(
    size sizeProvider : () -> CGSize,
    layoutAttributes layoutAttributesProvider : (CGSize) -> [LayoutAttributes]
) {
    let size = sizeProvider()
    let layoutAttributes = layoutAttributesProvider(size)

    self.size = size
    self.layoutAttributes = layoutAttributes
}
```
Which is essentially a measurement which hands off to a layout method.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -98,90 +90,86 @@ extension ElementContent {
...
-    func performLayout(attributes: LayoutAttributes) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
-
-        let childAttributes = layout.layout(size: attributes.bounds.size, items: layoutItems)
-
-        var result: [(identifier: ElementIdentifier, node: LayoutResultNode)] = []
-        result.reserveCapacity(children.count)
         
-        var identifierFactory = ElementIdentifier.Factory(elementCount: children.count)
+        /// Adds the given child element.
+        public mutating func add(element: Element, traits: LayoutType.Traits = LayoutType.defaultTraits, key: AnyHashable? = nil) {
```
kyleve, line 95: Deprecate in favor of two below

kyleve:
Ok, got it working with at least the demo app without needing to always defer to the constraint.maximum during layout

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -54,6 +54,24 @@ extension SizeConstraint {
 
 }
 
+extension SizeConstraint {
+    public struct Hugging : Equatable {
```
kyleve, line 58: Remove

PR #76, opened 2020-04-15
SwiftUI Preview Support

kyleve:
This PR adds support for previewing Blueprint elements via Xcode's SwiftUI preview functionality, plus some helper APIs to make creating a list of previews, and device previews easier.

```
import BlueprintUI

struct MyElement : Element {
   ...
}

#if DEBUG && canImport(SwiftUI) && !arch(i386)

import SwiftUI

@available(iOS 13.0, *)
struct TestingView_Preview: PreviewProvider {
    static var previews: some View {
        ElementPreview.commonDevices {
            MyElement()
        }
    }
}

#endif
```

Screen Shot 2020-04-14 at 5 43 12 PM


kyleve:
Now builds on CI, need to verify this also works for Archive builds.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -0,0 +1,240 @@
...
+    
+    // MARK: View
+    
+    public var body: some View {
+        ForEach(self.previewTypes, id: \.self) { previewType in
+            previewType.preview(for: ElementView(element: self.provider()))
+        }
+    }
+    
+    private struct ElementView : UIViewRepresentable {
```
kyleve, line 138: Not super related to this PR, but good example of how we can mix and match Blueprint into SwiftUI.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -0,0 +1,240 @@
...
+    }
+    
+    /// Creates a new `ElementPreview` with the provided preview types.
+    ///
+    /// You can pass as many preview types as you would like to see your
+    /// element rendered in those different environments.
+    ///
+    /// If you do not pass a preview type, `.thatFits` is used.
+    public init(
+        with previewTypes : [PreviewType] = [.thatFits],
```
kylebshr, line 123: I don’t think it really matters, but any reason for the default arg when that just duplicates the same init as the single preview type? Wondering if that confuses auto complete at all. Probably fine though!

kyleve:
> Now builds on CI, need to verify this also works for Archive builds.

Archive builds good to go.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
kyleve: Fair! Will remove; I’d added this method first without rethinks once I added the singular one above.

Also realizing I should assert in here that at least one value is passed, or maybe log since crashes in previews are hard to debug

kingreza:
is `!arch(i386)` needed?

kyleve:
> is `!arch(i386)` needed? how does that actually work? you don't want to run the preview on arch i386? shouldn't that prevent it from running on my MacBook Pro?

Yes, it's needed; see the comment here: https://github.com/square/Blueprint/pull/76/files#diff-472cce270b83724547e6fc8f775947cfR14 and more info here https://twitter.com/harlanhaskins/status/1250252601670504454

Your laptop runs `x84_64`, not `i386`.

kingreza:
> Your laptop runs `x86_64`, not `i386`.
This makes sense to me now since i386 is 32-bit. I'm not sure why running `arch` in command line returns `i386`. Anyways, not related to this. Thanks!
image

kyleve:
I believe that may be due to differences between how mac and ios sim builds are run? Not 100% sure.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -0,0 +1,238 @@
...
+@available(iOS 13.0, *)
+extension ElementPreview {
+   
+    /// The preview type to use to display an element in an Xcode preview.
+    ///
+    /// We provide three preview types: A specific device type, a fixed size, and the size that fits the view.
+    public enum PreviewType : Hashable {
+        
+        /// The preview will be inside the provided device (eg, iPhone X).
+        case device(Device)
```
bencochran, line 160: It’d be nice if this held a `PreviewDevice` and we made the current `Device` cases be `static var` extensions on `PreviewDevice` instead. I think it would be the same API from the call site, but would allow consumers to use new `PreviewDevice`s without having to bump Blueprint.

bencochran:
Just the one suggestion, and I think it could be a followup

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
kyleve: Oh interesting, I like that, will do

kyleve: Done!

PR #81, opened 2020-04-17
Expose accessibilityIdentifier to `AccessibilityElement`

chrissonoda:
Expose this so it can be used in UI testing


bencochran:
Thanks for this!

PR #83, opened 2020-04-17
Update RELEASING.md with the new restriction of not pushing to master

kyleve:


RELEASING.md
```diff
@@ -9,30 +9,36 @@ bundle exec pod trunk info BlueprintUI
...
 ```
 
-See [the CocoaPods documentation for pod trunk](https://guides.cocoapods.org/making/getting-setup-with-trunk) for more information about setting up credentials on your device.
+See [the CocoaPods documentation for pod trunk](https://guides.cocoapods.org/making/getting-setup-with-trunk) for more information about setting up credentials on your device. If you need to be added as an owner, ping in #blueprint on Slack (Square only).
 
-1. Make sure you're on the `master` branch
+1. Make sure you're on the `master` branch, and `git pull` to get the latest commits.
 
-1. Make sure that the library version in both `BlueprintUI.podspec` and `BlueprintUICommonControls.podspec` is correct (it should match the version number that you are about to release).
+1. Create a branch off `master` to update the version numbers and `Podfile.lock`. An example name would be `kve/release-0.9`.
```
watt, line 16: Maybe `your-username/release-0.1.0`?

RELEASING.md
```diff
@@ -9,30 +9,36 @@ bundle exec pod trunk info BlueprintUI
...
+
+1. Update the library version in both `BlueprintUI.podspec` and `BlueprintUICommonControls.podspec` if it has not already been updated (it should match the version number that you are about to release).
 
 1. Run `bundle exec pod install` to update the `Podfile.lock` with the new versions.
```
watt, line 20: Can you add "from the SampleApp directory"? Been meaning to change this since it moved from root.

RELEASING.md
```diff
@@ -9,30 +9,36 @@ bundle exec pod trunk info BlueprintUI
...
 
-1. Push your changes
+1. Push your branch and open a PR into `master`.
+
+1. Once the PR is merged, switch back to `master` and tag the release:
    ```bash
-   git push origin master && git push origin 0.1.0
+   git checkout master
+   git pull
+   git tag 0.1.0
```
watt, line 33: This should probably reference the exact commit of the merge now, in case HEAD is not the PR that was just merged, e.g. `git tag 0.1.0 <sha-from-merge>`

RELEASING.md
kyleve: Ooh yeah, will do!

RELEASING.md
kyleve: +1

RELEASING.md
kyleve: Ah true

RELEASING.md
kyleve: Actually, thought: Should we just un-nest the podfile? Always found it weird that it was in `SampleApp`, given it's a critical part of how we locally consume and  publish the `podspecs`, which are at the root.

bencochran: I like the setup that @amorde set up for Workflow with a Development podspec that contains all the samples: https://github.com/square/workflow/pull/964 (I don’t _think_ we’d even need the Podfile in that case)

watt: Either way is fine with me. The Podfile isn't related to publishing, it's only used by the SampleApp.

kyleve: I don't think the podspec-only tricks without a Podfile because you have no way to point to other local pods, right? Eg, in that example, doesn't this:

```
  s.dependency 'Workflow'
  s.dependency 'WorkflowUI'
```

Use the published versions of Workflow because it doesnt know to look locally? (Or maybe that's been resolved?)

amorde: You can point to local podspecs if necessary either using a Podfile or (what we did) using the `--local-sources` CLI option https://github.com/square/workflow/pull/964/files#diff-bc5b13d6a0956bdb7636d99b59384e18R1-R4

kyleve: Huh, interesting, TIL!

kyleve:
Should be good to go now.

PR #87, opened 2020-04-26
Fix ConstrainedSize measurement, ElementPreview sizing

kyleve:
Change how `ConstrainedSize` does measurement, so that the size constraint passed to the inner element is constrained to the maximum size of the `ConstrainedSize`.

**Details**
Previously, `ConstrainedSize` would measure its child element based on the constraint passed into its `measure` method, regardless of if that `ConstrainedSize` itself had maximum constraints on either the width or height of its contained element. This worked fine for elements which didn't have a dependency between their width and height, but it would break things like `Label`, where making the `Label` narrower should re-flow the text, and thus make the `Label` taller. The fix is to constrain the upper bound of the constraint passed into the `measure` method based on the maximum size of the `ConstrainedSize`, and then continue on with measurement.

**ElementPreview**
I found this via trying to make `ElementPreview` for Xcode previews respect the max size of the simulator, which is why there are changes in that type as well. We use `GeometryReader` to bridge the gap between the SwiftUI and Blueprint layout systems.

This is a similar issue to the one @watt fixed in stacks here: https://github.com/square/Blueprint/issues/58


BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -141,30 +141,32 @@ public struct ElementPreview : View {
...
-            return BlueprintView(element: self.element)
+            let view = BlueprintView()
+            view.element = self.element
+            
+            return view
         }
         
         func updateUIView(_ view: BlueprintView, context: Context) {
             view.element = self.element
+            view.layoutIfNeeded()
```
kyleve, line 167: Remove

BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -1,21 +1,44 @@
...
 public struct ConstrainedSize: Element {
 
-    public var wrappedElement: Element
+    /// The element whose measurement will be constrained by the `ConstrainedSize`.
+    public var wrapped: Element
 
+    /// The constraint to place on the width of the element.
     public var width: Constraint
+    
+    /// The constraint to place on the width of the element.
```
kyleve, line 23: height

BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -1,21 +1,44 @@
 import UIKit
 
-/// Constrains the measured size of the content element.
+/// Constrains the measured size of the contained element in the ranges specified by the `width` and `height` properties.
+///
+/// There are several constraint types available for each axis. See `ConstrainedSize.Constraint` for a full list and in-depth
+/// descriptions of each.
+///
+/// Notes
```
kyleve, line 8: I documented this existing behavior, but I don't love it... thoughts on changing the behavior so the max returned is the max of the measurement constrained OR the constrained size, whichever is smaller?

kyleve: (You can see how confusing this existing behaviour gets in the tests, in the `test_in_atMost` method.)

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -65,14 +65,23 @@ extension SizeConstraint {
 
         /// The measurement is unconstrained in the given dimension.
         case unconstrained
+        
+        /// Creates a `SizeConstraint` with the provided value.
+        public init(_ value : CGFloat) {
```
bencochran, line 70: Can we keep this `internal`?

kyleve: G1F1ECN

BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -67,18 +102,45 @@ extension ConstrainedSize {
...
 
         func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
             return LayoutAttributes(size: size)
         }
-
+        
+        private func constraining(_ axis : SizeConstraint.Axis, to constraint : Constraint) -> SizeConstraint.Axis {
+            let constrained = constraint.applied(to: axis.maximum)
+                
+            if constrained == .greatestFiniteMagnitude {
```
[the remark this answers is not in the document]
kyleve: Done, moved private to `Constraint`

PR #88, opened 2020-04-26
SwiftUI Style .foo() wrappers

kyleve:
This PR implements SwiftUI-style wrappers which I think improve the succinctness and readability of Blueprint code.

The main file to check out for updates/examples to written consumer code is `ViewController.swift`.

### SwiftUI-Style `.foo()` methods for wrapping types
Before this change, if you wanted to have a scroll view, which wrapped a box, which wrapped an inset, which wrapped a constrained size, which wrapped a label (and so on), the code would look something like this:

```
ScrollView(.fittingHeight) (
   wrapping: Box(
        backgroundColor .lightGrey,
       wrapping: Inset(
          uniformInset: 10.0,
          wrapping: ConstrainedSize(
             height: .atLeast(20.0),
             wrapping: Label(
                text: "Hello, world!"
             )
         )
      )
   )
)
```

Which obviously is pretty verbose... It's nice that from the shape of the code you see what's going on, but it's pretty hard to follow in practice, especially if these inner statements get broken up into intermediary variables.

After this change, to write this code, this becomes like the following, similar to SwiftUI:

```
Label(text: "Hello, World!")
   .constrainedTo(height: .atLeast(20.0))
   .inset(by: 20.0)
   .box(background: .lightGrey)
   .scrollable(.fittingHeight)

```


BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -77,3 +77,17 @@ public struct ConstrainedAspectRatio: Element {
...
+
+
+public extension ConstrainedAspectRatio {
+    /// Initializes with the given properties.
+    ///
+    /// - parameters:
+    ///   - aspectRatio: The aspect ratio that the content size should match.
+    ///   - contentMode: Whether the aspect ratio should be reached by expanding the content
+    ///     element's size to fill its parent or shrinking it to fit.
+    ///   - wrapping: The content element.
```
kyleve, line 89: Remove

BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -51,6 +51,17 @@ extension ConstrainedSize {
 
 }
 
+public extension Element {
+    
+    func constrainedTo(
```
kyleve, line 56: Missing comment

BlueprintUI/Sources/Layout/Inset.swift
```diff
@@ -63,13 +65,34 @@ public struct Inset: Element {
...
 
     public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
         return nil
     }
+}
+
+
+public extension Element {
+    
+    func inset(top: CGFloat = 0.0, bottom: CGFloat = 0.0, left: CGFloat = 0.0, right: CGFloat = 0.0) -> Inset {
```
kyleve, line 81: Missing comments here

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -7,9 +7,74 @@ import UIKit
 public protocol StackElement: Element {
     init()
     var layout: StackLayout { get }
-    var children: [(element: Element, traits: StackLayout.Traits, key: AnyHashable?)] { get set }
+    var children: [StackChild] { get set }
 }
 
+public struct StackChild {
```
kyleve, line 13: Missing comments

BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
```diff
@@ -19,5 +19,10 @@ public struct AccessibilityBlocker: Element {
             config[\.accessibilityElementsHidden] = true
         }
     }
+}
 
+public extension Element {
+    func disableAccessibility() -> AccessibilityBlocker {
```
kyleve, line 25: `blockAccessibility()`

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -105,5 +105,25 @@ public struct AccessibilityElement: Element {
             config[\.isAccessibilityElement] = true
         }
     }
+}
+
 
+public extension AccessibilityElement {
```
kyleve, line 111: Missing comment

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -103,7 +114,26 @@ extension Box {
         case none
         case simple(radius: CGFloat, opacity: CGFloat, offset: CGSize, color: UIColor)
     }
+}
 
+public extension Element {
```
kyleve, line 119: Missing comment

kylebshr:
**1) SwiftUI methods**

I really like this, and actually added similar functions in a side project over the weekend for some custom "view modifiers". I think it's both easier to read and write - I often start out with a base element like a label, and in the current form it's clunky to repeatedly wrap it in elements, and often "obscures" the real UI. 

**2) Stack improvements**

*StackChild*: I think this makes sense, having named static methods instead of having to read the docs for grow/shrink priority every time seems great (jokes aside, the names are way more readable and beginner friendly).

*Operators*: I'm not a huge fan of custom operators for UI, since they're not very discoverable, lead to divergence in code style (since it doesn't make sense to remove the functions) and it just doesn't "feel right" to me. Not a super strong opinion here though :)

kyleve:
@kylebshr re operators: generally agree, but I think there's a time and place for them, eg common operations like this. As long as you can do everything _without_ operators that you can do with them, I feel it's fine.

kyleve:
@kyleve: TODO: Split Stack changes out into a new PR

kyleve:
I've removed the stack changes from this PR

SampleApp/Sources/XcodePreviewDemo.swift
```diff
@@ -16,7 +16,7 @@ struct TestElement : ProxyElement {
             $0.verticalUnderflow = .justifyToStart
             
             for index in 1...12 {
-                $0.add(child: Label(text: "Hello, World") {
+                $0.add(growPriority: 0.0, shrinkPriority: 00, child: Label(text: "Hello, World") {
```
kyleve: Revert

BlueprintUI/Sources/Layout/Aligned.swift
```diff
@@ -111,3 +111,23 @@ public struct Aligned: Element {
...
+
+public extension Element {
+    /// Wraps the element in an `Aligned` element with the provided parameters.
+    ///
+    /// - parameters:
+    ///   - vertically: The vertical alignment. Defaults to `.centered`.
+    ///   - horizontally: The horizontal alignment. Defaults to `.centered`.
+    ///
+    func aligned(
+        vertically: Aligned.VerticalAlignment = .center,
```
kylebshr, line 123: I don't love a default in this case because it's not clear reading the code that `.aligned()` means center-aligned. What do you think about removing the default and adding a `centered()` function if that's a common use case?

Edit: looks like there's already `centered`! Let me know what you think about removing the default

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -77,3 +77,21 @@ public struct ConstrainedAspectRatio: Element {
...
+
+public extension ConstrainedAspectRatio {
+    /// Initializes with the given properties.
+    ///
+    /// - parameters:
+    ///   - aspectRatio: The aspect ratio that the content size should match.
+    ///   - contentMode: Whether the aspect ratio should be reached by expanding the content
+    ///     element's size to fill its parent or shrinking it to fit.
+    ///
+    func aspectRatio(
```
kylebshr, line 90: Would it be a good idea to always match the Layout name? I might type `.con..` expecting to find `ConstrainedAspectRatio` if I was already familiar with the type. No strong opinion though, there's an argument to be made for keeping them succinct

BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
```diff
@@ -19,5 +20,10 @@ public struct AccessibilityBlocker: Element {
             config[\.accessibilityElementsHidden] = true
         }
     }
+}
 
+public extension Element {
+    func blockAccessibility() -> AccessibilityBlocker {
```
kylebshr, line 26: Doc comment

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -105,5 +105,25 @@ public struct AccessibilityElement: Element {
...
+    func accessibility(
```
kylebshr, line 112: Doc comment

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -61,6 +68,20 @@ public struct ScrollView: Element {
     }
 }
 
+public extension Element {
+    func scrollable(
```
kylebshr, line 72: Doc comment

BlueprintUI/Sources/Layout/Aligned.swift
kyleve: I kept the defaults from the normal init on `Aligned`, since sometimes you only want to pass one value, and I think centered is a pretty reasonable default. Thoughts?

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
kyleve: Yeah, that's fair, will change – I shortened it since an "Aspect Ratio" is already inherently constrained, but adding `constrained` to the front is nice for symmetry with the type name and `ConstrainedSize` as well.

kyleve: Actually, `constrainedTo(aspectRatio:` works nicely as well, going with that.

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -77,3 +77,21 @@ public struct ConstrainedAspectRatio: Element {
         }
     }
 }
+
+
+public extension ConstrainedAspectRatio {
```
kyleve, line 82: This should be on `Element`, doh.

BlueprintUI/Sources/Layout/Aligned.swift
kyleve: Eh, what the hell, removing the default

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -105,5 +105,28 @@ public struct AccessibilityElement: Element {
             config[\.isAccessibilityElement] = true
         }
     }
+}
+
 
+public extension AccessibilityElement {
```
kyleve, line 111: Element

kylebshr:
Haha, didn’t realize the init had a default, I don’t like it there for the same reason but I’ll let you decide!

mattfaluotico:
Popping in for a comment on Stack Improvements, even though this is merged. IMO you get a bit more clarity about what is being added when the Element proceeds the stacking traits. 

Was [dabbling](https://github.com/mattfaluotico/Blueprint/commit/11ed5f11c1bbf5c65361247b1e7432ef1a2c98ca) and added a similar StackChild, with the flex traits using function builders (like this PR adds)

so you could get:  

```
Row {
    Label(text: "A truly great label")
        .growPriority(1)
}
```

I like @kyleve's use of fixed a lot! 

```
Row {
    Label(text: "A truly great label")
        .fixed
    Spacer()
    Label(text: "Another great label")    
}
```

PR #89, opened 2020-04-27
Guard against 32 bit arm as well

kyleve:


BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Kyle Van Essen on 4/14/20.
 //
 
-#if DEBUG && canImport(SwiftUI) && !arch(i386)
+#if DEBUG && canImport(SwiftUI) && !arch(i386) && !arch(arm)
```
watt, line 8: I probably missed the original discussion of this. Does a whitelist not work? e.g. `arch(x86_64)`?

Also should try to get this into a separate module?

kyleve: I figured it was either a whitelist or blacklist, but in theory every new arch will work, while there are only two we know which will not work. Also, this:
```
#if DEBUG && canImport(SwiftUI) && !arch(i386) && !arch(arm)
```
reads better than this to me:
```
#if DEBUG && canImport(SwiftUI) && (arch(arm64) || arch(x86_64))
```

PR #90, opened 2020-04-28
0.9.2

kyleve:


kyleve:
> changelog?

Doh! And it's my own damn changelog, too...

kyleve:
(I wonder if we can enforce the changelog thing with a build time rule somehow.. hmm)

PR #92, opened 2020-05-06
Update sizeThatFits to check each axis' constraint separately

kyleve:


BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -2,6 +2,70 @@ import XCTest
...
+            blueprintView.element = element
+            
+            XCTAssertEqual(blueprintView.sizeThatFits(CGSize(width: 200, height: 100)), CGSize(width: 100, height: 50))
+        }
+        
+        // Unconstrained in both axes.
+        
+        do {
+            let element = MeasurableElement { constraint in
+                XCTAssertEqual(constraint.maximum, CGSize(width: CGFloat.greatestFiniteMagnitude, height: CGFloat.greatestFiniteMagnitude))
```
watt, line 27: We should assert that the constraint is explicitly `unconstrained`. The existing behavior would have been the same for `maximum`.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -75,16 +75,41 @@ public final class BlueprintView: UIView {
...
-            constraint = SizeConstraint(size)
+        guard let element = element else {
+            return .zero
         }
-        return element.content.measure(in: constraint)
+        
+        return element.content.measure(in: Self.sizeThatFitsMeasurementConstraint(with: size))
+    }
+    
+    private static func sizeThatFitsMeasurementConstraint(with size : CGSize) -> SizeConstraint {
```
watt, line 104: nit: if this behavior is specific to `sizeThatFits`, can you inline it, or nest this func in `sizeThatFits`?

BlueprintUI/Tests/BlueprintViewTests.swift
kyleve: Ooh yeah, good point

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
kyleve: TIL you can nest functions in other functions, whoa

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -2,6 +2,70 @@ import XCTest
 @testable import BlueprintUI
 
 class BlueprintViewTests: XCTestCase {
+    
+    func test_sizeThatFits() {
```
kyleve, line 6: Check both `0.0`, and `.gfm`.

BlueprintUI/Tests/BlueprintViewTests.swift
kyleve: Done!

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
kyleve: Done

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -75,16 +75,51 @@ public final class BlueprintView: UIView {
...
+        guard let element = element else {
+            return .zero
         }
-        return element.content.measure(in: constraint)
+        
+        func measurementConstraint(with size : CGSize) -> SizeConstraint {
+            
+            let unconstrainedValues : [CGFloat]
+            
+            /// Note: `.infinity` is included because `.greatestFiniteMagnitude` falls back to `.infinity` on 32 bit devices.
```
kyleve, line 105: Can someone check me on this? Without this, the tests fail on iOS 10 because we're comparing to infinity. I assume it's a 32 bit float precision thing...

kylebshr: Weird, I'm going to run the tests myself but in a different project, this is `true` on 10.3.1 on iPhone 5 

```
let unconstrainedValues = [0.0, .greatestFiniteMagnitude]
print(unconstrainedValues.contains(.greatestFiniteMagnitude))
```

kyleve: Try removing the `.infinity` from here and running the tests on an iPhone 5 – they'll fail, which was why I added this. Agree, it's strange, doesn't seem quite right...

kylebshr: Aha, it's because some of the tests have `CGFloat.greatestFiniteMagnitude` and some have `.greatestFiniteMagnitude` which I assume is inferring `Double` for some reason. Perhaps both 32 and 64 bit should check for `infinity`? Or should it be `CGFloat(Double.greatestFiniteMagnitude)`? This is getting weird :P

kyleve: OHHHH!! Damn, that makes a lot of sense. Guess I'll update them to reference CGFloat explicitly, since that's what we actually want.

kylebshr: I think inferring `Double` happens if you use `x.0` for the other arg, since that's a double and not a CGFloat without casting explicitly. So it's probably pretty common for a size to be init'd with Double.greatestFiniteMagnitude.

kyleve: Fixed and removed the iOS version bifurcation here.

kyleve: Re initialization of CGSize: I need to test this, but CGSize does store things as CGFloat, so, maybe not a common issue?

```
    public var x: CGFloat

    public var y: CGFloat
```

(From generated Swift interface for CGSize)

kyleve: Updated to include `infinity` as well

PR #96, opened 2020-05-10
Fix iPhoneSE preview name now that there are two iPhone SEs.

kyleve:
Love me some stringly typed code

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/ElementPreview.swift
+++ b/BlueprintUI/Sources/Blueprint View/ElementPreview.swift
@@ -91,7 +91,7 @@ public struct ElementPreview : View {
         return Self(
             named: name,
             with: [
-                .device(.iPhoneSE),
+                .device(.iPhoneSE_1),
                 .device(.iPhone8),
                 .device(.iPhone8Plus),
                 .device(.iPhoneXs),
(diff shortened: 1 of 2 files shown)


kylebshr:
Just curious, how do these strings work? Did they change with an Xcode update?

kyleve:
Yeah, seems to have changed with Xcode 11.4.1 for the new SE. I guess I may need to do a tools check with an `#if` to preserve the old behaviour for older IDEs, actually..

kyleve:
Turns out there's no `__IPHONE_OS_VERSION_MAX_ALLOWED` in Swift, so I instead used a runtime check. Should be good to go now!

kyleve:
Also, re:

> Also, it might be less confusing to use iPhone 5s for that screen size

The 5S didn't get iOS 13, so it's not a possible device.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -251,43 +251,89 @@ extension ElementPreview {
...
+    /// iPhone 8
+    
     static var iPhone8 = PreviewDevice("iPhone 8")
     static var iPhone8Plus = PreviewDevice("iPhone 8 Plus")
     
-    static var iPhoneSE = PreviewDevice("iPhone SE")
+    /// iPhone SE
+    
+    static var iPhoneSE_1 : PreviewDevice {
+        if isAtLeast(13, 4, 1) {
```
watt, line 268: Does `#available(iOS 13.4.1, *)` not do this?

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
```diff
@@ -251,43 +251,89 @@ extension ElementPreview {
...
+            return PreviewDevice("iPhone SE (1st generation)")
+        } else {
+            return PreviewDevice("iPhone SE")
+        }
+    }
+    
+    static var iPhoneSE_2 : PreviewDevice {
```
watt, line 275: Should we mark this one `@available(iOS 13.4.1)` instead of making it possible for it to be 2 different devices?

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
kyleve: Oh yeah.. duh. Not sure why I didnt do that.

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
kyleve: :thumbsup:

BlueprintUI/Sources/Blueprint View/ElementPreview.swift
kyleve: Done

PR #100, opened 2020-05-20
Move default values into initialzer for Image Element

mattfaluotico:
Allow the `tintColor` and `contentMode` to get passed in via the initializer for Image, similar to how `Box` works.

I found most of the time i was using image, I needed to create a local var and change one of these properties. This would retain the ability to edit them, but also pass in via initialization to get everything done in one line.


kyleve:
Mind updating the CHANGELOG.md as well?

kyleve:
Also, congrats on PR #100! 🎊

mattfaluotico:
Added to changelog

PR #102, opened 2020-06-01
Environment keys

watt:
Some basic environment keys. Most line up with SwiftUI keys.

- calendar
- display scale
- layout direction
- locale
- safe area insets
- time zone

Safe area insets is the only one that doesn't exist in SwiftUI.

These keys' values are set by the hosting BlueprintView:

- display scale (based on `window.screen.scale`)
- layout direction (based on `semanticContentAttribute`)
- safe area insets

Notably absent is `traitCollection`. SwiftUI provides its own abstractions for trait collection properties. Not sure if we want to do the same or not.

I also updated the sample code and added an `AdaptedEnvironment` init that takes a keypath, so that `EnvironmentKey` types can be private.


BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -133,6 +133,18 @@ public final class BlueprintView: UIView {
         }
         return element.content.measure(in: constraint, environment: .empty)
     }
+
+    public override var semanticContentAttribute: UISemanticContentAttribute {
+        didSet {
+            setNeedsViewHierarchyUpdate()
```
kyleve, line 139: Nit: guard and return if the old and new value are the same

BlueprintUI/Sources/Environment/Keys/CalendarKey.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+extension Environment {
+    private enum CalendarKey: EnvironmentKey {
+        static var defaultValue: Calendar {
+            Calendar.current
```
kylebshr, line 6: There's an `autoupdatingCurrent` getter that updates if the calendar changes. Not sure we should use it, since you can end up using two different calendars if it changes and your UI doesn't update itself - but might be good to document that this defaults to just `current`

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -199,7 +209,23 @@ public final class BlueprintView: UIView {
         /// views that are actually generated by the root element.
         return rootController.children
     }
-    
+
+    private var environment: Environment {
```
kylebshr, line 213: Pretty nit-picky, but since this var is used to create a new environment when the view update happens, I think semantically it makes sense to make this a function (`makeCurrentEnvironment` or something). Wouldn't want someone to "optimize" this by making it a non-computed property.

BlueprintUI/Sources/Environment/Keys/CalendarKey.swift
watt: Good catch! I actually checked on this, and chose it because SwiftUI's calendar property defaults to `current`, not `autoupdatingCurrent`. I'll add a comment about the default.

kylebshr:
Awesome! Matching SwiftUI makes sense

PR #106, opened 2020-06-04
Add `UIViewElement` to make it easier to wrap existing UIViews

kyleve:
This PR introduces `UIViewElement`, and `Element` conforming protocol that makes it trivial to implement Elements which wrap UIViews which can size themselves. A basic example would look like this:

```
struct Label : UIViewElement {

    var text : String
    var font : UIFont

    typealias UIViewType = UILabel

    static func makeUIView() -> UILabel {
        UILabel()
    }

    func updateUIView(_ view: UILabel) {
        view.text = self.text
        view.font = self.font
    }
}
```

In theory this will make types like `Label`, `AttributedLabel`, `TextField`, etc, a lot easier to implement.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -269,6 +273,67 @@ private struct EnvironmentAdaptingStorage: ContentStorage {
     }
 }
 
+private struct WrappedUIViewStorage : ContentStorage {
+    
+    let element : WrappedUIViewElement
+    
+    // MARK: ContentStorage
+
+    let childCount: Int = 1
```
kyleve, line 282: This should be zero

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -269,6 +273,67 @@ private struct EnvironmentAdaptingStorage: ContentStorage {
...
+                    return new
+                }
+            }()
+            
+            viewDescription.apply(to: view)
+            
+            return view.sizeThatFits(bounds.size)
+        }
+        
+        struct Key : Hashable
```
kyleve, line 329: private

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,156 @@
...
+    /// You usually do not need to implement this method – the default implementation of
+    /// this method simply calls `sizeThatFits(_:)` on the provided view.
+    ///
+    /// The view is fully configured and updated before this method is called – you do not need to
+    /// update it in any way.
+    ///
+    /// When To Override
+    /// ----------------
+    /// You may want to override this method if you need to mutate the value returned from `sizeThatFits(_:)`,
+    /// for example to round it up to an integral pixel value in the case of a `UILabel`, or if you
```
watt, line 79: I think this is probably a bad example to have in the docs, for two reasons:
a) UILabel does its own rounding, so this isn't a valid use case
b) I don't want to encourage folks to make their own label wrappers

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    /// Provide the view for the element.
+    func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        UIViewType.describe {
+            $0.builder = {
+                Self.makeUIView()
+            }
+            
+            $0.apply {
+                self.updateUIView($0)
```
watt, line 111: I will continue to rally against the use of shorthand arguments in any block that's more than a single expression. This would be much clearer to anyone reading this code as:

```swift
UIViewType.describe { config in
  config.builder = Self.makeView
  config.apply { view in
    self.updateUIView(view)
  }
}
```

The shorthand isn't gaining anything here, it's just obfuscating the semantic meaning of arguments.

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+    
+    //
+    // MARK: Element
+    //
+    
+    /// Defer to the reused measurement view to provide the size of the element.
+    var content: ElementContent {
+        ElementContent {
+            UIViewElementMeasurer.shared.measure(element: self, in: $0)
```
watt, line 99: Is there any technical reason we can't build this into _this type_, instead of handing off to a global singleton?

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,156 @@
...
+    associatedtype UIViewType : UIView
+        
+    /// Create and return a new instance of the provided view type.
+    ///
+    /// Note
+    /// ----
+    /// Ensure that you do not pass any values to the initializer of your view type
+    /// that you cannot also update in `updateUIView(_:)`, as view instances
+    /// are reused for sizing and measurement.
+    static func makeUIView() -> UIViewType
```
watt, line 50: Instead of creating a new abstraction for making/updating a view, why not use ViewDescription?

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Oh right – UILabel rounds, the text measurement stuff doesnt. Always mix those up.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Wanted to match the SwiftUI API, which IMO is a lot clearer and more concise to write.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: I liked cleanly separating the cache out into its own type. You can't put static vars onto protocols or their extensions, so the cache type seemed a reasonable place to put it.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: As well, we wouldnt have view description (likely) if we did not have to type erase to element – a lot of the value of ViewDescription is the type safety of the creation and usage of the view. Since we can use an associated type here, it's clearer to just have two methods.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: I can change it, but again, it's pretty obvious what's happening from reading the code if you've ever seen Swift before...

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,152 @@
...
+            config.apply { view in
+                self.updateUIView(view)
+            }
+        }
+    }
+}
+
+
+/// An internal type which caches `UIViewElement` views to be reused for sizing and measurement.
+final class UIViewElementMeasurer {
```
kyleve, line 115: private

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Updated this to reference UIImageView instead

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,152 @@
...
+
+
+///
+/// An element type  which makes it easier to wrap an existing `UIView` instance that
+/// provides its own sizing via `sizeThatFits`. An instance of the view is used for
+/// sizing and measurement, so that you do not need to re-implement your own measurement.
+///
+/// Example
+/// -------
+/// If you were implementing a very basic `Image` element, your implementation would look something
```
watt, line 18: `Image` is also kind of a bad example, because it's easy to measure the size of an image without creating a UIImageView.

I think we should strive to make the docs clear that this is not the first tool you should reach for when building a new element. If you can measure the size without a view instance, that is preferable. This helps in the cases where that's impossible (because you don't own the view and the measurement is a black box) or cumbersome (because it's an existing composite view with complex logic you don't want to extract).

Maybe something like a fictitious `LegacyBarGraphView` for an example?

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -0,0 +1,152 @@
...
+//
+//  Created by Kyle Van Essen on 6/4/20.
+//
+
+import UIKit
+
+
+///
+/// An element type  which makes it easier to wrap an existing `UIView` instance that
+/// provides its own sizing via `sizeThatFits`. An instance of the view is used for
```
watt, line 13: We should make it clear that we are caching the instance indefinitely, not creating one every time.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Every example has a positive and negative. I chose this and previous example of label because it's pretty immediately obvious to the reader what the model and view backing it is

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: will update

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: I think UISwitch is a good example to use here; gonna switch to that, since I've run into it before

PR #107, opened 2020-06-05
Fixes to get POS working for 0.12.0

kyleve:
- Pass BlueprintView's environment through in sizing methods.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -50,8 +54,8 @@ extension ElementContent {
 }
 
 extension ElementContent {
-    public init(build builder: @escaping (Environment) -> Element) {
-        self.storage = LazyStorage(builder: builder)
+    public static func withEnvironment(_ builder: @escaping (Environment) -> Element) -> ElementContent {
```
kyleve, line 57: Swapped these out b/c the two init methods that take closures cause conflicts which confuse the compiler. Pushing them behind a static func makes the intention clearer and prevents future issues.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -84,19 +88,20 @@ extension ElementContent {
     }
 
     /// Initializes a new `ElementContent` with no children that delegates to the provided measure function.
-    public init(measureFunction: @escaping (SizeConstraint) -> CGSize) {
+    public static func withMeasurement(_ measureFunction: @escaping (SizeConstraint) -> CGSize) -> ElementContent {
```
kyleve, line 91: Swapped these out b/c the two init methods that take closures cause conflicts which confuse the compiler. Pushing them behind a static func makes the intention clearer and prevents future issues.

watt: These can be disambiguated by:
- providing the argument types, e.g. `ElementContent { constraint -> CGSize in ... }`
- not using a trailing closure

I'm not as keen on the static funcs; they're not very discoverable with all the inits on `ElementContent`. It's kind of weird to have `ElementContent(measurable: Measurable)` as an init but for the closure version to be a static factory.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -14,13 +14,17 @@ public struct ElementContent {
...
+        self.storage = storage
+    }
 
     /// Measures the required size of this element's content.
     /// - Parameters:
     ///   - constraint: The size constraint.
     ///   - environment: The environment to measure in.
     /// - returns: The layout size needed by this content.
-    public func measure(in constraint: SizeConstraint, environment: Environment) -> CGSize {
+    public func measure(in constraint: SizeConstraint, environment: Environment = .empty) -> CGSize {
```
watt, line 27: I don't think this should have a default. It's not usually called with `.empty`, and this opens up a class of mistakes.

BlueprintUI/Sources/Layout/Empty.swift
```diff
@@ -16,7 +16,7 @@ public struct Empty : Element {
     public init() {}
     
     public var content: ElementContent {
-        ElementContent { _ in
+        .withMeasurement { _ in
```
watt, line 19: Calls that aren't using the constraint could be replaced with `ElementContent(intrinsicSize:)` to dodge this entirely.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: Good point, removed

BlueprintUI/Sources/Layout/Empty.swift
kyleve: Changed

PR #111, opened 2020-06-09
Replace 'master' with 'main'

kyleve:


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Master]
+## [Main]
```
bencochran, line 8: It looks like GitHub no longer turns this into a link. Know what’s up with that?

kyleve: Unsure... maybe something to do with how the preview renders? Let's see what happens when it's merge...

kyleve: Hmm, still renders wrong even after merging... wonder if [main] would work; or if they've just special-cased "Master"

watt: These are [shortcut reference links](https://github.github.com/gfm/#shortcut-reference-link). There is a link reference named `[master]` at the bottom of this doc that defines the URL.

With a regular reference link, you define the URL elsewhere in the document and reference it, to [make the text more readable](https://daringfireball.net/projects/markdown/syntax#link). On Github, you can omit the reference if it matches the link text, e.g. `[foo]` is shorthand for `[foo][foo]`.

kyleve: TIL! Also explains why I didnt find that last `master`, since github hides it from the markdown view.

PR #113, opened 2020-06-17
[WIP] [DNR] Super rough update transition impl

kylebshr:
Big todos:
- Support `inherited`/`none` animations
- More blueprint-y way to build animations
- Support animating just some properties? https://github.com/square/Blueprint/pull/114/files#diff-11325f6dbed0b93fc2efe1a845704c0eR14


BlueprintUI/Sources/View Description/UpdateTransition.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by Kyle Bashour on 6/15/20.
+//
+
+import Foundation
+
+public struct UpdateTransition {
+
+    private var _perform: (UIView, @escaping () -> Void) -> Void
+
+    public init(perform: @escaping (UIView, @escaping () -> Void) -> Void) {
```
kyleve, line 14: I get exposing the view accessor for the custom case, but it'd be nice to make standard UIView animations easier. Eg, similar to this sort of thing: https://github.com/square/Blueprint/pull/94/files#diff-57cf1fe089b34d04ae53d570db4238f5R44

PR #115, opened 2020-06-24
Various performance improvements

kyleve:
This PR introduces a few performance optimizations. In order of importance:

### Per-layout pass opt-in measurement caching
`Environment` receives a new property, `measurementCache`, an instance of `MeasurementCache` (https://github.com/square/Blueprint/pull/115/files#diff-a9579b365c090152d977e2968b3c009d), which allows caching the measurement of elements which are generally value types like Labels. This results in a significant speedup for deeper element hierarchies, which re-measurement elements deeper in the tree many many times during a layout pass.

You opt into this caching by providing a `MeasurementCachingKey` to your `ElementContent` on `Element`, or by providing one directly on your `UIViewElement`.

### `BlueprintView` smarter during diffs
`BlueprintView` now avoids diffing empty child trees, and avoids needlessly trying to update subview indexes when nothing changes. There is further optimization we can do here, but this was good enough for now.

### `ElementPath` no longer has an inner instance
`ElementPath` used to COW an inner `Storage` type. This used to be faster in older versions of Swift, but is not anymore.

### What does this add up to?

In this sample element:

```
        let elements = [
            TestLabel(text: "Lorem ipsum dolor sit amet, consectetur adipiscing elit."),
            TestLabel(text: "Integer molestie et felis at sodales."),
            TestLabel(text: "Donec varius, orci vel suscipit hendrerit, risus massa ornare dui, at gravida elit sapien at lorem."),
            TestLabel(text: "Nunc in ipsum porttitor, tincidunt est eu, euismod odio."),
            TestLabel(text: "Duis posuere nunc sed mi auctor, in dictum elit iaculis."),
            TestLabel(text: "Ut vel varius est. Duis efficitur vel lorem quis tempor."),
            TestLabel(text: "Nulla porttitor, mi nec posuere bibendum, turpis ipsum ultrices tortor, a placerat sapien augue quis sem."),
            TestLabel(text: "Cras volutpat nisl vitae elit convallis, quis tempor massa faucibus."),
        ]
        
        let stack = Column { col in
            let row = Row { row in
                let col = Column { col in
                    elements.forEach {
                        col.add(child: $0)
                    }
                }
                
                for _ in 1...10 {
                    row.add(child: col)
                }
            }
            
            for _ in 1...10 {
                col.add(child: row)
            }
        }
```

The speedup is fairly obvious:

Before:

```
Iteration: 1, Duration : 0.9335100650787354
....
Iteration: 12, Duration : 0.8129500150680542
Iterations: 12, Average Time: 0.8496990899244944
```

After:

```
Iteration: 1, Duration : 0.11801004409790039
...
Iteration: 120, Duration : 0.08103799819946289
Iterations: 120, Average Time: 0.08354394237200419
```

Which adds up to a 10x / 90% speedup from the previous uncached version.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -128,7 +175,7 @@ fileprivate protocol ContentStorage {
 
 extension ElementContent {
 
-    public struct Builder<LayoutType: Layout> {
+    public struct Builder<LayoutType: Layout> : ContentStorage {
```
kyleve, line 156: I consolidated some extensions here

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -264,6 +264,12 @@ extension BlueprintView {
         fileprivate func update(node: NativeViewNode, appearanceTransitionsEnabled: Bool) {
             
             assert(node.viewDescription.viewType == type(of: view))
+            
+            // Bail out fast if we do not have any children to manage.
+            // This is a performance optimization for leaf elements.
+            guard self.children.isEmpty == false && node.children.isEmpty == false else {
```
kyleve, line 270: Note: We could also optimize single child elements, but I didnt do that here. It'd be a good followup, as many items are single child like Inset.

BlueprintUI/Sources/Internal/ElementPath.swift
```diff
@@ -1,31 +1,28 @@
 /// Represents a path into an element hierarchy.
 /// Used for disambiguation during diff operations.
-struct ElementPath: Hashable {
+struct ElementPath: Hashable, CustomDebugStringConvertible {
```
kyleve, line 3: The hop to the inner instance type is no longer faster; in fact its slower than just using a native struct.

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -34,10 +34,16 @@ public struct Label: Element {
     }
 
     public var content: ElementContent {
-        var element = AttributedLabel(attributedText: attributedText)
+        let text = self.attributedText
+        
+        var element = AttributedLabel(attributedText: text)
         element.numberOfLines = numberOfLines
```
kyleve, line 40: Need to include these in the cache as well!

BlueprintUI/Tests/PerformancePlayground.swift
```diff
@@ -0,0 +1,122 @@
...
+
+import XCTest
+@testable import BlueprintUI
+
+
+class PerformancePlayground : XCTestCase
+{
+    override func invokeTest() {
+        // Uncomment this line to run performance metrics, eg in Instruments.app.
+        super.invokeTest()
```
kyleve, line 16: Comment me out again

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -31,7 +31,10 @@ public struct AttributedLabel: Element {
             }
         }
 
-        return ElementContent(measurable: Measurer(attributedText: attributedText, roundingScale: roundingScale))
+        return ElementContent(
+            measurable: Measurer(attributedText: attributedText, roundingScale: roundingScale),
+            measurementCachingKey: .init(type: AttributedLabel.self, input: self)
```
kyleve, line 36: Need to cache # of lines too

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: (Nothing changed otherwise)

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -50,7 +62,11 @@ extension ElementContent {
 }
 
 extension ElementContent {
-    public init(build builder: @escaping (Environment) -> Element) {
+    public init(
```
kylebshr, line 65: Should document

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -61,8 +77,8 @@ extension ElementContent {
     ///
     /// - parameter element: The single child element.
     /// - parameter layout: The layout that will be used.
-    public init(child: Element, layout: SingleChildLayout) {
-        self = ElementContent(layout: SingleChildLayoutHost(wrapping: layout)) {
+    public init(child: Element, layout: SingleChildLayout, measurementCachingKey : MeasurementCachingKey? = nil) {
```
kylebshr, line 80: Missing doc for measurementCachingKey param

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -72,26 +88,28 @@ extension ElementContent {
     /// The given element will be used for measuring, and it will always fill the extent of the parent element.
     ///
     /// - parameter element: The single child element.
-    public init(child: Element) {
-        self = ElementContent(child: child, layout: PassthroughLayout())
+    public init(child: Element, measurementCachingKey : MeasurementCachingKey? = nil) {
```
kylebshr, line 91: Missing doc

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -72,26 +88,28 @@ extension ElementContent {
...
+        self = ElementContent(child: child, layout: PassthroughLayout(), measurementCachingKey: measurementCachingKey)
     }
 
     /// Initializes a new `ElementContent` with no children that delegates to the provided `Measurable`.
-    public init(measurable: Measurable) {
+    public init(measurable: Measurable, measurementCachingKey : MeasurementCachingKey? = nil) {
```
kylebshr, line 96: Missing doc

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -72,26 +88,28 @@ extension ElementContent {
...
             layout: MeasurableLayout(measurable: measurable),
-            configure: { _ in })
+            measurementCachingKey: measurementCachingKey,
+            configure: { _ in }
+        )
     }
 
     /// Initializes a new `ElementContent` with no children that delegates to the provided measure function.
-    public init(measureFunction: @escaping (SizeConstraint) -> CGSize) {
+    public init(measurementCachingKey : MeasurementCachingKey? = nil, measureFunction: @escaping (SizeConstraint) -> CGSize) {
```
kylebshr, line 105: Missing doc

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -11,12 +11,22 @@ public struct SizeConstraint: Hashable {
 
     /// The height constraint.
     public var height: Axis
+    
+    private let hash : Int
```
kylebshr, line 15: This won't be updated if with or height is mutated

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr, line 36: I think you can use `Self` here now, which just seems "nicer" IMO (and doesn't need updating if the name of the type changes)

```suggestion
            measurementCachingKey: .init(type: Self, input: self)
```

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -33,15 +34,7 @@ public struct Label: Element {
...
-        var element = AttributedLabel(attributedText: attributedText)
-        element.numberOfLines = numberOfLines
-        element.roundingScale = roundingScale
-        return ElementContent(child: element)
-    }
-
-    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
-        return nil
+    public var elementRepresentation: Element {
+        AttributedLabel(attributedText: attributedText)
```
kylebshr, line 38: Still need to set number of lines and rounding scale, no?

BlueprintUI/Sources/Measuring/SizeConstraint.swift
kyleve: Removing this; it's superceded by the caching in the keys anyways

BlueprintUICommonControls/Sources/Label.swift
kyleve: Doh! Righto

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Ah right, I always forget where this works; since it gradually got more and more permeated through a few versions of Swift.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -4,15 +4,25 @@ import UIKit
...
 
     private let storage: ContentStorage
+    
+    /// The key to use to cache measurement values.
+    private let measurementCachingKey : MeasurementCachingKey?
 
     /// Initializes a new `ElementContent` with the given layout and children.
     ///
     /// - parameter layout: The layout to use.
+    /// - parameter measurementCachingKey: The key to use to cache measurement values.
```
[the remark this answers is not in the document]
kyleve: Will include in `MeasurementCachingKey` to avoid duping everywhere

kyleve:
> I think this looks good - just two small notes
> 
> * I wish the cache wasn't special-cased in the env, but if it's actually faster it makes sense. Did you check that?
> * An Environment is tied to a specific bp view and dealloced when it is, right? Just want to make sure we don't have an infinitely growing cache for views that no longer exist. And if we keep pushing towards a single environment for all BP views to share, we may want to make sure to remove cached sizes that are no longer needed somehow.

1) Yes – it wasn't substantial, but it was there. This seemed fine given this is also a special case use of Env.

2) This is per layout pass only; the Env is made from scratch each time in `makeEnvironment()` in the Blueprint view.

kylebshr:
Oh that's right.. would there be a meaningful improvement if we could persist that cache across layouts?

kyleve:
> Oh that's right.. would there be a meaningful improvement if we could persist that cache across layouts?

I can try it and see what happens, will report back.

kyleve:
Per layout:
Iterations: 111, Average Time: 0.09038803598902247

Cached:
Iterations: 108, Average Time: 0.093000962778374

So, given the noise of the test, barely any diff.

kylebshr:
Cool, thanks for checking!

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -96,7 +105,15 @@ public extension UIViewElement {
...
     /// Defer to the reused measurement view to provide the size of the element.
     var content: ElementContent {
-        ElementContent {
+        let key : MeasurementCachingKey? = {
+            if let key = self.measurementCacheKey {
+                return .init(type: Self.self, input: key)
+            } else {
+                return nil
+            }
+        }()
```
watt, line 114: This looks like a job for `Optional.map`:

```swift
let key = self.measurementCacheKey.map { key in
  MeasurementCachingKey(type: Self.self, input: key)
}
```

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -57,4 +59,9 @@ public struct Environment {
             values[ObjectIdentifier(key)] = newValue
         }
     }
+    
+    /// Cache used to speed up measurements during a layout pass.
+    /// This is declared on the `Environment` directly, because access to the `measurementCache` is a hot path
+    /// which we can optimize by not needing the usual `env.values[...]` lookup.
```
watt, line 65: Surprised to see this. How slow is it? As a dictionary lookup I would expect it to be quite fast.

BlueprintUI/Sources/Element/MeasurementCachingKey.swift
```diff
@@ -0,0 +1,114 @@
...
+        var hasher = Hasher()
+        hasher.combine(self.elementType)
+        hasher.combine(self.input)
+        self.hash = hasher.finalize()
+    }
+    
+    public func hash(into hasher: inout Hasher) {
+        // We pre-compute the hash above to avoid potentially expensive work here.
+        // Computing the hash repeatedly for `input` may be slow (especially for strings).
+        hasher.combine(self.hash)
```
watt, line 112: I see MeasurementCache.Key also does this — do we need both?

BlueprintUI/Tests/PerformancePlayground.swift
```diff
@@ -0,0 +1,122 @@
...
+
+import XCTest
+@testable import BlueprintUI
+
+
+class PerformancePlayground : XCTestCase
+{
+    override func invokeTest() {
+        // Uncomment this line to run performance metrics, eg in Instruments.app.
+        // super.invokeTest()
```
watt, line 16: Does this prevent these tests from running normally? Since they perform no assertions they're probably not useful in CI.

BlueprintUI/Tests/PerformancePlayground.swift
```diff
@@ -0,0 +1,122 @@
...
+            let iterationStart = Date()
+            block()
+            let iterationEnd = Date()
+            let duration = iterationEnd.timeIntervalSince(iterationStart)
+            
+            iterations += 1
+
+            print("Iteration: \(iterations), Duration : \(duration)")
+            
+        } while Date() < start + seconds
```
watt, line 90: Some misc notes on benchmarking that you are free to ignore:

- Often the first run is thrown away, because it "warms" things and is a statistical outlier.
- `CFAbsoluteTimeGetCurrent` might be a better choice than `Date()`. Or `CACurrentMediaTime`, which is monotonically increasing.
- If you run a set number of iterations and time the whole thing, you can avoid checking the time between each iteration.

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Truuuuu

BlueprintUI/Sources/Environment/Environment.swift
kyleve: It wasn't a lot, but it was in the hot path, so it was on the Instruments trace

BlueprintUI/Sources/Element/MeasurementCachingKey.swift
kyleve: Ooh good call; yeah we only need this one, I think...

BlueprintUI/Tests/PerformancePlayground.swift
kyleve: Yeah, this cuts them off from CI

BlueprintUI/Tests/PerformancePlayground.swift
kyleve: I intentionally only wanted to run the tests for a given amount of time; so going to leave this as is.

kyleve:
> Mind rebasing to squash the commit history?

Oh yeah for sure... this is my default state before I merge in master or merge

PR #118, opened 2020-06-26
Introduce PointerInteraction type for wrapping UIPointerInteraction

kyleve:
This PR introduces a simple way to use `UIPointerInteraction` for Blueprint elements. You will usually just call `.pointerInteraction()` on your element, and that will apply the desired hover effect.


BlueprintUICommonControls/Sources/PointerInteraction.swift
```diff
@@ -0,0 +1,303 @@
...
+        with style : @escaping PointerInteraction.StyleProvider = { _, _ in .automatic }
+    ) -> Element {
+        PointerInteraction(self, with: style)
+    }
+}
+
+
+extension PointerInteraction {
+    
+    public enum Style {
```
kyleve, line 90: Document me

BlueprintUICommonControls/Sources/PointerInteraction.swift
```diff
@@ -0,0 +1,303 @@
+//
+//  PointerInteraction.swift
+//  BlueprintUICommonControls
```
kyleve, line 3: Thoughts on in here vs in BlueprintUI? I think BlueprintUI might actually make more sense since this isn't really a control; it's a general part of UIKit you can apply to any view.

BlueprintUICommonControls/Sources/PointerInteraction.swift
```diff
@@ -0,0 +1,303 @@
...
+    /// How to provide the style based on the coordinate space and pressed keys.
+    public var style : StyleProvider
+    
+    /// Creates a new `PointerInteraction` instance that wraps the provided element,
+    /// styling the interaction via the `StyleProvider`.
+    ///
+    /// You don't need to provide a `StyleProvider` – the default one will simply
+    /// return a `.automatic` style.
+    ///
+    /// **Note** – On iOS versions before 13.4, this has no effect.
```
kylebshr, line 38: If you do ` - Note: Foo` it makes a header in the rendered docs. Just fyi - no need to update unless you want to.

Screen Shot 2020-10-27 at 2 01 16 PM

kylebshr: (**Discussion** is automatic, for any docs after a newline)

BlueprintUICommonControls/Sources/PointerInteraction.swift
```diff
@@ -0,0 +1,303 @@
...
+///
+/// **Note** – On iOS versions before 13.4, this has no effect.
+///
+public struct PointerInteraction : Element
+{
+    /// The element to which the pointer action will be applied.
+    public var wrapping : Element
+    
+    /// How to provide the style based on the coordinate space and pressed keys.
+    public typealias StyleProvider = (UICoordinateSpace, UIKeyModifierFlags) -> Style
```
kylebshr, line 27: Could we design this so you don't need a closure/builder unless you're actually building a style that needs those arguments? so you can do, e.g., 

```swift
element.pointerInteraction(style: .hidden)
element.pointerInteraction(style: .shape { coordinates, flags in
    // do the thing
})
```

it's small but nice IMO. If it seems weird/annoying to shape everything so this works, don't worry about it

BlueprintUICommonControls/Sources/PointerInteraction.swift
kyleve: Agree, will do

BlueprintUICommonControls/Sources/PointerInteraction.swift
kyleve: Ahh true, good call – I can add a helper method for that – should be just a couple lines

BlueprintUICommonControls/Sources/PointerInteraction.swift
kyleve: Oh yeah... I go back and forth on this, since the `- Note: Foo` is more immediately obvious in the generated docs, but the `**note**` when reading the code directly is more immediately obvious. I can convert it though, since I bet `- **Note**: Foo` works too.

BlueprintUI/Sources/UIKit/PointerInteraction.swift
```diff
@@ -0,0 +1,309 @@
...
+import UIKit
+
+
+/// An `Element` which lets you add `UIPointerInteraction` interactions
+/// to the contained `Element`.
+///
+/// A `UIPointerInteraction` allows you to respond to hover events for buttons,
+/// text, icons, etc, when a user is using an iPad that has support for a pointer, usually from a trackpad.
+///
+/// - Note: On iOS versions before 13.4, this has no effect.
```
kylebshr, line 17: Should we just make this available in 13.4+? If you're using it in your project a lot and target less than that, it's easy enough to add your own wrapper that does nothing on < 13.4. Seems a little weird to have that built in though.

kyleve: I don't think so. It's annoying to have to turd your code with these checks, when the end result you want is a no-op on earlier iOS versions anyway.

BlueprintUICommonControls/Sources/PointerInteraction.swift
watt: Feels more like a CommonControls thing to me, for the same reason that's where Tappable lives. It's not part of the core layout stuff.

kyleve: In my brain, CommonControls is more of "these are the vanilla UIKit controls" that are basically the Blueprint version of UIKit. But things that on their own dont have a visual style like button, imo, should live in just core Blueprint. Eg, I'd move things like `ScrollView` into core Blueprint.

PR #120, opened 2020-07-07
Label measuring and other enhancements

### LINKED ISSUE #93: AttributedLabel measuring is wrong sometimes
# Problems

There are a couple of related problems with the `AttributedLabel` element (and by extension, `Label`).

1. Measurement is wrong when `numberOfLines` is set to a value other than `0`.

    The current measurement behavior uses [`NSAttributedString.boundingRect(with:options:context:)`](https://developer.apple.com/documentation/foundation/nsattributedstring/1529154-boundingrect), which has no affordance for line limits.
    
2. Rounding assumes the main screen's scale by default, and will be wrong if rendered on a different screen.

   This one is definitely an edge case, but we should at least consider solving this if it's feasible.

# Potential solutions

## Static UILabel for measuring

We solved problem 1 internally by using a static prototype instance of `UILabel`. During measuring we apply the view description to the label and then call `sizeThatFits`. This solution operates on the assumption that measuring is always done on the main thread, so two labels cannot be measured concurrently.

This does not solve problem 2: if `UILabel` is not in a window, it returns results rounded to the main screen's scale.

## TextKit & UITextView

TextKit measuring methods allow us to do line limits, but in experimentation I have found that TextKit measuring doesn't match `UILabel`. In fact, it's quite difficult (maybe impossible, without knowing its internals) to accurately reproduce its measuring behavior across every combination of line break mode and line limit.

It's important to get those values right — if they're off slightly, it can cause text to be truncated in places where it should have fit. If we want to use TextKit we may have to switch to `UITextView` or do our own string rendering to ensure it matches our measuring.

We could also potentially switch between using a `UILabel` for single-line text and a `UITextView` for multi-line text, or offer these as separate elements.

Since TextKit does not automatically do any rounding, we'd be free to do this ourselves (perhaps by passing the screen scale down through the Environment).

## Custom rendering (hard mode)

As mentioned above, we could side-step UIKit entirely and use TextKit/CoreText to measure and then render strings.

This seems like overkill but it's an option.

watt:
This PR fixes #93 by taking the approach of using a UILabel for measuring.

This means the `roundingScale` property is no longer used, so it's been stripped out along with some associated tests.

It also bundles some other changes:

- add `isAccessibilityElement` to Label/AttributedLabel
- add `lineHeight` in Label, which includes an alignment setting to match some non-iOS ideas of what line height means
  - `top`: adds extra space at the bottom of the line, like traditional leading
  - `center`: adds half leading on both sides, and matches the CSS model of line height
  - `bottom`: adds extra space the top, which is the default iOS behavior
- add `textRectOffset` to AttributedLabel, which is used to support line height alignment modes, since there does not seem to be a way to do it with only NSAttributedString attributes.

Apologies for not making that stuff a separate PR.


BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -23,69 +22,79 @@ class AttributedLabelTests: XCTestCase {
...
+            expectedSize: CGSize(width: 30, height: 235.5))
+        test(
+            in: CGSize(width: 100, height: 300),
+            expectedSize: CGSize(width: 95, height: 105.5))
+        test(
+            in: CGSize(width: 120, height: 300),
+            expectedSize: CGSize(width: 107, height: 67))
+        test(
+            in: CGSize(width: 8000, height: 4000),
+            expectedSize: CGSize(width: 153.5, height: 38.5))
```
watt, line 72: These tests used to assert against NSAttributedString measurements.

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -23,69 +22,79 @@ class AttributedLabelTests: XCTestCase {
 
         let string = NSAttributedString(string: "Hello, world. This is some long text that runs onto several lines.")
         var element = AttributedLabel(attributedText: string)
-        element.roundingScale = 1
+
+        func constrained() -> Element {
+            element.constrainedTo(width: .atMost(100), height: .atMost(800))
```
watt, line 27: These tests are now auto-sized instead of using an explicit frame, so they also test that the label measuring is correct.

BlueprintUICommonControls/Tests/Sources/XCTestCaseExtensions.swift
```diff
@@ -104,6 +104,8 @@ extension XCTestCase {
             view.frame = CGRect(origin: .zero, size: size)
         } else {
             view.sizeToFit()
+            view.frame.size.width.round(.up, by: scale)
+            view.frame.size.height.round(.up, by: scale)
```
watt, line 108: This allows snapshots to continue working in 1x scale even when the underlying elements are using the screen scale to measure.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -5,40 +5,60 @@ public struct AttributedLabel: Element {
...
-            config[\.attributedText] = attributedText
-            config[\.numberOfLines] = numberOfLines
+        return LabelView.describe { (config) in
+            config.apply(update)
         }
     }
+}
 
+extension AttributedLabel {
+    private final class LabelView: UILabel {
```
watt, line 51: Private type here means I couldn't use `UIViewElement`.

kyleve: me rn

tumblr_inline_n23y8zrALB1rz15mt

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -12,18 +12,24 @@ public struct Label: Element {
...
     }
     
     private var attributedText: NSAttributedString {
-        let paragraphStyle = NSParagraphStyle.default.mutableCopy() as! NSMutableParagraphStyle
+        let paragraphStyle = NSMutableParagraphStyle()
         paragraphStyle.alignment = alignment
         paragraphStyle.lineBreakMode = lineBreakMode
+
+        if case .custom(let lineHeight, _) = lineHeight {
+            paragraphStyle.minimumLineHeight = lineHeight
```
kyleve, line 29: Assume this defaults to a sane value without setting these defaults, right?

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -36,7 +42,24 @@ public struct Label: Element {
...
+        element.isAccessibilityElement = isAccessibilityElement
+
+        if case let .custom(lineHeight, lineHeightAlignment) = lineHeight {
+            switch lineHeightAlignment {
+            case .top:
+                let leading = lineHeight - font.lineHeight
+                element.textRectOffset = CGPoint(x: 0, y: -leading)
+
+            case .center:
+                let halfLeading = (lineHeight - font.lineHeight) / 2
```
kyleve, line 54: Does this need any rounding?

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -45,3 +68,27 @@ public struct Label: Element {
     }
 
 }
+
+extension Label {
+    public enum LineHeight {
```
kyleve, line 73: Equatable for what consumers do with em?

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -45,3 +68,27 @@ public struct Label: Element {
...
+            /// and matches the model used by CSS.
+            case center
+
+            /// Align text to the bottom of the available line height, with extra space added at the top.
+            /// This is the default behavior of `UILabel` on iOS.
+            case bottom
+        }
+
+        /// Use the default line height of the label's font.
+        case font
```
kyleve, line 89: ``default``?

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: (Though, wondering: Do we want to make this public so we don't need to re-implement this behaviour in MarketUI for other UIView-constructed views which need this same behaviour?)

kylebshr: Is there an attribute to make it importable without being public/automatically imported? Like, how does `import UIKit.UIGestureRecognizerSubclass` work?

kyleve: This would be a custom module map, which is definitely doable.

kyleve:
```
spec.module_map = 'source/module.modulemap'
```

And then in the module map do something like this: https://git.sqcorp.co/projects/IOS/repos/register/browse/Frameworks/RegisterUI/RegisterUI.modulemap

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -5,40 +5,60 @@ public struct AttributedLabel: Element {
...
     public var numberOfLines: Int = 0
-    /// The scale to which pixel measurements will be rounded. Defaults to `UIScreen.main.scale`.
-    public var roundingScale: CGFloat = UIScreen.main.scale
+    public var isAccessibilityElement = false
+
+    /// An offset that will be applied to the rect used by `drawText(in:)`.
+    ///
+    /// This can be used to adjust the positioning of text within each line's frame, such as adjusting
+    /// the way text is distributed within the line height.
+    public var textRectOffset: CGPoint = .zero
```
kylebshr, line 14: Would `UIOffset` be more appropriate here?

BlueprintUICommonControls/Sources/Label.swift
kylebshr: Yeah, if it's not set it just uses whatever the font has built in

kylebshr: (I'd still ask for a `switch`, in case there's more cases added)

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Doesn't seem worth the effort for a 12-line class to me. I'd rather copy it.

kyleve: I guess, but the module also isn't much more work, and it would be nice to keep these in sync. Oh well, too bad the swift SPI stuff isnt ready for primetime yet.

bencochran: Why wouldn’t something like MarketLabel be built on top of AttributedLabel rather than the underlying view?

kyleve: This is more referring to using the underlying Label view for the controls we have to build with UIKit because they're highly stateful, eg the floating labels in text fields.

BlueprintUICommonControls/Sources/Label.swift
watt: Nope, the text system handles that.

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Sure, I didn't know that was a thing!

BlueprintUICommonControls/Sources/Label.swift
watt: This is more explicit about what this case is, though? `default` doesn't tell you anything about the actual line height being used.

BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -12,18 +12,38 @@ public struct Label: Element {
...
 
     public init(text: String, configure: (inout Label) -> Void = { _ in }) {
         self.text = text
         configure(&self)
     }
+
+    private var effectiveLineBreakMode: NSLineBreakMode {
+        // These line break modes don't work when numberOfLines is 1, and they break line height adjustments.
+        // Normalize them to clipping mode instead.
+        if numberOfLines == 1 && (lineBreakMode == .byCharWrapping || lineBreakMode == .byWordWrapping) {
```
watt, line 26: I discovered that the text rect offset we're using to support line height is ignored by UILabel in these specific cases. UILabel does not "wrap" these cases at all, but treats them like clipping mode, so I think we can work around this by normalizing to clipping mode.

I've updated the snapshot tests for custom line heights to validate it works in various configurations.

BlueprintUICommonControls/Sources/Label.swift
watt: changed to switch

BlueprintUICommonControls/Sources/Label.swift
kyleve: Seeing `.font` inline seems more confusing to me than `.default`, imo, given iOS SDKs usually use `default` to refer to what the system would usually do, but this is also fine.

kyleve:
Still LGTM

PR #121, opened 2020-07-09
Fill Element

kyleve:
Adds a `Fill` element that allows children to take up all of the space provided by the parent element.


BlueprintUI/Sources/Layout/Fill.swift
```diff
@@ -0,0 +1,81 @@
...
+    enum Axes : Equatable {
+        case horizontal
+        case vertical
+        case both
+        
+        func measure(in constraint: SizeConstraint, child: Measurable) -> CGSize {
+            switch self {
+            case .horizontal:
+                let size = child.measure(in: constraint)
+                return CGSize(width: constraint.width.maximum, height: size.height)
```
kyleve, line 60: Not quite... need to re-measure based on the new width or height

kyleve:
Need to pull over tests from Gil: https://git.sqcorp.co/projects/IOS/repos/register/pull-requests/46150/diff#Frameworks/SquareCheckoutAppletCore/Tests/Unit/FillElementTests.swift

BlueprintUI/Sources/Layout/Fill.swift
kyleve: Actually, maybe not?

watt:
What is the use case where you would want this instead of Aligned with a `fill` option? It seems to me like an anti-pattern to have an element report its size as infinite.

kyleve:
This was an ask in The Blueprint channel a while back and just got around to copying over the tests from POS. The element isn't infinite; it takes up the maximum size provided by the parent.

kyleve:
See existing usages in POS here: https://git.sqcorp.co/projects/IOS/repos/register/browse/Frameworks/SquareCheckoutAppletCore/Sources/OrderEntry/Header/OrderEntryHeaderScreen.swift#53

kyleve:
Ping on this

watt:
> This was an ask in The Blueprint channel a while back and just got around to copying over the tests from POS. The element isn't infinite; it takes up the maximum size provided by the parent.

The measured size is infinite if the constraint is `unconstrained`. So it will result in infinite sizes from `intrinsicContentSize` as well as `sizeThatFits(.zero)`.

The use case in POS seems like a kludge for getting custom title views in BackStackContainer/RGUIImposterNavigationBar to fill the header. We could probably solve that more cleanly with something like a flag on the nav item to indicate that it should be given all available space, rather than centered-size-to-fit. I don't think I would recommend this for general purpose use.

kyleve:
That's true for a fair number of elements though; eg even the Blueprint text field can be infinite width if given infinite width; https://github.com/square/Blueprint/blob/main/BlueprintUICommonControls/Sources/TextField.swift#L74.

kylebshr:
I don't have a strong opinion here - I think for some elements, it makes sense that measuring in an unconstrained constraint results in an unconstrained size.. it's literally meant to fill the given size. It kinda sucks that `intrinsicContentSize` and `sizeThatFits` don't exactly make sense with that approach though

kyleve:
Bumping this, it came up again in Slack: https://square.slack.com/archives/CKSU9GBFS/p1606945883420300

kyleve:
Running into a couple more cases where this would make things a lot easier – re-opening.

BlueprintUI/Sources/Layout/Fill.swift
```diff
@@ -0,0 +1,97 @@
...
+                
+            case .vertical:
+                let size = child.measure(in: constraint)
+                return CGSize(width: size.width, height: constraint.height.maximum)
+                
+            case .both:
+                return constraint.maximum
```
kyleve, line 83: Rather than just .maximum, if the axis is unconstrained, just measure the element.

BlueprintUI/Sources/Layout/Fill.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+//  Created by Kyle Van Essen on 7/9/20.
+//
+
+import UIKit
+
+
+/// An element which wraps a child element, and takes up all the available space vended to it by its
+/// parent element. Use this when you want an element to take up all the space it possibly can during a layout pass.
+public struct Fill : Element {
```
kyleve, line 13: FillParent

BlueprintUI/Sources/Layout/Fill.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+}
+
+
+public extension Element {
+    
+    /// Wrap the element in a `Fill` element, so that it will take up the maximum space it is allowed to
+    /// during layout and measurement passes.
+    func fill(along axes : Fill.Axes = .both) -> Fill {
```
kyleve, line 94: fillParent

BlueprintUI/Sources/Layout/Fill.swift
n8chur: Do you mean for each Axes case or only in the `.both` scenario?

kyleve: For all the cases here  – eg fill means nothing in an unconstrained measurement, so if either axes is unconstrained, a normal measurement is likely smartest.

PR #122, opened 2020-07-12
Skip update path if element remains nil

kyleve:
This is a minor perf optimization I noticed in Listable; in particular we'd still go through the update path even if the element for the background view or selected background view remained nil (which is usually the case). This saves a bit of time, which adds up during scroll events as cells get recycled (about 1-2ms in testing; you only get 16ms to maintain 60fps scrolling).


kyleve:
Bump on reviews for this

PR #123, opened 2020-07-20
Add extension to Element for wrapping in a Button

connorcimowsky:


kyleve:
This conceptually LGTM; this this for use in POS? Eg I thought this might have some styling behaviour we don't want by default.

watt:
I won't block this, but conceptually I think this different from the "modifier"-like elements like Inset, and shouldn't have an extension like this. There's no `button()` view modifier in SwiftUI.

bencochran:
Hmm, yeah that’s a good point. I could see this sticking with a more structural-style constructor layout with `tappable` being the purely-modifier one…

PR #126, opened 2020-07-29
Add AccessibilityContainer element

kylebshr:
Adds an element that recursively searches its subviews and acts as a container for any views where `isAccessibilityElement` is true. This is useful for groups of multiple elements that are each accessible, where your layout would prevent them from being navigated in order.

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -48,7 +48,7 @@ extension AccessibilityContainer {
     private final class AccessibilityContainerView: UIView {
         override var accessibilityElements: [Any]? {
             get { recursiveAccessibleSubviews() }
-            set { fatalError() }
+            set { fatalError("This property is not settable") }
         }
     }
 }
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -1,0 +1,68 @@
+import BlueprintUI
+import UIKit
+
+/// Acts as an accessibility container for any accessible subviews.
+///
+/// Accessible subviews are found using the following algorithm:
+///
+/// Recurse subviews until a view is found that either
+/// - has`isAccessibilityElement` set to `true` or
+/// - returns a non-nil value from `accessibilityElements` (i.e., is a container itself)
+///
+/// If an accessibility element is found, we add it to the `accessibilityElements`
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -0,0 +1,53 @@
+import BlueprintUI
+import UIKit
+
+/// Acts as an accessibility container for any subviews
+/// where `isAccessibilityElement == true`.
+public struct AccessibilityContainer: Element {
```
kyleve, line 6: Would be good to flesh this comment out a bit more

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -0,0 +1,53 @@
...
+    func accessibilityContainer() -> Element {
+        AccessibilityContainer(wrapping: self)
+    }
+}
+
+extension AccessibilityContainer {
+    private final class AccessibilityContainerView: UIView {
+        override var accessibilityElements: [Any]? {
+            get { recursiveSubviews().filter { $0.isAccessibilityElement } }
+            set { fatalError() }
```
kyleve, line 41: Maybe a real error message on this?

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+
+extension UIView {
+    fileprivate func recursiveSubviews() -> [UIView] {
```
kyleve, line 47: Random Q: I forget how extensions on ObjC types work in Swift; do we need to prefix this like you would in ObjC land to avoid collisions?

watt: Should be safe if it's not `@objc`

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
kylebshr: Agreed, I updated the behavior and rounded out the doc comment more; open to any improvements to the wording, describing "computer science" is not my forte

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
kylebshr: Sure (though it's impossible to hit unless you're editing this implementation)

kylebshr:
Added tests!

PR #127, opened 2020-07-30
Expose setting font for TextField

lechristian:
Add font property to TextField. The default lazy loaded UIFont for UITextField is `.preferredFont(forTextStyle: .body)`. Setting `UITextField.font` to nil results in a smaller `fontSize`


bencochran:
Thanks for this!

PR #130, opened 2020-08-04
Update Tutorial2.md

cbowns:
Fixed up a few arguments which have changed order.


cbowns:
Couple other changes here, one sec.

bencochran:
Thanks!

Side-note: I wonder if we should be pushing people to use `.inset(…)` and `.constrainedTo`? (Thoughts, @kyleve?)

kyleve:
Yeah, seems like a good idea for us to update the demos to use the swiftui syntax

bencochran:
Filed #131

PR #134, opened 2020-08-10
Window size environment key

watt:
This is an environment value that holds the size of the window.

Putting this one in Blueprint allows the BlueprintView to set it automatically, like it does with display scale, which is handy.

### DIFF
--- a/BlueprintUI/Sources/Environment/Keys/WindowSizeKey.swift
+++ b/BlueprintUI/Sources/Environment/Keys/WindowSizeKey.swift
@@ -1,0 +1,16 @@
+import UIKit
+
+extension Environment {
+    private enum WindowSizeKey: EnvironmentKey {
+        static var defaultValue: CGSize? {
+            nil
+        }
+    }
+
+    /// The size of the window that contains the hosting `BlueprintView`.
+    /// Defaults to `nil` if the hosting view is not in a window.
+    public var windowSize: CGSize? {
+        get { self[WindowSizeKey.self] }
+        set { self[WindowSizeKey.self] = newValue }
+    }
+}
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -238,6 +238,10 @@ public final class BlueprintView: UIView {
             environment.safeAreaInsets = safeAreaInsets
         }
 
+        if let window = window {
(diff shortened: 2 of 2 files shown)


BlueprintUI/Sources/Environment/Keys/WindowSizeKey.swift
```diff
@@ -0,0 +1,15 @@
...
+
+extension Environment {
+    private enum WindowSizeKey: EnvironmentKey {
+        static var defaultValue: CGSize {
+            UIApplication.shared.windows.first?.bounds.size ?? .zero
+        }
+    }
+
+    /// The size of the window that contains the `BlueprintView`.
+    public var windowSize: CGSize {
```
kylebshr, line 11: I wonder if this should be optional, and nil by default (and nil if the blueprint view isn't in a window)?

watt: I could see an argument for that, but it would make dealing that nil case a pain in elements that use this. Same reasoning probably applies to the other keys that are set by BlueprintView (display scale, layout direction, and safe area insets).

kylebshr:
Hmm, I get what you mean but I feel like those other values have more reasonable defaults - and `?? .zero` isn’t that hard to add. Fine either way though.

kyleve:
We could also make the defaultValue assert in the case we're not in a window; or just return the UIScreen's size (which isn't bad).

kyleve:
(Tbh, the UIScreen size is likely better, because it's deterministic – the first window could change based on what windows are being shown and hidden.)

kyleve:
(Though, maybe the _first_ window will never change order in the array and this is fine?)

kylebshr:
It just seems like, if we’re not in a window, we likely can/should bail on whatever we want to do based on the window. Guarding against nil seems better than guard against zero, and being able to do so seems better than not (because we default to a different window or UIScreen). If the user wants to do ?? .zero, that’s not very hard ¯\_(ツ)_/¯

watt:
Thanks for all the thoughts on this.

> if we’re not in a window, we likely can/should bail on whatever we want to do based on the window

I think I'm inclined to agree. None of these other options are really good defaults. Will update to default to `nil`.

watt:
> Don't forget to add to the changelog!

Gonna follow up with a changelog-only PR since I also forgot this in #135 🙈

PR #135, opened 2020-08-10
GeometryReader

### LINKED ISSUE #65: GeometryReader
It would be handy to have an element similar to SwiftUI's [GeometryReader](https://developer.apple.com/documentation/swiftui/geometryreader).

This would allow for layout-time sizing behaviors to be written ad hoc without having to build a Layout implementation.

watt:
Here's an implementation of GeometryReader, which uses the existing LazyStorage mechanism, modified to include the size constraint alongside the environment.

GeometryReader makes it easy to build ad-hoc layouts, but is more powerful than that. With the Layout protocol you can change the layout of children based on the size constraint but you cannot change their contents. This allows the content itself to vary too.

With the size constraint and the environment both available, you can also measure children and vary content based on that. There's a function built into the GeometryProxy to make this easy to do. There's a new "responsive layout" screen in the demo app that demonstrates this.

Resolves #65

### DIFF
--- a/SampleApp/Sources/RootViewController.swift
+++ b/SampleApp/Sources/RootViewController.swift
@@ -24,6 +24,9 @@ final class RootViewController : UIViewController
             DemoItem(title: "Keyboard Scrolling", onTap: { [weak self] in
                 self?.push(ScrollViewKeyboardViewController())
             }),
+            DemoItem(title: "GeometryReader Responsive Layout", onTap: { [weak self] in
+                self?.push(ResponsiveViewController())
+            }),
         ]
     }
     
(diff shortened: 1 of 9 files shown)


kyleve:
Nice

SampleApp/Sources/RootViewController.swift
```diff
@@ -24,6 +24,9 @@ final class RootViewController : UIViewController
             DemoItem(title: "Keyboard Scrolling", onTap: { [weak self] in
                 self?.push(ScrollViewKeyboardViewController())
             }),
+            DemoItem(title: "Responsive Layout", onTap: { [weak self] in
```
kylebshr, line 27: My only nit would be to mention "GeometryReader" in this title so someone looking for an example would notice it!

PR #137, opened 2020-08-12
Set a shadow path on Box

kylebshr:
Setting an explicit shadow path improves performance, since the system doesn't have to draw the contents of the view offscreen first to figure out the shadow path itself. While this prevents the shadow from being applied to the view contents if the background is transparent, this seems like reasonable default behavior for `Box` since it is, well, a "box."


kylebshr:
6e7bd30c1770ec12ce304221f8801b2b

bencochran:
Did a quick cursory glance to make sure no one in POS was relying on the old behavior, and everyone that specifies a shadow style also has a background color, so we should be safe with this

kylebshr:
Good thinking, thanks for checking @bencochran!

PR #139, opened 2020-08-25
Recursively layoutIfNeeded on backing views

watt:
It's common for backing views to have internally managed subviews, and to lay them out during `layoutSubviews`. Currently, such layout is not done during the containing BlueprintView's layout pass, but asynchronously whenever UIKit calls it. This can break expectations with things like animations.

Consider this composition. A row contains 1 or 3 red boxes. The row is wrapped in a box that clips its content.

```swift
let segment = Box(
    backgroundColor: .red,
    wrapping: Spacer(width: 50, height: 50))

let bar = Row { row in
    row.add(child: segment)
    row.minimumHorizontalSpacing = 5
    if expanded {
        row.add(child: segment)
        row.add(child: segment)
    }
}
.box(borders: .solid(color: .blue, width: 1), clipsContent: true)
.tappable { self.expanded.toggle() }

var container = TransitionContainer(wrapping: bar)
container.layoutTransition = .specific(.init(
    duration: 1.0,
    curve: .easeInOut,
    allowUserInteraction: true
))
```

`Box` is a typical backing view that lays out its content view during `layoutSubviews`. Its content view is not resized during the animation block, and the contents appear to overflow the blue border during the expansion transition, rather than being clipped.

unclipped animation

After this PR, the content view is laid out synchronously with the hosting BlueprintView, and the animation works as expected (at least as far as clipping goes).

clipped animation

This change could be a problem if a backing view does a lot of work during its layout, causing the hosting BlueprintView's layout to block the main thread for a long time.

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -278,6 +278,12 @@ extension BlueprintView {
             layoutAttributes = node.layoutAttributes
             
             viewDescription.apply(to: view)
+
+            // After this view's children are updated, allow it to run a layout pass.
+            // This ensures backing view layout changes are contained in animation blocks.
+            defer {
+                view.layoutIfNeeded()
+            }
             
             // Bail out fast if we do not have any children to manage.
             // This is a performance optimization for leaf elements, as the below update
--- a/BlueprintUI/Tests/BlueprintViewTests.swift
+++ b/BlueprintUI/Tests/BlueprintViewTests.swift
@@ -152,6 +152,50 @@ class BlueprintViewTests: XCTestCase {
         view.element = nil
         XCTAssertEqual(view.needsViewHierarchyUpdate, false)
     }
+
+    func test_recursiveLayout() {
+        let view = BlueprintView()
+
+        var layoutRecursed = false
+
+        func onLayoutSubviews() {
+            layoutRecursed = true
+        }
+
+        struct TestElement: Element {
+            var onLayoutSubviews: () -> Void
+
+            var content: ElementContent {
+                ElementContent(intrinsicSize: .zero)
+            }
+
+            func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+                TestView.describe { (config) in
+                    config.apply { (view) in
+                        // make sure UIKit knows we want a chance for layout
+                        view.setNeedsLayout()
+                    }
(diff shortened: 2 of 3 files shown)


BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -278,6 +278,12 @@ extension BlueprintView {
             layoutAttributes = node.layoutAttributes
             
             viewDescription.apply(to: view)
+
+            // After this view's children are updated, allow it to run a layout pass.
+            // This ensures backing view layout changes are contained in animation blocks.
+            defer {
```
kyleve, line 284: Nice

watt:
> Is this something we could add a behaviour test for in the BlueprintView tests?

Good idea, I'll add that.

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -152,6 +152,50 @@ class BlueprintViewTests: XCTestCase {
...
+                    super.layoutSubviews()
+                    onLayoutSubviews?()
+                }
+            }
+        }
+
+        view.element = TestElement(onLayoutSubviews: onLayoutSubviews)
+
+        // trigger a layout pass
+        _ = view.currentNativeViewControllers
```
kylebshr, line 195: Nice trick for calling `updateViewHierarchyIfNeeded` - any reason not to just make that method (and `setNeedsViewHierarchyUpdate`) internal for test cases like this?

kyleve:
Thought: Do we want to test this in SPOS in the new checkout applet? Eg, profile before and after in instruments. Since this could affect performance, we might want to verify it doesnt regress it too much since we've had problems with that applet in particular.

BlueprintUI/Tests/BlueprintViewTests.swift
kyleve: +1

watt:
> Thought: Do we want to test this in SPOS in the new checkout applet? Eg, profile before and after in instruments. Since this could affect performance, we might want to verify it doesnt regress it too much since we've had problems with that applet in particular.

Profiled against the checkout applet. There doesn't seem to be a significant impact there. I saw 5-10% of variance in the time for a render pass, but it seemed to happen regardless of whether this call was added. I also tried adding an opt-in flag on the ViewDescription, and that didn't show any difference.

kyleve:
Sweet, thank you for checking!

PR #140, opened 2020-08-25
Pass Environment to measurement-closure-based ElementContent

watt:
This change is meant to help support certain cases where leaf content needs the Environment to measure correctly, and it's inconvenient to get with an EnvironmentReader.

The primary use case is a stateful element that wraps an internal BlueprintView:

```swift
struct StatefulButton: Element {
    var makeElement: (UIControl.State) -> Element

    var content: ElementContent {
        ElementContent { (constraint) -> CGSize in
            let body = self.makeElement(.normal)
            return body.content.measure(in: constraint, environment: .empty) // <-- Environment is needed here
        }
    }

        ButtonView.describe { (config) in
            config[\.makeElement] = self.makeElement
        }
    }

    private class ButtonView: UIControl {
        let blueprintView = BlueprintView()
        var makeElement: ((UIControl.State) -> Element)?

        // snip: update() gets called when state changes

        func update() {
            blueprintView.element = makeElement?(self.state)
        }
    }
}
```

To implement with EnvironmentReader would require wrapping it in an outer element.

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -109,9 +109,10 @@ extension ElementContent {
         measurementCachingKey : MeasurementCachingKey? = nil
     ) {
         self = ElementContent(
-            layout: MeasurableLayout(measurable: measurable),
             measurementCachingKey: measurementCachingKey,
-            configure: { _ in }
+            measureFunction: { constraint, _ in
+                measurable.measure(in: constraint)
+            }
         )
     }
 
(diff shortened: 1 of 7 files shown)


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -121,21 +122,15 @@ extension ElementContent {
     /// - parameter measureFunction: How to measure the `ElementContent` in the given `SizeConstraint`.
     public init(
         measurementCachingKey : MeasurementCachingKey? = nil,
-        measureFunction: @escaping (SizeConstraint) -> CGSize
+        measureFunction: @escaping (SizeConstraint, Environment) -> CGSize
```
kyleve, line 125: Fwiw, just changing this vs adding a second method might be a bit of a PITA to adopt in SPOS (IIRC last time I did this there were a lot of these calls), but also is only a one time thing.

watt:
Going to decline this for now. It doesn't actually solve the environment bridging problem, and it seems that a solution for that will make this unnecessary.

kyleve:
Got it – maybe a "good enough" thing that we could land? Eg, worst case is an extra param chillin around; but it does seem useful for the embedded measurement case.

PR #141, opened 2020-08-28
Automatic doc generation

watt:
This is a script to generate jazzy docs at https://square.github.io/Blueprint/, and a Github workflow to automatically build and republish them after every PR.

(I'll be working on improving the actual content of the docs separately.)


.jazzy.yaml
```diff
@@ -0,0 +1,72 @@
+module: Blueprint
```
kyleve, line 1: Is this file generated, or manually maintained? (not super familiar with how jazzy works)

.gitignore
```diff
@@ -8,9 +8,11 @@ gen/
...
 # Cocoapods
-
 Pods/
 *.xcworkspace
 
 # SPM
 .swiftpm/
+
+# jazzy
+docs/
```
kylebshr, line 18: What do you think about using `.docs` if it's ignored, in case someone wants to add a `docs` folder? (I guess they could change it then, but I could see someone scratching their head for a while 😅)

.jazzy.yaml
kylebshr: Manual unfortunately

.jazzy.yaml
```diff
@@ -0,0 +1,72 @@
...
+title: Blueprint
+github_url: https://github.com/square/Blueprint
+skip_undocumented: false
+theme: fullwidth
+# TODO: to enable this, first rename docs whose names conflict with symbols
```
kylebshr, line 6: Is this a TODO for this PR or future work?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: Generate docs with jazzy and publish to Github pages
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Generate Docs
+        run: |
+          brew install sourcekitten
+          gem install jazzy
```
kyleve, line 20: Should this be added to the gemfile?

kyleve: (Maybe not, since it's only for doc generation? But would be good to stabilize the version somehow)

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+  build:
+    name: Generate docs with jazzy and publish to Github pages
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Generate Docs
+        run: |
+          brew install sourcekitten
```
kyleve, line 19: Do we want to pin to a version?

.gitignore
watt: Sure

.jazzy.yaml
watt: Future work.

.github/workflows/docs.yml
watt: Sure, that's easy enough.

.github/workflows/docs.yml
watt: Unfortunately it's not possible to pin a version with Homebrew, but I think I can adjust the script to use the sourcekitten binary embedded in jazzy and bypass this install entirely.

watt:
> We should add a note to CONTRIBUTING about adding elements to jazzy config if you add a new element. It'd be nice if we could detect elements that aren't getting docs generated. IMO the nicer structure is worth the manual config though

Made #142 to track that

PR #143, opened 2020-09-09
Add convenience methods for adding children to stacks

kylebshr:
We've mentioned upstreaming these in Market PRs - just copy-pasted @watt's implementation :)

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -74,7 +74,7 @@ extension StackElement {
     ///
     ///   - child: The child element to add to this stack
     ///
-    mutating public func add(growPriority: CGFloat = 1.0, shrinkPriority: CGFloat = 1.0, key: AnyHashable? = nil, child: Element) {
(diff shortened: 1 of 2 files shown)


kyleve (approved):

kyleve:
Don't forget the changelog!

kylebshr:
Was typing it when I got the push for that message 😄

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -82,6 +82,24 @@ extension StackElement {
         ))
     }
 
+
+    /// Convenience method for adding a child with a grow and shrink priority of 0.0
+    ///
+    /// - parameters:
+    ///   - child: The child element to add to this stack
+    ///
+    mutating func addFixed(child: Element) {
```
bencochran, line 91: These should be `public`

watt (approved):

BlueprintUI/Sources/Layout/Stack.swift
kylebshr: doi! nice catch

PR #144, opened 2020-09-12
Add accessibility size to AccessibilityElement

kylebshr:
Similar to how you might want to set a `minimumTappableSize` on a button, you might want to customize the size of the frame that voice over draws around your element. This PR allows you to set a size for the frame. Open to other options here, such as a closure property that gets called with the view frame, but that seemed needlessly complex.

I used this in Market to customize the frame of a `Button` that was 16x16 since it wrapped a small image (the tap target is the default 44x44):

Screen Shot 2020-09-11 at 10 52 55 PM

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -28,6 +28,7 @@ public struct AccessibilityElement: Element {
     public var hint: String?
     public var identifier: String?
     public var traits: Set<Trait>
+    public var accessibilityFrameSize: CGSize?
     public var wrappedElement: Element
 
     public init(
@@ -36,6 +37,7 @@ public struct AccessibilityElement: Element {
         hint: String? = nil,
         identifier: String? = nil,
         traits: Set<Trait> = [],
+        accessibilityFrameSize: CGSize? = nil,
         wrapping element: Element
     ) {
         self.label = label
@@ -43,6 +45,7 @@ public struct AccessibilityElement: Element {
         self.hint = hint
         self.identifier = identifier
         self.traits = traits
+        self.accessibilityFrameSize = accessibilityFrameSize
         self.wrappedElement = element
     }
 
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -96,36 +99,61 @@ public struct AccessibilityElement: Element {
...
             config[\.accessibilityValue] = value
             config[\.accessibilityHint] = hint
             config[\.accessibilityIdentifier] = identifier
             config[\.accessibilityTraits] = accessibilityTraits
             config[\.isAccessibilityElement] = true
+            config[\.accessibilityFrameSize] = accessibilityFrameSize
+        }
+    }
+
+    private class AccessibilityView: UIView {
```
kyleve, line 113: Nit: Final

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -28,6 +28,7 @@ public struct AccessibilityElement: Element {
     public var hint: String?
     public var identifier: String?
     public var traits: Set<Trait>
+    public var accessibilityFrameSize: CGSize?
```
kyleve, line 31: Maybe instead of an explicit size, a minimum size would be better? No strong opinions either way, but that feels less fragile than a fixed size; eg if the wrapped view changes due to a design change.

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kylebshr: I actually started with that but thought that wouldn't let you make a smaller rect in some cases.. might be worth adding an enum for different options in the future? But probably will just add this for now

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kylebshr, line 113:
```suggestion
    private final class AccessibilityView: UIView {
```

watt (approved):

PR #145, opened 2020-09-12
[Box] Add support for boxes with semicircular edges

nononoah:
### Summary

This PR adds the `capsule` case to `Box.CornerStyle`. This case sugars the following pattern

```
GeometryReader { geometry in
  Box(cornerStyle: .rounded(geometry.constraint.height.maximum / 2.0))
}
```

into

```
Box(cornerStyle: .capsule)
```

Given the effect of `GeometryReader's` closure on rightward drift and capture semantics and the relative ubiquity of this design element, I believe that this sugar is merited. I would further argue that this change comes at a small cost: a slight expansion of API surface area, and almost no implementation disruption.

### Changes

- updates `var radius: CGFloat` on Box.CornerStyle to `func radius(for:) -> CGFloat`
- extends and organizes relevant snapshot tests

### Additional comments

This is my first PR, so I'm likely to have made a few process mistakes. Please correct me where I've gone wrong.


BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -102,6 +102,7 @@ extension Box {
 
     public enum CornerStyle {
         case square
+        case semicircular
```
kyleve, line 105: I wonder if “pill” might be clearer? No preference, but in my head that makes me more immediately realize what it does.

kyleve (approved):
Looks good! Can you also update the Changelog.md in the root of the repo with this change?

BlueprintUICommonControls/Sources/Box.swift
nononoah: I'm open to changing it of course, but biased towards `semicircular` because

- it fits grammatically with the other cases (as an adjective)
- it is a  generally understood geometric term (like square)

In my mind, `pill` is jargon and no different than `chip`, what Google's Material calls it. If you're familiar with the argot, it's familiar. Since most people working with Blueprint share this familiarity, I think it's a fine choice.

nononoah:
Pending additional review, this is ready to be merged. 

If someone with write access would like to bring it home, I'd appreciate that.

kylebshr (approved):

BlueprintUICommonControls/Sources/Box.swift
bencochran: Way overthought bikeshed:

I think part of the issue here is that the existing cases are purely the corner style where this new case is wanting to describe a shape as a whole (that is, affecting more than each corner individually). And I think we're left trying avoid using a “shape” word to describe a “corner.”

And my mind _still_ reads it as a shape and expects a half a circle that's flat on the other side (that is to say, a literal semicircle).

FWIW SwiftUI avoids this issue by having this as a shape type (called [Capsule](https://developer.apple.com/documentation/swiftui/capsule)) rather than a corner style.

I could see a few options:

1. Rename this type/property to be “Shape” and then something like “Capsule” or “Pill” seems more right to me
2. Put this into its own element type altogether (again, as Pill or Capsule IMO)
3. Land as-is with the same or a different name (in which case I'd probably still lean against “semicircle” and just accept conflating shape and corner a bit)

(And Google's wrong, “chip” is a weird name)

kyleve: capsule is also a good name for this property 

I wouldn't make this a different element since we'd have to duplicate all the shadow and border style stuff.

nononoah: Given the enthusiasm for `capsule `, I updated it. Still, I think it stands in contrast to its sibling cases.

> And my mind still reads it as a shape and expects a half a circle that's flat on the other side (that is to say, a literal semicircle).

That's fair, though this is one impression I tried to avoid when renaming it from `semicircle` to `semicircular`. The latter may be ascribed to an object forming a semicircle, which is a fair attribution to such corners. This is similar if not identical semantically to `square`, I'd argue.

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -140,10 +141,12 @@ public extension Element {
 
 extension Box.CornerStyle {
 
-    fileprivate var radius: CGFloat {
+    fileprivate func radius(for bounds: CGRect) -> CGFloat {
         switch self {
         case .square:
             return 0
+        case .semicircular:
+            return bounds.height / 2.0
```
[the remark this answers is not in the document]
nononoah: Good call. Updated and added a screenshot.

kylebshr (approved):

BlueprintUICommonControls/Tests/Sources/BoxTests.swift
```diff
@@ -18,13 +18,31 @@ class BoxTests: XCTestCase {
             identifier: "clear")
     }
 
-    func test_cornerRadius() {
```
bencochran, line 21: Could you delete the old image please?

BlueprintUICommonControls/Sources/Box.swift
bencochran: Yeah, we've definitely got a less than idea parallelism here. I think were this has landed is more intuitive, though less … precise.

bencochran (approved):

BlueprintUICommonControls/Tests/Sources/BoxTests.swift
nononoah: I think you mean this file, which has been deleted, though it's a bit tricky to tell from Github.

Screen Shot 2020-09-13 at 11 55 12 PM

Please let me know if you're referring to another one.

bencochran: Oh oops, yeah sorry. GitHub on my phone really didn't want me to see it. Thanks!

PR #146, opened 2020-09-13
[Gardening] Ignore artifacts of `swift package generate-xcodeproj`

nononoah:
### Summary

This is a purely gardening PR which extends the `gitignore` to cover artifacts generated by `swift package generate-xcodeproj`.


bencochran (approved):
Thanks! (Also in case you didn't know, Xcode can now open SPM projects directly so to work on Blueprint I typically do `xed .` at the repo root rather than generating a project)

nononoah:
> Thanks! (Also in case you didn't know, Xcode can now open SPM projects directly so to work on Blueprint I typically do xed . at the repo root rather than generating a project)

Actually, I wasn't aware of this, so thanks for sharing. 

Can you produce builds required for unit testing with this approach?

PR #147, opened 2020-09-14
Add Opacity wrapper

bencochran:


BlueprintUI/Sources/Layout/Opacity.swift
```diff
@@ -0,0 +1,60 @@
...
+    /// The opacity of the wrapped element.
+    public var opacity: CGFloat
+
+    /// Initializes an `Opacity` with the given content element and opacity.
+    ///
+    /// - parameters:
+    ///   - opacity: The opacity to be applied to the wrapped element.
+    ///   - wrapping: The content element to be made transparent.
+    public init(
+        opacity: CGFloat,
```
kyleve, line 18: Was going to ask if we need to `allowsGroupOpacity`, but apparently I am 7 years out of date and this is now the default! https://developer.apple.com/documentation/quartzcore/calayer/1621277-allowsgroupopacity

kylebshr:
Thanks for putting this up! What was your thinking behind not putting it in `CommonControls`? It's not exactly a control, so I could see it living in `BlueprintUI`, just wondering if it was intentional. @watt do you have an opinion on where this should live?

kylebshr (approved):

kyleve (approved):

BlueprintUI/Sources/Layout/Opacity.swift
kylebshr: I've literally never heard of that property 🙈

bencochran:
> What was your thinking behind not putting it in `CommonControls`?

Yeah, tbh I went back and forth on this. Technically since alpha is part of LayoutAttributes, this is a custom layout like `Inset`, etc. but that’s almost an implementation detail more than anything. I believe part of the distinction between BlueprintUI and BlueprintUICommonControls is a holdout from when Blueprint could theoretically support macOS (where things requiring UIKit directly would push toward CommonControls). But that line gets blurrier and blurrier every day, so I don’t have a strong opinion here.

kylebshr:
Sounds good, I think keeping it in BP makes sense

watt:
> @watt do you have an opinion on where this should live?

Not really! My usual litmus test is whether it's view-backed, but to echo what Ben said, this one kind of lives in a gray area.

watt (approved):

PR #152, opened 2020-09-17
Run all tests on iOS 10.3.1-14.0

kylebshr:
- Updates snapshot tests to include major iOS version in file name
- Generate snapshot images for iOS 10, 11, 12, 13 and 14
- Specify a matrix of destinations so tests run on every version we support
- Update tests to run using Xcode 12 (so we can test on iOS 14)
- Updates `test_measuring` with disparate behavior on iOS 10, as measuring text is different on iOS 10
- Bumps swift version to 5.2 and uses the `.v10` constant for the platform version

Even though this quintuples the snapshot images in the repo, IMO it's worth it as most of the images are 1-20kb, though some are ~0.5mb


BlueprintUICommonControls/Tests/Sources/XCTestCaseExtensions.swift
```diff
@@ -30,7 +30,8 @@ extension XCTestCase {
 
 
         let testName = testName.prefix(testName.count - 2)
-        imageURL.appendPathComponent("\(testName)_\(identifier ?? "").png")
+        let majorVersion = UIDevice.current.systemVersion.components(separatedBy: ".")[0]
```
kyleve, line 33: `ProcessInfo.processInfo.operatingSystemVersion` gives you a `majorVersion` property to avoid the string parsing here.

kyleve (approved):

BlueprintUICommonControls/Tests/Sources/XCTestCaseExtensions.swift
kylebshr: Nice, will update!

watt (approved):

bencochran (approved):

kylebshr:
Ah it's failing because there's new snapshots.. gotta update those

Issue #17, opened 2019-09-23
Wrapped elements should be the last parameter


watt:
Many Blueprint elements that wrap a single child have an initializer defined like this:

`WrapperElement(wrapping: Element, param1: X, param2: Y)`

Where the wrapped element is the first parameter. Unfortunately, this ordering does not lend itself to inline-nested Element compositions, because it pushes the domain-related parameters away from the type they're associated with.

Consider the following element construction:

```swift
var constrainedInsetRow: Element {
    return ConstrainedSize(
        wrapping: Inset(
            wrapping: Row { row in
                row.verticalAlignment = .center

                for leadingElement in self.leadingElements {
                    row.add(growPriority: 0, shrinkPriority: 0, child: leadingElement)
                }

                if let contentElement = self.contentElement {
                    row.add(child: contentElement)
                }

                for trailingElement in self.trailingElements {
                    row.add(growPriority: 0, shrinkPriority: 0, child: trailingElement)
                }
            },
            uniformInset: 8
        ),
        height: .atLeast(44)
    )
}
```

The `uniformInset` parameter is pushed 16 lines away from the `Inset` it is associated with, and `height` is 19 lines away from `ConstrainedSize`.

If we move the `wrapping` parameters to the end of the argument list, it looks like this:

```swift
var constrainedInsetRow: Element {
    return ConstrainedSize(
        height: .atLeast(44),
        wrapping: Inset(
            uniformInset: 8,
            wrapping: Row { row in
                row.verticalAlignment = .center

                for leadingElement in self.leadingElements {
                    row.add(growPriority: 0, shrinkPriority: 0, child: leadingElement)
                }

                if let contentElement = self.contentElement {
                    row.add(child: contentElement)
                }

                for trailingElement in self.trailingElements {
                    row.add(growPriority: 0, shrinkPriority: 0, child: trailingElement)
                }
            }
        )
    )
}
```

Now it is immediately clear that `height` goes with `ConstrainedSize` and `uniformInset` goes with `Inset`, without really affecting the clarity of the nesting either.

Can we change the order of the arguments on `ConstrainedSize`, `Inset`, etc. to consistently have the wrapped element be last?

PR #153, opened 2020-09-17
Alignment guides

watt:
This PR adds alignment guides to stacks, allowing you to fine-tune cross axis alignment. The API is inspired by SwiftUI's alignment guides.

Alignment guides are stored on the `Traits` type and are specified with a new closure parameter on the `add()` function, which takes `ElementDimensions` and returns the value of the guide in the element's own coordinate space.

```swift
Row { row in
    row.verticalAlignment = .bottom

    row.add(
        alignmentGuide: { (d: ElementDimensions) -> CGFloat in
            // Add a guide 10 points from bottom.
            d.height - 10
        },
        child: MyElement()
    )

    row.add(
        alignmentGuide: { (d: ElementDimensions) -> CGFloat in
            // Align to the center by referencing another guide.
            d[VerticalAlignment.center]
        },
        child: MyElement()
    )
    
    // Use the default alignment
    row.add(child: MyElement())
}
```

`ElementDimensions` is analogous to SwiftUI's `ViewDimensions`: you can use it to read the size of the element or get the value of another alignment.

Stack children will be relatively aligned to each other, and then the entire contents will be aligned to the bounds of the stack.

Alignment can push content outside the bounds of the stack. Stack measurement takes into account the distance from the "lowest" edge to the "highest" edge of all children, so a stack's measured size is big enough to accommodate its aligned content. Basically, it does the thing you'd expect.

The standard alignments have been converted into the new `HorizontalAlignment` and `VerticalAlignment` types, which work just like the SwiftUI types. You can define a new alignment, but it's not necessary for alignment guides.

Stacks wrap these alignments in another enum type in order to support `fill` mode, which works unlike any alignment.

`Row`'s `.leading` and `.trailing` alignments are now aliases for `.top` and `.bottom` alignments, and have been marked as deprecated.

## Differences from SwiftUI

* Alignment does not propagate — it is entirely internal to a stack.
* Because it does not propagate there is no need to specify which alignment a guide applies to. It always applies to the one being used by the stack.
* SwiftUI stacks are always the size of their content, but Blueprint stacks can be any size. This is the reason for a two-phase alignment, which aligns relatively first and then aligns the content to the stack bounds.
* There are no built-in baseline alignments, because there is no way to propagate such information from an element containing text up to the stack. You can do baseline alignment, but you must specify the values yourself.

### DIFF
--- a/BlueprintUI/Sources/Layout/Alignment.swift
+++ b/BlueprintUI/Sources/Layout/Alignment.swift
@@ -1,0 +1,106 @@
+import CoreGraphics
+
+/// Types used to identify alignment guides.
+///
+/// Types conforming to `AlignmentID` have a corresponding alignment guide value,
+/// typically declared as a static constant property of `HorizontalAlignment` or 
+/// `VerticalAlignment`.
+///
+public protocol AlignmentID {
+
+    /// Returns the value of the corresponding guide, in `context`, when not
+    /// otherwise set in `context`.
+    static func defaultValue(in context: ElementDimensions) -> CGFloat
+}
+
+/// An alignment position along the horizontal axis.
+public struct HorizontalAlignment: Equatable {
+
+    var id: AlignmentID.Type
+
+    /// Creates an instance with the given ID.
+    ///
+    /// Note: each instance should have a unique ID.
+    public init(_ id: AlignmentID.Type) {
+        self.id = id
+    }
+
+    public static func == (lhs: HorizontalAlignment, rhs: HorizontalAlignment) -> Bool {
+        return lhs.id == rhs.id
+    }
+}
+
+/// An alignment position along the vertical axis.
+public struct VerticalAlignment: Equatable {
+
+    var id: AlignmentID.Type
+
+    /// Creates an instance with the given ID.
+    ///
+    /// Note: each instance should have a unique ID.
+    public init(_ id: AlignmentID.Type) {
+        self.id = id
+    }
+
+    public static func == (lhs: VerticalAlignment, rhs: VerticalAlignment) -> Bool {
+        return lhs.id == rhs.id
+    }
+}
+
+extension HorizontalAlignment {
+    enum Leading: AlignmentID {
+        static func defaultValue(in context: ElementDimensions) -> CGFloat {
+            0
+        }
+    }
+
+    /// A guide marking the leading edge of the element.
+    public static let leading = HorizontalAlignment(Leading.self)
+
+    enum Center: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
+            d.width / 2
+        }
+    }
+
+    /// A guide marking the horizontal center of the element.
+    public static let center = HorizontalAlignment(Center.self)
+
+    enum Trailing: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
+            d.width
+        }
+    }
+
+    /// A guide marking the trailing edge of the element.
+    public static let trailing = HorizontalAlignment(Trailing.self)
+}
+
+extension VerticalAlignment {
+    enum Top: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
+            0
+        }
+    }
+
+    /// A guide marking the top edge of the element.
+    public static let top = VerticalAlignment(Top.self)
+
+    enum Center: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
+            d.height / 2
+        }
+    }
+
+    /// A guide marking the vertical center of the element.
+    public static let center = VerticalAlignment(Center.self)
+
+    enum Bottom: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
+            d.height
+        }
+    }
+
+    /// A guide marking the bottom edge of the element.
+    public static let bottom = VerticalAlignment(Bottom.self)
+}
--- a/BlueprintUI/Sources/Layout/Column.swift
+++ b/BlueprintUI/Sources/Layout/Column.swift
@@ -2,10 +2,51 @@ import UIKit
 
 /// Displays a list of items in a linear vertical layout.
 public struct Column: StackElement {
+    /// Describes how the column's children will be horizontally aligned.
+    public enum ColumnAlignment: Equatable {
+        /// Children will be stretched to fit the horizontal size of the column.
+        case fill
+
+        /// Using the specified alignment, children will be aligned relatively to each other, and
+        /// then all the contents will be aligned to the column's bounding box.
+        ///
+        /// This case can be used for custom alignments. For common alignments you can use the
+        /// existing static instances`leading`, `center`, and `trailing`.
+        ///
+        case align(to: HorizontalAlignment)
+
+        /// Children will be aligned to the leading edge of the column.
+        public static let leading = ColumnAlignment.align(to: .leading)
+        /// Children will be horizontally centered in the column.
+        public static let center = ColumnAlignment.align(to: .center)
+        /// Children will be aligned to the trailing edge of the column.
+        public static let trailing = ColumnAlignment.align(to: .trailing)
+
+        init(_ stackAlignment: StackLayout.Alignment) {
+            switch stackAlignment {
+            case .fill:
+                self = .fill
+            case let .align(to: id):
+                self = .align(to: HorizontalAlignment(id))
+            }
+        }
+
+        var stackAlignment: StackLayout.Alignment {
+            switch self {
+            case .fill:
+                return .fill
+            case let .align(to: alignment):
+                return .align(to: alignment.id)
+            }
+        }
+    }
 
     public var children: [(element: Element, traits: StackLayout.Traits, key: AnyHashable?)] = []
 
-    private (set) public var layout = StackLayout(axis: .vertical)
+    private (set) public var layout = StackLayout(
+        axis: .vertical,
+        alignment: ColumnAlignment.leading.stackAlignment
+    )
 
     public init() {}
 
@@ -19,9 +60,14 @@ public struct Column: StackElement {
         set { layout.overflow = newValue }
     }
 
-    public var horizontalAlignment: StackLayout.Alignment {
-        get { return layout.alignment }
-        set { layout.alignment = newValue }
+    /// Specifies how children will be aligned horizontally.
+    public var horizontalAlignment: ColumnAlignment {
+        get {
+            ColumnAlignment(layout.alignment)
+        }
+        set {
+            layout.alignment = newValue.stackAlignment
+        }
     }
 
     public var minimumVerticalSpacing: CGFloat {
--- a/BlueprintUI/Sources/Layout/ElementDimensions.swift
+++ b/BlueprintUI/Sources/Layout/ElementDimensions.swift
@@ -1,0 +1,75 @@
+import CoreGraphics
+
+/// An element’s size and its alignment guides in its own coordinate space.
+///
+/// You can access the size of the element through the `width` and `height` properties. You can
+/// access alignment guide values by subscripting with the specific alignment.
+///
+/// These dimensions are typically used when setting an alignment guide on a stack, with
+/// `StackElement.add(...)`.
+///
+/// ## Example
+///
+/// ```
+/// // get the alignment guide value for `VerticalAlignment.center`, falling back to the default
+/// // value if no alignment guide has been set
+/// dimensions[VerticalAlignment.center]
+///
+/// // get the alignment guide value for `HorizontalAlignment.trailing`, or `nil` if none has been
+/// // set.
+/// dimensions[explicit: .trailing]
+/// ```
+///
+/// ## See Also
+/// [StackElement.add()](x-source-tag://StackElement.add)
+///
+public struct ElementDimensions: Equatable {
+
+    /// The element's width
+    public internal(set) var width: CGFloat
+
+    /// The element's height
+    public internal(set) var height: CGFloat
+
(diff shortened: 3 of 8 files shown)


BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -2,10 +2,46 @@ import UIKit
 
 /// Displays a list of items in a linear vertical layout.
 public struct Column: StackElement {
+    /// Describes how the column's children will be horizontally aligned.
+    public enum ColumnAlignment: Equatable {
+        /// Children will be stretched to fit the horizontal size of the column.
+        case fill
+        /// Using the specified alignment, children will be aligned relatively to each other, and
+        /// then all the contents will be aligned to the column's bounding box.
```
kylebshr, line 10: I would mention the fact that you can using `leading` etc. without having to use this directly

BlueprintUI/Sources/Layout/Alignment.swift
```diff
@@ -0,0 +1,106 @@
+import CoreGraphics
+
+/// Types used to identify alignment guides.
+///
+/// Types conforming to `AlignmentID` have a corresponding alignment guide value,
+/// typically declared as a static constant property of `HorizontalAlignment` or 
+/// `VerticalAlignment`.
+///
+public protocol AlignmentID {
```
kylebshr, line 9: Does this need to be public? While I think I've wrapped my head around the impl. and why it's needed, I wonder if it's best kept as an implementation detail. Or do you think it's useful to let folks create custom H/V alignments with a custom ID?

kylebshr (commented):
I think this generally makes sense, but I need to take a look at how it's used in Market to make sure I follow why everything is needed. 

Not sure I totally follow what the alignment guides are offering, vs. just adding the closure that provides an `ElementDimensions` and returns an offset for the element.

BlueprintUI/Sources/Layout/ElementDimensions.swift
```diff
@@ -0,0 +1,51 @@
...
+    private var horizontalGuideValues: [ObjectIdentifier: CGFloat] = [:]
+    private var verticalGuideValues: [ObjectIdentifier: CGFloat] = [:]
+
+    init(size: CGSize) {
+        self.width = size.width
+        self.height = size.height
+    }
+
+    /// Accesses the value of the given guide.
+    public internal(set) subscript(guide: HorizontalAlignment) -> CGFloat {
```
kylebshr, line 21: Could just be me but not the biggest fan of keypaths here because I always find this less discoverable. At the very least, please document that you need to use keypaths to get at the guide values for this type :)

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -70,14 +70,40 @@ extension StackElement {
...
     ///
     ///   - child: The child element to add to this stack
     ///
-    public mutating func add(growPriority: CGFloat = 1.0, shrinkPriority: CGFloat = 1.0, key: AnyHashable? = nil, child: Element) {
+    /// - Tag: StackElement.add
+    ///
+    mutating public func add(
+        growPriority: CGFloat = 1.0,
+        shrinkPriority: CGFloat = 1.0,
+        alignmentGuide: ((ElementDimensions) -> CGFloat)? = nil,
```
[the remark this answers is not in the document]
kyleve: Because you usually build the element inline, so it looked weird having little "parameter turds" hanging off the end like this:

```
stack.add(
child: MyElement(
   thatWrap: AnotherThing(
     with: more,
     params: likeThis
   ) { 
     $0.configuring...
  }
  growPriority: 1,
  shrinkPriority: 1,
)
```

BlueprintUI/Sources/Layout/Alignment.swift
kyleve: > Or do you think it's useful to let folks create custom H/V alignments with a custom ID?

I believe this is the goal and aligns w/ how SwiftUI works.

BlueprintUI/Sources/Layout/Stack.swift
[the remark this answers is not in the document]
kyleve: https://github.com/square/Blueprint/issues/17

BlueprintUI/Sources/Layout/Alignment.swift
watt: It's not very useful today, because these do not propagate, but as Kyle said, this _aligns_ us with SwiftUI. 😏

BlueprintUI/Sources/Layout/Column.swift
watt: Sure, I'll update this doc to make it clearer that it's only needed for custom alignments.

BlueprintUI/Sources/Layout/ElementDimensions.swift
watt: Do you mean subscripts, not keypaths? Or are you talking about the static instances on `HorizontalAlignment` and `VerticalAlignment`?

kylebshr: Monday 😅 I did mean subscripts. But thinking about how a func would be shaped, not sure it's better - happy keeping it as a subscript

kylebshr (approved):

BlueprintUI/Sources/Layout/ElementDimensions.swift
watt: Gotcha. This API is meant to mirror [ViewDimensions](https://developer.apple.com/documentation/swiftui/viewdimensions), but I will do a documentation pass.

BlueprintUI/Sources/Layout/Stack.swift
bencochran, line 96: Sorry, really late in the review with this, but why a closure here and not a `HorizontalAlignment` or `VerticalAlignment`? Would make us split out this method between the Column / Row, but feels like it makes using alignment in the common case much more straightforward/discoverable.

watt: Then you'd have to declare a new alignment ID type for each element you wanted to give a unique guide value, which could be cumbersome if you are aligning several elements.

I probably should have been clearer in the description about the 2 distinct features in this PR:
* Custom alignments, made with `HorizontalAlignment` and `VerticalAlignment` instances, are useful if you have an alignment that can be generalized across many cases (e.g. `top` is going to be `0` on all elements).
* Alignment guides are generally unique to the specific items you are aligning, like choosing the visual center on a particular image.

This API mimicks the [alignmentGuide modifier](https://developer.apple.com/documentation/swiftui/view/alignmentguide(_:computevalue:)-330fz) in SwiftUI, which lets you attach a closure to specify the guide value on individual elements.

kyleve:
Finally getting to this today; but since @kylebshr and @bencochran have already looked, don't block on my review.

BlueprintUI/Sources/Layout/Alignment.swift
```diff
@@ -0,0 +1,106 @@
...
+    var id: AlignmentID.Type
+
+    /// Creates an instance with the given ID.
+    ///
+    /// Note: each instance should have a unique ID.
+    public init(_ id: AlignmentID.Type) {
+        self.id = id
+    }
+
+    public static func == (lhs: HorizontalAlignment, rhs: HorizontalAlignment) -> Bool {
```
kyleve, line 28: Could this be synthesized automatically if the `id` parameter was an `ObjectIdentifier`, which was automatically synthesized to return `ObjectIdentifier(type(of: self))`?

kyleve: Er, no coffee yet, I misspoke on this one

If the ID was an `ObjectIdentifier` that was set in init from `AlignmentID.Type`.

Not a big deal since it's only one property, though.

BlueprintUI/Sources/Layout/Alignment.swift
```diff
@@ -0,0 +1,106 @@
...
+        static func defaultValue(in context: ElementDimensions) -> CGFloat {
+            0
+        }
+    }
+
+    /// A guide marking the leading edge of the element.
+    public static let leading = HorizontalAlignment(Leading.self)
+
+    enum Center: AlignmentID {
+        static func defaultValue(in d: ElementDimensions) -> CGFloat {
```
kyleve, line 61: Thought: Is there a more descriptive name for this than `value`? Seeing `d.width / 2` here, I'm not sure what value the halved width is getting applied to. The centerpoint, the min axis, the max axis, etc.

kyleve: (I guess it's really dependent on the context of the alignment, so maybe no better name?)

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -36,69 +36,138 @@ extension StackElement {
...
+        key: AnyHashable? = nil,
+        child: Element
+    ) {
         children.append((
             element: child,
-            traits: StackLayout.Traits(growPriority: growPriority, shrinkPriority: shrinkPriority),
+            traits: StackLayout.Traits(
+                growPriority: growPriority,
+                shrinkPriority: shrinkPriority,
+                alignmentGuide: alignmentGuide.map(StackLayout.AlignmentGuide.init)
```
kyleve, line 110: TIL, forgot you could do this

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -560,50 +673,97 @@ extension StackLayout {
...
+                repeating: Segment(origin: 0, magnitude: availableCross),
+                count: items.count
+            )
+
+            return segments
+        }
+
+        func alignSegments(to alignment: AlignmentID.Type) -> [Segment] {
+            // Get the alignment values for each child
+            let alignmentValues = items.indices.map { i -> CGFloat in
```
kyleve, line 705: Minor: Could also be a `zip`

kyleve: (Though I guess if there's three accesses in here; meh?)

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -560,50 +673,97 @@ extension StackLayout {
...
+                let value = alignmentGuide?.computeValue(dimensions) ?? alignment.defaultValue(in: dimensions)
+
+                return value
+            }
+
+            // Find the relative distance from the "lowest" edge to the "highest" edge.
+            // This may be greater than the size of any single child, if alignment values push
+            // children outside of their normal bounds.
+
+            let minAlignedCross = -alignmentValues.max()!
```
kyleve, line 722: Any reason to be concerned about performance in these max calls for many elements?

kyleve: (Eg, is this called once per layout, or N times, one for each item?)

BlueprintUI/Tests/StackTests.swift
```diff
@@ -67,65 +67,285 @@ class StackTests: XCTestCase {
...
+            expectedOrigins: [
+                100,
+                0,
+                100,
+            ]
+        )
 
+        // In this test, 4 boxes are aligned relatively based on an arbitrary offset.
+        // The relatively aligned content is larger than the stack bounds,
+        // so the alignment is used to anchor the content to the center.
```
kyleve, line 231: I find it a bit odd that if the content is larger than the bounds; everything ends up being centered; eg it's weird one outlier element can cause others' positions to be changed. Does this match SwiftUI, or is it for another reason?

kyleve (approved):

BlueprintUI/Sources/Layout/Alignment.swift
watt: I think the only reason that won't work is that we need the metatype to bridge back & forth to `StackLayout.Alignment`. I experimented with making stacks generic around the alignment type but it turned into a much bigger change.

BlueprintUI/Sources/Layout/Alignment.swift
watt: Copying the SwiftUI API here. I'm not sure what would be a better name though, since as you say, this is just an alignment value, and its use depends on the context.

BlueprintUI/Sources/Layout/Stack.swift
watt: Yep, I had to give up the zip here since we're indexing 3 sequences.

BlueprintUI/Sources/Layout/Stack.swift
watt: Shouldn't be a problem. `max` is O(n), so it's the same iterating over the children again, which we do several times during a stack layout. `alignSegments` is only called once per layout.

BlueprintUI/Tests/StackTests.swift
watt: > if the content is larger than the bounds; everything ends up being centered 

This is where the different layout models make things more complicated. In SwiftUI, the size of the stack is the size of its contents, so there is no concept of aligning the contents within the bounds — they are relatively aligned and then it's done. But in Blueprint, the size of the stack can be anything, so we still have to figure out where to place the contents after relatively aligning them.

So the stack's alignment serves two purposes here:
- first it's used during relative alignment to get the guide value on each child that doesn't have an explicit alignment guide
- then it's used to align the contents within the stack bounds

In this case the contents are centered because the row's `verticalAlignment` is `center`.

I also experimented with an API that allowed you to specify separate alignments for each of these tasks, but it just came out over-engineered and harder to understand.

> it's weird one outlier element can cause others' positions to be changed

This part does match SwiftUI, and it's key to making relative alignments work with values from the elements' own coordinate spaces. If we pushed outliers outside the stack bounds, we wouldn't be able to do things like baseline alignment, because everything would be getting pushed out.

kyleve: Got it, that makes sense.

PR #154, opened 2020-09-18
Remove extra call to NativeViewController.update

watt:
During the view update part of a render pass, when a _new_ node is being inserted, `update(node:appearanceTransitionsEnabled:)` is called twice: first in the init of `NativeViewController`, and then a bit later after the new view has been inserted into the view hierarchy.

That's a bunch of redundant work being done for new nodes, so this PR removes one of the calls.

I chose to remove the first call, because the second one matches the time that `update` is called for existing nodes (after applying layout attributes and inserting the view).

I did some quick & dirty profiling on the POS checkout applet and found it cut the time of view hierarchy update on the first render pass down by 30-50%. No difference in functionality as far as I can tell.

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -263,7 +263,6 @@ extension BlueprintView {
             self.layoutAttributes = node.layoutAttributes
             self.children = []
             self.view = node.viewDescription.build()
-            update(node: node, appearanceTransitionsEnabled: false)
         }
 
         fileprivate func canUpdateFrom(node: NativeViewNode) -> Bool {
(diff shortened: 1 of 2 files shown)


watt:
Here are the two calls, starting from [line 348 in BlueprintView.swift](https://github.com/square/Blueprint/blob/651146b58d31006d5a017bb471eed062cced1564/BlueprintUI/Sources/Blueprint%20View/BlueprintView.swift#L348):

```swift
// first call in this init
let controller = NativeViewController(node: child)
newChildren.append((path: path, node: controller))

}

// second call
controller.update(node: child, appearanceTransitionsEnabled: false)
```

kyleve:
Oooh yeah! I ran into this with the old transactions v2 PR; always found it weird this was called in init.

kyleve (approved):
Don't forget the changelog!

PR #155, opened 2020-09-19
[TransitionContainer] Sugar transition composition and extend initializer

nononoah:
### Summary

This PR adds methods for composing an element transition. In effect, it sugars the following pattern

```swift
let element = MyElement()
    .composed()
var transitionContainer = TransitionContainer(wrapping: element)
transitionContainer.layoutTransition = .none
return transitionContainer
```

into

```swift
return MyElement()
  .composed()
  .transition(appearAndDisappear: .fade) 
```

### Discussion

- I'll defer to more experienced maintainers on matters of grammar and naming. I did my best to maintain consistency with analogous sugar, but may have fallen short.
- The addition of `VisibilityTransition.none` may be regarded by some as a smell. I'm open to alternatives, such as using an optional at these sites, which would be a more disruptive but perhaps more correct change.
- No tests exist for `TransitionContainer`. As such tests would be very thin (i.e. do we map initializer parameters to properties properly), this may be fine. If not, I can add a test suite and tests for this new functionality.

### Changes
- shifts `TransitionContainer ` default property values to default initializer values, thereby extending the initializer for greater customization
- adds `transition(appear:disappear:layout)` and `transition(appearAndDisappear:)` methods
- adds `VisibilityTransition.none`
- updates CHANGELOG

### DIFF
--- a/BlueprintUICommonControls/Sources/TransitionContainer.swift
+++ b/BlueprintUICommonControls/Sources/TransitionContainer.swift
@@ -6,14 +6,22 @@ import UIKit
 /// disappears, or changes layout.
 public struct TransitionContainer: Element {
 
-    public var appearingTransition: VisibilityTransition = .fade
-    public var disappearingTransition: VisibilityTransition = .fade
-    public var layoutTransition: LayoutTransition = .specific(AnimationAttributes())
+    public var appearingTransition: VisibilityTransition
+    public var disappearingTransition: VisibilityTransition
+    public var layoutTransition: LayoutTransition
 
     public var wrappedElement: Element
 
-    public init(wrapping element: Element) {
+    public init(
+        wrapping element: Element,
+        appearingTransition: VisibilityTransition = .fade,
+        disappearingTransition: VisibilityTransition = .fade,
+        layoutTransition: LayoutTransition = .specific(AnimationAttributes())
+    ) {
         self.wrappedElement = element
+        self.appearingTransition = appearingTransition
+        self.disappearingTransition = disappearingTransition
+        self.layoutTransition = layoutTransition
     }
 
     public var content: ElementContent {
@@ -28,4 +36,39 @@ public struct TransitionContainer: Element {
         }
     }
 
+}
+
+public extension Element {
+
+    /// Wraps the element in a transition container to provide an animated transition.
+    ///
+    /// - Parameters:
+    ///   - appear: The transition to use when the element appears. By default, `.none`.
+    ///   - disappear: The transition to use when the element disappears. By default, `.none`.
+    ///   - layout: The animation to use when the element changes layout. By default, `.none`.
+    func transition(
+        appear: VisibilityTransition = .none,
+        disappear: VisibilityTransition = .none,
+        layout: LayoutTransition = .none
+    ) -> TransitionContainer {
+        TransitionContainer(
+            wrapping: self,
+            appearingTransition: appear,
+            disappearingTransition: disappear,
+            layoutTransition: layout
+        )
+    }
+
+    /// Wraps the element in a transition container to provide an animated transition when its visibility changes.
+    ///
+    /// - Parameters:
+    ///   - appearAndDisappear: The transition to use when the element appears and disappears.
+    func transition(appearAndDisappear: VisibilityTransition) -> TransitionContainer {
+        TransitionContainer(
+            wrapping: self,
+            appearingTransition: appearAndDisappear,
+            disappearingTransition: appearAndDisappear,
+            layoutTransition: .none
+        )
+    }
 }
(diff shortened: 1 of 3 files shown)


kyleve (approved):

BlueprintUICommonControls/Sources/TransitionContainer.swift
```diff
@@ -6,14 +6,22 @@ import UIKit
...
+    public var appearingTransition: VisibilityTransition
+    public var disappearingTransition: VisibilityTransition
+    public var layoutTransition: LayoutTransition
 
     public var wrappedElement: Element
 
-    public init(wrapping element: Element) {
+    public init(
+        wrapping element: Element,
+        appearingTransition: VisibilityTransition = .fade,
```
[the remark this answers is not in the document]
nononoah: I considered it, though refrained as It appeared that we typically reserved `on` usage for closures that described event handling.

That said, semantically, there's a strong connection between these closures and these animations. I'll update.

BlueprintUICommonControls/Sources/TransitionContainer.swift
```diff
@@ -29,3 +37,38 @@ public struct TransitionContainer: Element {
...
+            disappearingTransition: disappear,
+            layoutTransition: layout
+        )
+    }
+
+    /// Wraps the element in a transition container to provide an animated transition when its visibility changes.
+    ///
+    /// - Parameters:
+    ///   - appearAndDisappear: The transition to use when the element appears and disappears.
+    func transition(appearAndDisappear: VisibilityTransition) -> TransitionContainer {
```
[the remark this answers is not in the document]
nononoah: I like it, and also like that it let's me dodge this rather unfortunate naming.

BlueprintUICommonControls/Sources/TransitionContainer.swift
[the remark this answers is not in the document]
nononoah: Update: it looks like our convention is to reuse the property name at the init site (checked `Box`, `Image`, `Button`, and `Rule`).

I'm going to use the `on` syntax for the extension on `Element` and leave this be for now. I'm happy to update it however if you think it's best.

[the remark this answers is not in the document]
kyleve: Up to you, though the convention isnt really set in stone either way; it just so happens that most parameters don't convey actions. Note it's also Swift's convention to leave type names off of parameters (eg children instead of childViewControllers), because the type system conveys that for us.

BlueprintUICommonControls/Sources/TransitionContainer.swift
```diff
@@ -29,3 +37,38 @@ public struct TransitionContainer: Element {
...
+    /// Wraps the element in a transition container to provide an animated transition.
+    ///
+    /// - Parameters:
+    ///   - onAppear: The transition to use when the element appears. By default, `.none`.
+    ///   - onDisappear: The transition to use when the element disappears. By default, `.none`.
+    ///   - onLayout: The animation to use when the element changes layout. By default, `.none`.
+    func transition(
+        onAppear: VisibilityTransition = .none,
+        onDisappear: VisibilityTransition = .none,
+        onLayout: LayoutTransition = .none
```
watt, line 52: I'm not sure these are the quite the defaults we want. In particular, the `.none` layout transition explicitly prevents inherited animations from taking effect on this element or its children.

It might be better to match the defaults on `ViewDescription` (`nil` for visibility, `inherited` for layout), so that not specifying one of these parameters will get you the same behavior as if you didn't have a TransitionContainer at all.

kyleve: Ahh shit, you're right. I forgot about this distinction before merging.

kyleve: @nononoah: Mind PRing this change? If not, I can do it since I did merge this.

watt: I have some thoughts on taking this a little further, which I'm drafting in #158.

nononoah: In the discussion, I mentioned that I was unsure of the direction of `.none`, but I was also unaware of its potential interaction with inherited animations.

If I understand your point correctly, any transition will override inherited transitions. This is reasonable behavior, and it should probably be documented in `TransitionContainer`. 

#158 looks like the alternative I first wrote up, but discarded as it was more disruptive. Please let me know if there's anything else I can do to help short of reviewing the PR, which I may not have permission to do.

watt: @nononoah No worries! I'll try to write a better explanation of the changes I'm suggesting in #158.

PR #156, opened 2020-09-21
Remove grid layout

kyleve:
Remove GridLayout; it's incomplete and was never really intended to be consumed widely. The intended replacement is putting EqualStacks inside of a Column, or Rows inside a Column


kylebshr (approved):

bencochran (approved):
I think there’s a case for _something_ like Grid that’s not covered by EqualStack + Column (that is, equal-sized items across all rows and columns), but agree that the current state of this layout is pretty incomplete and could use a re-think overall.

kyleve:
Yeah, agree. This ain't it, especially since this is a layout, not an element.

PR #157, opened 2020-09-21
Properly constrain EqualStack measurement

bencochran:
We know the constrained size when measuring an EqualStack, since it is divided evenly. Use that knowledge when measuring. This fixes measurement issues when adding multiline text elements to an EqualStack.

**Before:**
before
**After:**
after

### DIFF
--- a/BlueprintUI/Sources/Layout/EqualStack.swift
+++ b/BlueprintUI/Sources/Layout/EqualStack.swift
@@ -65,7 +65,21 @@ extension EqualStack {
         var spacing: CGFloat
 
         func measure(in constraint: SizeConstraint, items: [(traits: Void, content: Measurable)]) -> CGSize {
-            let itemSizes = items.map { $1.measure(in: constraint) }
+            let totalSpacing = (spacing * CGFloat(items.count - 1))
+            let itemConstraint: SizeConstraint
+            switch direction {
+            case .horizontal:
+                itemConstraint = SizeConstraint(
+                    width: (constraint.width - totalSpacing) / CGFloat(items.count),
+                    height: constraint.height
+                )
+            case .vertical:
+                itemConstraint = SizeConstraint(
+                    width: constraint.width,
+                    height: (constraint.height - totalSpacing) / CGFloat(items.count)
+                )
+            }
+            let itemSizes = items.map { $1.measure(in: itemConstraint) }
 
             let maximumItemWidth = itemSizes.map { $0.width }.max() ?? 0
             let maximumItemHeight = itemSizes.map { $0.height }.max() ?? 0
@@ -88,16 +102,17 @@ extension EqualStack {
         func layout(size: CGSize, items: [(traits: (), content: Measurable)]) -> [LayoutAttributes] {
             guard items.count > 0 else { return [] }
 
+            let totalSpacing = (spacing * CGFloat(items.count - 1))
             let itemSize: CGSize
             switch direction {
             case .horizontal:
                 itemSize = CGSize(
-                    width: (size.width - (spacing * CGFloat(items.count - 1))) / CGFloat(items.count),
+                    width: (size.width - totalSpacing) / CGFloat(items.count),
                     height: size.height)
             case .vertical:
                 itemSize = CGSize(
                     width: size.width,
-                    height: (size.height - (spacing * CGFloat(items.count - 1))) / CGFloat(items.count))
+                    height: (size.height - totalSpacing) / CGFloat(items.count))
             }
 
             var result: [LayoutAttributes] = []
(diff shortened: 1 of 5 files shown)


kyleve (approved):
Dope, thanks!

CHANGELOG.md as well!

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -65,7 +65,21 @@ extension EqualStack {
         var spacing: CGFloat
 
         func measure(in constraint: SizeConstraint, items: [(traits: Void, content: Measurable)]) -> CGSize {
-            let itemSizes = items.map { $1.measure(in: constraint) }
+            let totalSpacing = (spacing * CGFloat(items.count - 1))
```
kylebshr, line 68: Will this cause any issues if there's no elements, and the spacing is negative?

kylebshr: (I guess the array of sizes will just be empty too, so probably not)

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -114,5 +127,51 @@ extension SizeConstraint {
             }
         }
 
+        /// Divides an Axis by a scalar value. If the Axis is unconstrained the
+        /// result will remain unconstrained.
+        public static func /(lhs: SizeConstraint.Axis, rhs: CGFloat) -> SizeConstraint.Axis {
```
kylebshr, line 132: Think it's worth adding tests for these operators?

kylebshr (approved):

BlueprintUI/Sources/Layout/EqualStack.swift
bencochran: Yeah, the `.max() ?? 0` below handles the zero-element case, but we could probably add an early bail to be more explicit and avoid the negative spacing weirdness

BlueprintUI/Sources/Measuring/SizeConstraint.swift
bencochran: Not a _ton_ of code in these, but it’s nice to have a seatbelt test just in case I mixed up a `+` and a `*` or something. Added.

BlueprintUI/Sources/Layout/EqualStack.swift
bencochran: Added an early return

BlueprintUI/Sources/Measuring/SizeConstraint.swift
kylebshr: Thanks!

BlueprintUI/Tests/EqualStackTests.swift
```diff
@@ -61,6 +61,77 @@ class EqualStackTests: XCTestCase {
...
+                    stack.spacing = 0
+                    stack.children = children
+                }
+
+                // 300 / 3 = 100
+                // 1200 / 100 = 12
+                XCTAssertEqual(stack.content.measure(in: constraint), CGSize(width: 300, height: 12))
+            }
+
+            // spacing = 50
```
watt, line 88: Your comment doesn't match your test value

BlueprintUI/Tests/EqualStackTests.swift
```diff
@@ -61,6 +61,77 @@ class EqualStackTests: XCTestCase {
...
+                // 300 - 30 - 30 = 240
+                // 240 / 3 = 80
+                // 1200 / 80 = 15
+                XCTAssertEqual(stack.content.measure(in: constraint), CGSize(width: 300, height: 15))
+            }
+        }
+
+        // direction = .vertical
+        do {
+            let constraint = SizeConstraint(width: .unconstrained, height: .atMost(400))
```
watt, line 104: The vertical tests use different numbers?

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -65,7 +65,23 @@ extension EqualStack {
...
         func measure(in constraint: SizeConstraint, items: [(traits: Void, content: Measurable)]) -> CGSize {
-            let itemSizes = items.map { $1.measure(in: constraint) }
+            guard items.count > 0 else { return .zero }
+
+            let totalSpacing = (spacing * CGFloat(items.count - 1))
+            let itemConstraint: SizeConstraint
+            switch direction {
+            case .horizontal:
+                itemConstraint = SizeConstraint(
+                    width: (constraint.width - totalSpacing) / CGFloat(items.count),
```
watt, line 75: These asymmetrical operators were slightly surprising at first glance but it does seem a lot better than switching inline.

watt (approved):

BlueprintUI/Tests/EqualStackTests.swift
bencochran: Ah, yep, changed it later to make the math in the comment cleaner. Will update the comment

BlueprintUI/Tests/EqualStackTests.swift
bencochran: Yep, that was intentional just to have some variance in the tests

PR #160, opened 2020-09-23
[DRAFT| Inline conditionals

kyleve:
Inspired by https://fivestars.blog/swiftui/conditional-modifiers.html

### DIFF
--- a/BlueprintUI/Sources/Layout/Conditionals.swift
+++ b/BlueprintUI/Sources/Layout/Conditionals.swift
@@ -79,8 +79,17 @@ extension Element {
         return copy
     }
     
-    func modify(_ modify : (Self) -> Element) -> Element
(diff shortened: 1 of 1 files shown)


BlueprintUI/Sources/Layout/Conditionals.swift
```diff
@@ -0,0 +1,93 @@
...
+            modify(&copy)
+            return copy
+        } else {
+            return self
+        }
+    }
+    
+    func `if`(
+        _ isTrue : Bool,
+        onTrue : (inout Self) -> (),
```
kyleve, line 31: Actually, should probably convert this to return; or add a second override for return-based.

kyleve:
This is part of KB reader PR now

PR #161, opened 2020-09-24
Remove support for iOS 10

kyleve:
This removes support for iOS 10 since SPOS has now dropped it. Note that I have _not_ changed CI in this PR; since it's significantly re-worked here: https://github.com/square/Blueprint/pull/152, once this lands, we can then update to only run tests on 11 through 14.


SampleApp/Sources/PostsViewController.swift
```diff
@@ -43,14 +43,6 @@ final class PostsViewController: UIViewController {
         blueprintView.element = element
     }
 
-    private var viewSafeAreaInsets: UIEdgeInsets {
```
kyleve, line 46: This probably broke the tutorial build. Also a good time to make sure these build as part of the SampleApp!

SampleApp/SampleApp.xcodeproj/xcshareddata/xcschemes/SampleApp.xcscheme
```diff
@@ -20,6 +20,62 @@
                ReferencedContainer = "container:SampleApp.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
```
kyleve, line 23: This makes the tutorials build as part of the Cocoapods build; so we can't accidentally break them.

SampleApp/Sources/PostsViewController.swift
kyleve: Nope, it didnt; wasnt used! But added tutorials as part of SampleApp build.

kylebshr:
This is breaking since CI currently runs on iOS 10 - how about we merge #152 first, and drop the iOS 10 device from this PR?

kyleve:
Yep, sounds good

kyleve:
(I actually intended that, but forgot to note it here; which is why I didnt touch the .travis.yml file)

watt (approved):

kylebshr:
It's merged! You can rebase and remove the iOS 10 sim from the matrix

kylebshr:
Went ahead and updated it again since I merged the yml change and it was conflicting

PR #162, opened 2020-09-24
Only build main branch

kylebshr:
Travis is set to build PRs and Branches (so main will build on merge). This means tests actually run twice for PRs. This change means **PRs** will build on creation & pushes, but only the main **branch** will build on pushes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ sudo: false
 notifications:
   email: false
 
+branches:
+  only:
+    - master
+
 os: osx
 osx_image: xcode12
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
 
 branches:
   only:
-    - master
+    - main
 
 os: osx
 osx_image: xcode12


.travis.yml
```diff
@@ -3,6 +3,10 @@ sudo: false
 notifications:
   email: false
 
+branches:
+  only:
+    - master
```
bencochran, line 8: should this be `main`?

bencochran: nvm, you’re too fast

kylebshr: lmao just caught that myself. it's updated!

bencochran (approved):

kyleve (approved):

watt:
I think the existing behavior is desirable? Building your branch by itself is different from building the result after merging it.

kylebshr:
> I think the existing behavior is desirable? Building your branch by itself is different from building the result after merging it.

The existing behavior builds branches that aren't PRs, and in doing so, builds for PRs twice. Most projects don't build branches until a PR is opened, but do build PRs, which this change matches. You can see this PR still has the travis job running - old PRs have two jobs for no reason.

watt (approved):

PR #165, opened 2020-10-01
Implement keyboard reader, which allows observing the position of the keyboard on-screen.

kyleve:
This PR introduces a `KeyboardReader` type, which allows you to programmatically respond to the keyboard frame as it changes. It does this by maintaining an inner `BlueprintView` which updates when the keyboard updates.

The simplified use case looks like this:

```
KeyboardReader { context in
   myElement.inset(bottom: context.keyboardFrame.height)
}
```
Where you return an element in a provider, which updates when the keyboard frame is changed.

There's also some convenience initializers for layering, say, a background (like a list or table) which already respects a keyboard below, eg a bottom button that you want to hug the keyboard:

```
KeyboardReader {
    myList
} adjustedForKeyboard: { context in
    myButton.aligned(horizontal: .fill, vertical: .bottom)
}
```

### Changes

1) As mentioned, introduce `KeyboardReader` which allows responding to keyboard positioning within an element.

2) From a discussion on Slack; introduce `Conditionals.swift`, which allows `if`, `if/else`, `if/let`, etc, operators on elements, to reduce the need for temporary variables and nests logic visually when writing code.

3) Alongside the demo included, add some helpers to `Overlay` to make building `Overlay`s easier.

4) Add similar helpers to `StackElement`.

5) Introduce `inheritedEnvironment` on `BlueprintView`, so when we're passing through to an internal `BlueprintView`, we can pass through the base environment to use. I went this route instead of using an `AdaptedEnvironment` because we want the inner `BlueprintView` to override the inherited properties with its own properties (eg, the safe area), but still want to pass through the other values from the outer environment. Because the `Key`s used to write into the environment are private, there's no good way to provide a generic merging strategy here.

6) Change `BlueprintView` and its child views to update how touches are handled; instead of absorbing all touches, instead only absorb touches on subviews. This is done by calling through to `self.point(insideSubviewsOnly: point, with: event)`. This is needed because the `KeyboardReader` maintains an internal `BlueprintView`; but we want to pass through touches to lower views as long as those touches do not intersect with an element. Without this change, `KeyboardReader` erroneously ends up eating touches.

7) Copy over some changes from Listable into `KeyboardObserver`, which are needed to move the keyboard observer to be a shared observer – it needs to be a shared observer, because the only way for us to know what the keyboard frame is, is to be constantly observing the keyboard did change notifications. Also copied over the tests.


BlueprintUI/Sources/Layout/KeyboardObserver.swift
```diff
@@ -1,14 +1,14 @@
 //
```
kyleve, line 1: Changes in this file copied from Listable, in particular the `shared` instance, which ensures the keyboard frame is always being observed.

BlueprintUI/Sources/Layout/Overlay.swift
```diff
@@ -14,6 +14,20 @@ public struct Overlay: Element {
     public init(elements: [Element]) {
```
kyleve, line 14: Document me

BlueprintUI/Tests/KeyboardObserverTests.swift
```diff
@@ -1,11 +1,83 @@
 import XCTest
```
kyleve, line 1: Also copied from Listable

BlueprintUI/Sources/Internal/Extensions/UIView.swift
```diff
@@ -2,8 +2,28 @@ import UIKit
...
+    /// An implementation of `point(inside:with:)` that only hit tests on visible subviews,
+    /// bypassing hit testing on the view instance itself, making it transparent to the UIKit hit-testing system.
+    ///
+    /// If you are implementing a custom `UIView` or `Element` that manages displaying its own `Element`,
+    /// you will likely want to override `point(inside:with:)` and call this method – this ensures that
+    /// your view won't occlude touches to views lower in the view hierarchy.
+    func point(insideSubviewsOnly point: CGPoint, with event: UIEvent?) -> Bool {
+        
+        self.subviews.first { subview in
+            guard subview.isHidden == false && subview.alpha > 0 else {
```
kyleve, line 19: Remove the `subview.alpha > 0` check – UIKit doesn't do this.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -48,17 +48,73 @@ public final class BlueprintView: UIView {
...
+    ///         ...
+    ///
+    ///         private func updateElement() {
+    ///             self.blueprintView.inheritedEnvironment = self.environment
+    ///             self.blueprintView.element = self.provider()
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    public var inheritedEnvironment : Environment? = nil {
```
kyleve, line 100: Just remove this and used `AdaptedEnvironment`?

kyleve: Never mind, this makes merging the environments together difficult. We want to use this as a base, and then have the view override its defined params.

BlueprintUI/Sources/Layout/KeyboardReader.swift
```diff
@@ -0,0 +1,228 @@
...
+/// The initializer for `KeyboardReader` takes an escaping closure, which is called each time
+/// the keyboard frame changes, or the element's frame changes, in order to update the element
+/// to account for the new keyboard position.
+///
+/// ```
+/// KeyboardReader { info in
+///     myElement.inset(bottom: info.keyboardFrame.height)
+/// }
+/// ```
+public struct KeyboardReader : ProxyElement {
```
kyleve, line 29: Likely also consider some sort of inputAccessoryView integration here? Eg for bottom-pinned buttons or bars, that's the only way to get the interactive dismissal behaviour.

kyleve: This definitely simplifies a lot of common cases

PR #166, opened 2020-10-14
[DRAFT] Support individual corner rounding on Box

kyleve:
Update the `case rounded(radius: CGFloat)` case on `Box` to `case rounded(radius: CGFloat, corners : UIRectCorner = .allCorners)`, to allow rounding specific corners.

Note I also deleted all the iOS 10 snapshot images while updating the snapshot tests, since we now only support iOS 11+.

**Important Note**: This PR also moves us back to using Cocoapods for running snapshot tests. Why? Turns out `maskedCorners` on `CALayer` are new-ish, and don't render standalone in the `layer.render(in:)` API we were using in the snapshot tests. The layer needs to live in an app host for this to work, apparently: https://github.com/pointfreeco/swift-snapshot-testing/issues/358


BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -38,85 +38,74 @@ public struct Box: Element {
...
-
-
+        Box.View.describe { config in
+            
+            config.builder = {
+                Box.View(frame: bounds, model: self.toViewModel)
+            }
+            
+            config.apply { view in
+                view.model = self.toViewModel
```
kyleve, line 60: Move application of the element's properties onto the view itself.

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -214,34 +204,160 @@ extension Box.ShadowStyle {
...
+        
+        private var needsUpdateFromModel : Bool = true
+        private var lastSize : CGSize? = nil
+        
+        override func layoutSubviews() {
+            super.layoutSubviews()
+            
+            contentView.frame = bounds
+
+            if self.lastSize != self.bounds.size || self.needsUpdateFromModel {
```
kyleve, line 277: Ensures we only update the on-screen view when we actually have to

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -214,34 +204,160 @@ extension Box.ShadowStyle {
...
     
-    required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+    fileprivate final class View : UIView {
+        
+        let contentView = UIView()
+        
+        var model : ViewModel {
+            didSet {
+                guard oldValue != self.model else { return }
```
kyleve, line 247: Ensures we only update the on-screen view when we actually have to

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -214,34 +204,160 @@ extension Box.ShadowStyle {
...
+                self.model.applyTo(view: self)
+            }
+        }
     }
     
-    override func layoutSubviews() {
-        super.layoutSubviews()
-        contentView.frame = bounds
+    
+    fileprivate struct ViewModel : Equatable {
```
kyleve, line 287: We utilize a `ViewModel` instead of simply passing through the `Box`, since the `Box's` wrapped element could contain long lived closures, etc, so want to avoid capturing that context.

PR #60, opened 2020-02-27
[WIP DNR] Stack View Layout Rewrite for #58?

kyleve:
Don't review this please

PR #168, opened 2020-10-18
[Layout] Add `ConstrainedSizeFraction`, a fraction-based sizing strategy

nononoah:
#60 
### Summary

`ConstrainedSizeFraction ` describes the layout of children which occupy a fraction of their parents' layout space. 

Previously, this effect could be achieved by composing a `ConstrainedSize` layout as a child of a `GeometryReader`.

```swift
GeometryReader { geometry in
  Row { row in 
    ...
    let availableWidth = geometry.constraint.width.maximum
    row.addFixed(child: self.element.constrainedTo(width: .absolute(width * 0.25)))
  }
}
```

`ConstrainedPercent` allows the following alternative:

```swift
Row { row in
  ...
  row.addFixed(child: element.constrainedTo(widthFraction: 0.25))
}
```

### Alternatives

It seemed tidy and semantically correct to add a `.fraction(width:height:)` case to `ConstrainedSize`, but the [layout strategy](https://github.com/square/Blueprint/blob/main/BlueprintUI/Sources/Layout/ConstrainedSize.swift#L126-L141) would require special casing much like that found [here](https://github.com/square/Blueprint/pull/168/files#diff-fa05a4a771a01f26a2434eb090466ebaaaabd36db6221e8d2766e812a333c0c3R43-R60). 

Perhaps someone has a special insight here about how we could make `apply` idempotent in the `percent` case (akin to `.absolute`), but such insight eluded me.

### Additional considerations

If there's a need, we should consider adding an API parallel to that of `ConstrainedSize` (e.g. `.absolute`, `.atLeast`, `.atMost`). Since I've only ever personally found a need for the `.absolute` case, I decided to start with it exclusively. 

### Looking for help

As always, I defer to the group's best judgment when it comes to naming, as I'm still finding my way.

Additionally, I should note that I'm a bit out of my element when it comes to understanding the layout system, but I've tested this on production code to positive results. In one case, a `Row` with a `minimumHorizontalSpacing`, percentage-based elements, fixed elements, and flexible elements maintained the exact layout it had previously, down to the coordinate point.

Please let me know if there's anything I can do to build further confidence in these changes.

### DIFF
--- a/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
@@ -6,38 +6,38 @@ public struct ConstrainedSizeFraction: Element {
     public var wrappedElement: Element
     /// The fraction of the parent's horizontal layout space the content element should occupy. `nil` if the width is
     /// unconstrained.
-    public var width: CGFloat?
+    public var fractionalWidth: CGFloat?
     /// The fraction of the parent's vertical layout space the content element should occupy. `nil` if the height is
     /// unconstrained.
-    public var height: CGFloat?
+    public var fractionalHeight: CGFloat?
 
     /// Initializes with the given properties.
     ///
     /// - parameters:
-    ///   - width: The fraction of the parent's horizontal layout space the content element should occupy. `nil` if
+    ///   - fractionalWidth: The fraction of the parent's horizontal layout space the content element should occupy. `nil` if
     ///   the width is unconstrained. By default, `nil`.
-    ///   - height: The fraction of the parent's vertical layout space the content element should occupy. `nil` if
+    ///   - fractionalHeight: The fraction of the parent's vertical layout space the content element should occupy. `nil` if
     ///   the height is unconstrained. By default, `nil`.
     ///   - wrapping: The content element.
-    public init(width: CGFloat? = nil, height: CGFloat? = nil, wrapping wrappedElement: Element) {
+    public init(fractionalWidth: CGFloat? = nil, fractionalHeight: CGFloat? = nil, wrapping wrappedElement: Element) {
         precondition(
-            width.map { $0 >= 0 && $0 <= 1} ?? true,
+            fractionalWidth.map { $0 >= 0 && $0 <= 1} ?? true,
             "The provided width fraction must be a value in the range of `0...1`."
         )
         precondition(
-            height.map { $0 >= 0 && $0 <= 1} ?? true,
+            fractionalHeight.map { $0 >= 0 && $0 <= 1} ?? true,
             "The provided height fraction must be a value in the range of `0...1`."
         )
 
-        self.width = width
-        self.height = height
+        self.fractionalWidth = fractionalWidth
+        self.fractionalHeight = fractionalHeight
         self.wrappedElement = wrappedElement
     }
 
     public var content: ElementContent {
         ElementContent(
             child: wrappedElement,
-            layout: Layout(width: width, height: height)
+            layout: Layout(width: fractionalWidth, height: fractionalHeight)
         )
     }
 
@@ -80,15 +80,15 @@ public extension Element {
     /// Constrains the element to a fraction of its parent's size.
     ///
     /// - parameters:
-    ///   - width: The fraction of the parent's horizontal layout space the element should occupy. `nil`
+    ///   - fractionalWidth: The fraction of the parent's horizontal layout space the element should occupy. `nil`
     ///   if the width is unconstrained. By default, `nil`.
-    ///   - height: The fraction of the parent's vertical layout space the element should occupy. `nil`
+    ///   - fractionalHeight: The fraction of the parent's vertical layout space the element should occupy. `nil`
     ///   if the height is unconstrained. By default, `nil`.
     func constrainedTo(
-        widthFraction width: CGFloat? = nil,
-        heightFraction height: CGFloat? = nil
+        fractionalWidth: CGFloat? = nil,
+        fractionalHeight: CGFloat? = nil
     ) -> ConstrainedSizeFraction
     {
-        ConstrainedSizeFraction(width: width, height: height, wrapping: self)
+        ConstrainedSizeFraction(fractionalWidth: fractionalWidth, fractionalHeight: fractionalHeight, wrapping: self)
     }
 }
--- a/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
@@ -21,12 +21,12 @@ public struct ConstrainedSizeFraction: Element {
     ///   - wrapping: The content element.
     public init(width: CGFloat? = nil, height: CGFloat? = nil, wrapping wrappedElement: Element) {
         precondition(
-            width.map { $0 > 0 && $0 <= 1} ?? true,
-            "The provided width fraction must be greater than 0 and less than or equal to 1."
+            width.map { $0 >= 0 && $0 <= 1} ?? true,
+            "The provided width fraction must be a value in the range of `0...1`."
         )
         precondition(
-            height.map { $0 > 0 && $0 <= 1} ?? true,
-            "The provided height fraction must be greater than 0 and less than or equal to 1."
+            height.map { $0 >= 0 && $0 <= 1} ?? true,
+            "The provided height fraction must be a value in the range of `0...1`."
         )
 
         self.width = width
--- a/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
@@ -1,0 +1,94 @@
+import UIKit
+
+/// Constrains the size of the content element to a fraction of the parent's layout space.
+public struct ConstrainedSizeFraction: Element {
+    /// The element being constrained.
+    public var wrappedElement: Element
+    /// The fraction of the parent's horizontal layout space the content element should occupy. `nil` if the width is
+    /// unconstrained.
+    public var width: CGFloat?
+    /// The fraction of the parent's vertical layout space the content element should occupy. `nil` if the height is
+    /// unconstrained.
+    public var height: CGFloat?
+
+    /// Initializes with the given properties.
+    ///
+    /// - parameters:
+    ///   - width: The fraction of the parent's horizontal layout space the content element should occupy. `nil` if
+    ///   the width is unconstrained. By default, `nil`.
+    ///   - height: The fraction of the parent's vertical layout space the content element should occupy. `nil` if
+    ///   the height is unconstrained. By default, `nil`.
+    ///   - wrapping: The content element.
+    public init(width: CGFloat? = nil, height: CGFloat? = nil, wrapping wrappedElement: Element) {
+        precondition(
+            width.map { $0 > 0 && $0 <= 1} ?? true,
+            "The provided width fraction must be greater than 0 and less than or equal to 1."
+        )
+        precondition(
+            height.map { $0 > 0 && $0 <= 1} ?? true,
+            "The provided height fraction must be greater than 0 and less than or equal to 1."
+        )
+
+        self.width = width
+        self.height = height
+        self.wrappedElement = wrappedElement
+    }
+
+    public var content: ElementContent {
+        ElementContent(
+            child: wrappedElement,
+            layout: Layout(width: width, height: height)
+        )
+    }
+
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        return nil
+    }
+
+    private struct Layout: SingleChildLayout {
+        var width: CGFloat?
+        var height: CGFloat?
+
+        func measure(in sizeConstraint: SizeConstraint, child: Measurable) -> CGSize {
+            /// Get the target layout space, constraining it by the specified fractions.
+            let targetWidth = width.map { $0 * sizeConstraint.width.maximum }
+            let targetHeight = height.map { $0 * sizeConstraint.height.maximum }
+
+            let constraint = SizeConstraint(
+                width: targetWidth.map { .atMost($0) } ?? .unconstrained,
+                height: targetHeight.map { .atMost($0) } ?? .unconstrained
+            )
+
+            /// Measure the child in the constrained layout space.
+            let measurement = child.measure(in: constraint)
+
+            /// Override the child's measurement with a target dimension if one exists.
+            return CGSize(
+                width: targetWidth ?? measurement.width,
+                height: targetHeight ?? measurement.height
+            )
+        }
+
+        func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
+            return LayoutAttributes(size: size)
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


kylebshr:
This seems like a great addition - haven't had a chance to properly review it, but what do you think about using `fraction` instead of percent for all the names? "Percent" translates a little weirdly to code IMO, since it's written as a fraction * 100 in English, but in code expressed as just the fraction. 

There's some precedence for this in UIKit API too, i.e., `UIViewPropertyAnimator.fractionComplete`. 

Will take a look at the actual code tomorrow or Monday!

nononoah:
> This seems like a great addition - haven't had a chance to properly review it, but what do you think about using `fraction` instead of percent for all the names? "Percent" translates a little weirdly to code IMO, since it's written as a fraction * 100 in English, but in code expressed as just the fraction.
> 
> There's some precedence for this in UIKit API too, i.e., `UIViewPropertyAnimator.fractionComplete`.
> 
> Will take a look at the actual code tomorrow or Monday!

I'm a big fan of this suggestion, and in fact, I started with `fraction`, but leaned `percent` after KVE suggested as much (implicitly in a response).

BlueprintUI/Sources/Layout/ConstrainedPercent.swift
```diff
@@ -0,0 +1,85 @@
...
+public extension Element {
+    /// Constrains the element to a percentage of its parent's size.
+    ///
+    /// - parameters:
+    ///   - widthPercent: The percentage of the parent's width the content element should occupy. Values between 0-1
+    ///   correspond to 0-100%. `nil` if the width is unconstrained. By default, `nil`.
+    ///   - heightPercent: The percentage of the parent's height the content element should occupy. Values between 0-1
+    ///   correspond to 0-100%. `nil` if the height is unconstrained. By default, `nil`.
+    func constrainedTo(
+        widthPercent: CGFloat? = nil,
```
kyleve, line 79: +1 to fraction from @kylebshr

BlueprintUI/Tests/ConstrainedPercentTests.swift
```diff
@@ -0,0 +1,44 @@
+import XCTest
+import BlueprintUI
+
+class ConstrainedPercentTests: XCTestCase {
```
kyleve, line 4: It would be good to add a test using a Flexible Element like we do in constrained size, to ensure that elements that "re-flow" (like text, where the width affects the height as the text re-flows) are handled in a reasonable way: https://github.com/square/Blueprint/blob/main/BlueprintUI/Tests/ConstrainedSizeTests.swift#L340

kyleve (approved):
LGTM with a couple suggestions

BlueprintUI/Sources/Layout/ConstrainedPercent.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// Initializes with the given properties.
+    ///
+    /// - parameters:
+    ///   - width: The fraction of the parent's horizontal layout space the content element should occupy. `nil` if
+    ///   the width is unconstrained. By default, `nil`.
+    ///   - height: The fraction of the parent's vertical layout space the content element should occupy. `nil` if
+    ///   the height is unconstrained. By default, `nil`.
+    ///   - wrapping: The content element.
+    public init(width: CGFloat? = nil, height: CGFloat? = nil, wrapping wrappedElement: Element) {
+        self.width = width
```
kyleve, line 23: Do we want to validate the width and height is between 0 and 1? Could end up with some weirddd behavior if someone, say, passes in -1 or 2, etc. Precondition checks would be valuable.

BlueprintUI/Sources/Layout/ConstrainedPercent.swift
nononoah: Updated `ConstrainedPercent` to `ConstrainedSizeFraction`. This site became `widthFraction` in the shift.

BlueprintUI/Sources/Layout/ConstrainedPercent.swift
kyleve: (either that, or clamp between 0 and 1)

BlueprintUI/Tests/ConstrainedPercentTests.swift
nononoah: So, I spent some time looking at this example, and it doesn't seem to apply. `ConstrainedSizeFraction` currently functions like `ConstainedSize's` `.absolute` case in that it overrides element measurements when a constraint exists in a given dimension.

Please correct me if I'm off base here!

BlueprintUI/Sources/Layout/ConstrainedPercent.swift
nononoah: I'm open to this, and I considered it myself, but stopped short because I didn't know our general disposition here. I'll add a check.

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// Initializes with the given properties.
+    ///
+    /// - parameters:
+    ///   - width: The fraction of the parent's horizontal layout space the content element should occupy. `nil` if
+    ///   the width is unconstrained. By default, `nil`.
+    ///   - height: The fraction of the parent's vertical layout space the content element should occupy. `nil` if
+    ///   the height is unconstrained. By default, `nil`.
+    ///   - wrapping: The content element.
+    public init(width: CGFloat? = nil, height: CGFloat? = nil, wrapping wrappedElement: Element) {
+        precondition(width.map { $0 >= 0 && $0 <= 1} ?? true)
```
kyleve, line 23: Mind adding an error message as well? Eg `The provided fraction must be greater than zero or less than or equal to one, to be a valid fractional width.`

kyleve: (Also, the check should probably be `> 0`, not `>= 0`? It doesn't _hurt_, a zero-width item doesnt really mean anything.

BlueprintUI/Tests/ConstrainedPercentTests.swift
kyleve: Got it, I think that makes sense

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
nononoah: Updated, thanks.

kyleve:
Still LGTM

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
kylebshr: Sorry for the thrash here, but don't a lot of fraction based UI's end up using 0? A progress bar for example

nononoah: Seems like a valid use case to me. I'll restore it.

kyleve: Maybe in the case of zero, it makes sense to not include the child element from the elementContent and to short circuit the measurement? Eg still worried about all the possible divide by zeros we might run into in the wild, since elements rarely are given an empty size.

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
```diff
@@ -0,0 +1,94 @@
...
+public extension Element {
+    /// Constrains the element to a fraction of its parent's size.
+    ///
+    /// - parameters:
+    ///   - width: The fraction of the parent's horizontal layout space the element should occupy. `nil`
+    ///   if the width is unconstrained. By default, `nil`.
+    ///   - height: The fraction of the parent's vertical layout space the element should occupy. `nil`
+    ///   if the height is unconstrained. By default, `nil`.
+    func constrainedTo(
+        widthFraction width: CGFloat? = nil,
```
kylebshr, line 88: Naming nit: what do you think about `fractionalWidth` / `fractionalHeight`? could also use for the properties of the element themselves, it's not immediately obvious that setting the `element.width` is setting a fractional width

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
kylebshr: Shouldn't we just fix those bugs if they come up? Elements aren't required not to have an empty size, are they? Math in Blueprint should defend against that. 

Aaaand one more thing I just thought of, why clamp at one? Making an element 2x the width of another one also seems like a perfectly valid use case 🤔

kyleve: But in this case, its the width of the _parent_, which is weird (and probably wrong, eg for views that clip their subviews).

kylebshr (approved):
lgtm other than naming nits that I don't feel too strongly about! and the 1x clamp question

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
kylebshr: mmm good point.. would that make sense for `aligned` elements though? tbh I was thinking more about constraints, which you can set on any two elements in the tree 😅 leaving at 1 probably makes sense

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
nononoah: I don't mind the update here, and considered it. If anyone wants to tiebreak, this I'd welcome that, and if not, I'll just change it.

If we update the property name, we're adding a bit of a stutter `ConstrainedSizeFraction.widthFraction`, which I'd prefer to avoid. Again, not strongly opposed, so please let me know if you'd like me to make this change.

kyleve: Could all of the properties be updated to be `fractional*`, including the vars and init?

nononoah: Yep, that's the tiebreak! I will update now.

kyleve:
Still LGTM, feel free to merge whenever

CHANGELOG.md
```diff
@@ -33,6 +33,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  Previously, this effect could be achieved by composing a `ConstrainedSize` layout as a child of a `GeometryReader`.
+
+  ```swift
+  GeometryReader { geometry in
+    Row { row in 
+
+      ...
+
+      let availableWidth = geometry.constraint.width.maximum
+      row.addFixed(child: self.element.constrainedTo(width: .absolute(width * 0.25)))
```
watt, line 49: These examples are not functionally equivalent. `Row` measures its children twice: 
1. constrained by the full width of the row, to get "basis sizes"
2. constrained to the actual allocated width, to get the height

During the first measure, you'll get the same result, but during the second one, the child will wind up being doubly constrained to 1/16 of the available width. If the child is flexible, like a label, it could result in an incorrect height.

Here's an example that demonstrates the difference:
```swift
    var element: Element {
        Column { col in
            col.minimumVerticalSpacing = 1
            col.add(child: a)
            col.add(child: b)
        }
        .centered()
    }

    var a: Element {
        GeometryReader { geometry in
            let width = geometry.constraint.width.maximum
            return Row { row in
                row.add(child: Spacer(width: 300, height: 20).box(background: .red))
                row.add(child: Label(text: "a b c d e f g h i j k l m n o")
                    .box(background: .green)
                    .constrainedTo(width: .absolute(width * 0.25)))
            }
        }
    }

    var b: Element {
        Row { row in
            row.add(child: Spacer(width: 300, height: 20).box(background: .red))
            row.add(child: Label(text: "a b c d e f g h i j k l m n o")
                .box(background: .green)
                .constrainedTo(fractionalWidth: 0.25))
        }
    }
```

Looks like this:

Screen Shot 2020-10-19 at 6 10 14 PM

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+
+public extension Element {
+    /// Constrains the element to a fraction of its parent's size.
+    ///
+    /// - parameters:
+    ///   - fractionalWidth: The fraction of the parent's horizontal layout space the element should occupy. `nil`
+    ///   if the width is unconstrained. By default, `nil`.
```
watt, line 84: Might not hurt to explain that the fraction is expressed as a value between 0 & 1, to be totally clear on that.

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
```diff
@@ -0,0 +1,94 @@
...
+        return nil
+    }
+
+    private struct Layout: SingleChildLayout {
+        var width: CGFloat?
+        var height: CGFloat?
+
+        func measure(in sizeConstraint: SizeConstraint, child: Measurable) -> CGSize {
+            /// Get the target layout space, constraining it by the specified fractions.
+            let targetWidth = width.map { $0 * sizeConstraint.width.maximum }
```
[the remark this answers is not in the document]
nononoah: My approach here is exactly the same as `ConstrainedSize's`.

```
let maximumConstraint = SizeConstraint(
    width: .init(self.width.applied(to: constraint.width.maximum)),
    height: .init(self.height.applied(to: constraint.height.maximum))
)
```

CHANGELOG.md
nononoah: The second sizing pass is going to be a potentially unresolvable problem for us. 

We assume that provided `SizeConstraint` is always the size constraint of the parent, which appears to be too naive. Do you have any ideas as to an alternative?

BlueprintUI/Sources/Layout/ConstrainedSizeFraction.swift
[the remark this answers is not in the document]
watt: Hmm, I think `ConstrainedSize` is probably wrong too now that you point it out, but it hasn't caused any problems in practice because it only affects `atLeast` constraints?

Regardless, I think it is a bit nonsensical to do fraction math on `unconstrained`. 50% of an infinite canvas is still an infinite canvas.

CHANGELOG.md
watt: Yeah, the measure function's contract is more like "given this much space, how big are you". There's no guarantee that the constraint is the size of the parent.

The GeometryReader solution seems reasonable in this case, where you actually want to base your size on the constraint passed to the row (not the constraints the row passes to its children).

Alternatively, if the main use case you're trying to solve is making it easier to create proportionally-divided stacks, you might want something new that works more like `EqualStack`. (Or maybe add ratio values to `EqualStack`.)

nononoah: OK, I'm going to close this one for now and consider such a path forward when I next have time to do so. Thanks Andrew.

watt: Sure thing, sorry for being the bearer of bad news! Happy to talk through this any time.

nononoah: Oh, please don't be sorry! I'm grateful for your insight, and not at all deterred from potentially landing a contribution that fulfills the semantics I'm looking for.

PR #175, opened 2020-10-30
Create NativeViewController without animations

kylebshr:
When creating a `NativeViewController` from a new, the layout attributes are applied with animations disabled. However, other attributes might be applied while building the view, which happens in the `NativeViewController` initializer - in order to prevent unintended animations, we also need to build the view with disabled animations. The node is also updated outside of `performWithoutAnimations`, and even though it does so with `appearanceTransitionsEnabled: false`, applying the update can be caught in an existing transition.

This is a little awkward, since line 359 needs to be outside of the `performWithoutAnimations` but also needs a reference to the controller; open to suggestions on how to do that without the IUO 😬

Example videos shared in slack.

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -342,17 +342,21 @@ extension BlueprintView {
                         controller.update(node: child, appearanceTransitionsEnabled: true)
                     }
                 } else {
-                    let controller = NativeViewController(node: child)
-                    newChildren.append((path: path, node: controller))
-                    
-                    UIView.performWithoutAnimation {
-                        child.layoutAttributes.apply(to: controller.view)
-                    }
-                    
-                    contentView.insertSubview(controller.view, at: index)
+                    var controller: NativeViewController!
 
-                    controller.update(node: child, appearanceTransitionsEnabled: false)
-                    
+                    // Building the view and applying the initial layout and update need to be wrapped in
+                    // performWithoutAnimation so they're not caught up inside an occuring transition.
+                    UIView.performWithoutAnimation {
+                        controller = NativeViewController(node: child)
+                        child.layoutAttributes.apply(to: controller.view)
+
+                        contentView.insertSubview(controller.view, at: index)
+
+                        controller.update(node: child, appearanceTransitionsEnabled: false)
+                    }
+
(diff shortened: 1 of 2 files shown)


kyleve (approved):

kyleve:
The IUO seems fine here; it's too bad `performWithoutAnimation` isnt generic on the return type of the closure.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -342,17 +342,19 @@ extension BlueprintView {
...
+                    // performWithoutAnimation so they're not caught up inside an occuring transition.
                     UIView.performWithoutAnimation {
+                        controller = NativeViewController(node: child)
                         child.layoutAttributes.apply(to: controller.view)
+                        controller.update(node: child, appearanceTransitionsEnabled: false)
                     }
-                    
+
+                    newChildren.append((path: path, node: controller))
                     contentView.insertSubview(controller.view, at: index)
```
kyleve, line 356: Note the update is now happening _after_ the insert subview... Would this break anything? I guess not; _unless_ the update call walks the view hierarchy?

kyleve: And the only element I know of that walks the view hierarchy in SPOS (ok, Dashboard) goes down the tree: https://git.sqcorp.co/projects/IOS/repos/register/browse/Frameworks/SquareSalesReports/Sources/Elements/ExclusiveTouch.swift#29

kylebshr: I thought about that a little, but I would hope if you're doing that you're not making assumptions about how blueprint works; and instead, using `didMoveToSuperview` or something. Do you think that's reasonable?

kyleve: Yeah, definitely – just figured I'd call it out

kylebshr: I don't think there's any harm to inserting the view in the block though, maybe I should just do that

watt (approved):

kylebshr:
Tested in Checkout Applet and Notification Center and didn't see any issues. Going to merge!

kyleve:
thumbs-up-guy-1

PR #176, opened 2020-10-30
Fix pull-to-refresh inset for iOS 13+

NSHaoSong:
On iOS 13, `UIRefreshControl` will change `adjustedContentInset` automatically as needed.
No need to add extra `contentInset` manually.

Related radar: http://www.openradar.appspot.com/35866834

Before on iOS 13:
before-fix

After on iOS 13:
after-fix

### DIFF
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -401,12 +401,17 @@ extension ScrollView
             finalContentInset.bottom -= safeAreaInsets.bottom
         }
         
-        // The refresh control lives above the content and adjusts the
-        // content inset for itself when visible and refreshing.
-        // Do the same adjustment to our expected content inset.
-        
-        if case .refreshing = refreshControlState {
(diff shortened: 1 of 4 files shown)


kylebshr (approved):
Thanks for the patch! Can you update the CHANGELOG.md?

kyleve (approved):

kyleve:
Thank you!

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -406,7 +406,13 @@ extension ScrollView
         // Do the same adjustment to our expected content inset.
         
         if case .refreshing = refreshControlState {
```
watt, line 408: Minor, it might be nice to put this line and the comments above all within the `else`

watt (approved):

NSHaoSong:
> Thanks for the patch! Can you update the CHANGELOG.md?

updated. do we also put the pr link at bottom of the change log?

Issue #49, opened 2020-01-22
Secondary elements


watt:
SwiftUI has a concept of [secondary views](https://netsplit.com/swiftui/secondary-views/), which are attached to a parent view but do not play a part in layout. Handy for e.g. putting a badge on an icon without changing the layout of the icon.

This can almost be replicated in Blueprint with Overlays, but really needs custom layout in order to let the secondary element take any size or position without affecting the parent element. Blueprint should provide a dedicated element for this pattern.

bencochran:
Could you sketch what you imagine this API looks like?

watt:
Oops, forgot to respond to this. I imagine a wrapper that simply takes two elements, and "attaches" the second to the first.

```swift
Attached(
    element: primaryElement,
    attachment: secondaryElement
)
```

Here `primaryElement` is wrapped and laid out normally, and `secondaryElement` is overlaid, but unlike `Overlay`, the layout of `secondaryElement` can not affect `primaryElement` or the `Attached` element.

PR #178, opened 2020-11-05
Introduce Decorate element

kyleve:
This PR introduces the `Decorate` element, which lets you put a decoration element above or below a content element, and position it depending on insets, corner, or custom. This is useful for drawing backgrounds behind components, drawing badges on elements, etc.


BlueprintUI/Sources/Layout/DecorateBackground.swift
```diff
@@ -0,0 +1,185 @@
...
+extension DecorateBackground {
+    fileprivate struct Layout : BlueprintUI.Layout {
+        
+        var overflow : UIEdgeInsets
+        
+        func measure(in constraint: SizeConstraint, items: [(traits: (), content: Measurable)]) -> CGSize {
+            
+            precondition(items.count == 2)
+            
+            let wrapped = items[1]
```
kyleve, line 158: Is there a better way to pass this through?

BlueprintUI/Sources/Layout/DecorateBackground.swift
```diff
@@ -0,0 +1,185 @@
...
+    /// │ ┗━━━━━━━━━━━━━━━┛ │ ▼
+    /// └───────────────────┘
+    ///   ◀───────────────▶
+    /// ```
+    public var background : Element
+    
+    /// How much the background should overflow the measured bounds of the
+    /// element. Positive values overflow outside of the bounds, and negative
+    /// values underflow to inside the bounds.
+    public var overflow : UIEdgeInsets
```
kylebshr, line 63: I feel like we might want to make this an enum? insets might not work for every layout you'd want to do with a background - i.e., what if I want a fixed size decoration, positioned on the top right corner behind the wrapped element? Not saying we need to add those all now, but making this an enum now means easier migration

BlueprintUI/Sources/Layout/DecorateBackground.swift
```diff
@@ -0,0 +1,185 @@
...
+    
+    /// Creates a new instance with the provided overflow, background, and wrapped element.
+    public init(
+        overflow: UIEdgeInsets,
+        background: () -> Element,
```
kylebshr, line 68: `@autoclosure` maybe? Curious why you're switching up the API from what most of the elements do, especially since you're just immediately calling the func?

kylebshr (approved):
Looks good for the most part, not sure about the `Layout` question 🤔 we should add some snapshot tests too!

watt:
> This is useful for rendering tap or selection states which should overflow the natural bounds of the Element, similar to a shadow.

Is anyone using an element like this now?

I think in most elements I've seen where the tap target is larger than its visible affordance, it lays out the background normally, and then adjusts the tap target in the backing control. This seems to invert that idea, making the layout size match the tap target, and then taking the background out of layout. I'd like to see how this is expected to be used.

BlueprintUI/Sources/Layout/DecorateBackground.swift
kyleve: Actually; this gives me a better idea. Make this be a `Decoration` element, and you can put the decoration in front of or behind the content, and then also as you describe, more carefully control the layout. This is basically what is described in https://github.com/square/Blueprint/issues/49.

kylebshr: I think that makes a lot of sense

kyleve:
Redoing this ontop of `LayoutWriter` from https://github.com/square/Blueprint/pull/187, ignore commits until I update docs etc.

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,216 @@
...
+    }
+    
+    public enum Corner : Equatable {
+        case topLeft
+        case topRight
+        case bottomRight
+        case bottomLeft
+    }
+    
+    public enum Position {
```
kyleve, line 128: Test

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,216 @@
...
+//  BlueprintUI
+//
+//  Created by Kyle Van Essen on 11/4/20.
+//
+
+import UIKit
+
+
+///
+/// Places a decoration element behind or in front of the given `wrapped` element,
```
kyleve, line 12: Needs updated docs

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,216 @@
...
+                layout.add(with: wrappedFrame, child: self.wrapped)
+            }
+        }
+    }
+}
+
+
+extension Decorate {
+    
+    public enum Layering : Equatable {
```
kyleve, line 116: Docs

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,216 @@
...
+        case above
+        case below
+    }
+    
+    public enum Corner : Equatable {
```
kyleve, line 121: Docs

BlueprintUI/Sources/Layout/Decorate.swift
kyleve, line 128: Docs

BlueprintUI/Sources/Layout/DecorateBackground.swift
kyleve: Done

kyleve:
This should be good for another review pass.

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,236 @@
...
+
+
+extension Element {
+    
+    /// Places a decoration element behind or in front of the given `wrapped` element,
+    /// and positions it according to the `position` parameter.
+    ///
+    /// See the `Decorate` element for more.
+    ///
+    public func decorate(
```
watt, line 211: Can we expose this with [background] and [overlay] modifiers instead (maybe renaming the latter to disambiguate from the `Overlay` element)?

[background]: https://developer.apple.com/documentation/swiftui/view/background(_:alignment:)
[overlay]: https://developer.apple.com/documentation/swiftui/view/overlay(_:alignment:)

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,236 @@
...
+/// │    Decoration     │     │      │
+/// │ ┏━━━━━━━━━━━━━━━┓ │ ▲   │      ┣━━━━━━━━━━┓   ▲
+/// │ ┃               ┃ │ │   └─┳────┘          ┃   │
+/// │ ┃    Wrapped    ┃ │ │     ┃    Wrapped    ┃   │
+/// │ ┃               ┃ │ │     ┃               ┃   │
+/// │ ┗━━━━━━━━━━━━━━━┛ │ ▼     ┗━━━━━━━━━━━━━━━┛   ▼
+/// └───────────────────┘
+///   ◀───────────────▶         ◀───────────────▶
+/// ```
+public struct Decorate : ProxyElement {
```
watt, line 32: This looks really close to an implementation of #49, with just a few API tweaks!

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,236 @@
...
+        
+        /// The decoration is displayed above the content element.
+        case above
+        
+        /// The decoration is displayed below the content element.
+        case below
+    }
+    
+    /// What corner the decoration element should be positioned in.
+    public enum Corner : Equatable {
```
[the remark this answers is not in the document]
kyleve: I'm not sure this is worth it to lose the custom frame case – providing an "escape hatch" for people still seems valuable so we're not on the hook for implementing every needed case within this type.

BlueprintUI/Sources/Layout/Decorate.swift
kyleve: I went with this because of potential confusion – eg background sounds like I'm giving a background to my current element, and overlay is confused with `Overlay` as you mention.

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -0,0 +1,236 @@
...
+    public var layering : Layering
+    
+    /// How the `decoration` should be positioned in respect to the `wrapped` element.
+    public var position : Position
+    
+    /// Creates a new instance with the provided overflow, background, and wrapped element.
+    public init(
+        layering : Layering,
+        position: Position,
+        wrapping: () -> Element,
```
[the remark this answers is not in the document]
kyleve: I like the closure since it gives you an immediate scope to set things up, but I'll remove for consistency for now.

BlueprintUI/Sources/Layout/Decorate.swift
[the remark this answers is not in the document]
kyleve: What about replacing Corner with Alignment; seems about the same but added functionality for centering on a given side?

BlueprintUI/Sources/Layout/Decorate.swift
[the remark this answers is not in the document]
kyleve: I think I'll keep it on the `Element` extension though since it makes the code read a bit more nicely, eg:

```
        .thingy()
        .decorate(layering: .below, position: .inset(5)) {
            Box(backgroundColor: .init(white: 0.0, alpha: 0.1), cornerStyle: .rounded(radius: 17))
        }
```

BlueprintUI/Sources/Layout/Decorate.swift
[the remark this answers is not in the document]
kyleve: Actually, I think to do this "right" it's gonna take unifying some types between alignment guides, the aligned type, and this. Let's go ahead and merge this as is, and then I can follow up with that in another PR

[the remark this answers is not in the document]
watt: Hmm, this is really the main sticking point for me in this PR. If we use Alignment it's almost exactly the same as SwiftUI's secondary views, but if we invent new types for this it moves us further away.

I guess I would be OK merging as-is as long as we follow up before we do another release? Alternatively, I could make a PR onto this branch with suggested changes, seeing as how I've worked with alignments already, I have a pretty good idea of the changes involved.

> What about replacing Corner with Alignment; seems about the same but added functionality for centering on a given side?

It's not. The static cases have similar names, but alignments are opaque — resolving them gives you an x/y position in the coordinate space of the parent.

> providing an "escape hatch" for people still seems valuable

I think the custom frame escape hatch is still achievable using Alignment if you wrap your content in `.constrainedSize()`.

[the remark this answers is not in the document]
kyleve: Got it – this makes sense. I'll do another PR into this one to separate the changes. I plan on keeping the outer `positioning` though; and will replace corner with alignment.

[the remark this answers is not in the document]
kyleve: So something I don't quite understand looking at the existing alignments defined in Alignments.swift – is each of them provide a different edge – eg leading provides 0, center provides the center, and trailing provides the width – but in the case of eg the Aligned element, I need to ensure the contained wrapped element just buts up against the current edge – but this edge depends on the Alignment – but since the alignments are now just types, I can't switch over them easily... I think I'd run into the same issue here... Maybe this would be better to discuss tomorrow in 1:1? I feel like we're missing something to make this work everywhere, or I'm missing something obvious

kyleve:
I'm going ahead and land this, we can update with the alignments later.

PR #180, opened 2020-11-12
Add `AccessibilityContainer.identifier`

zradke:
While accessibility containers themselves are generally not
accessibility elements, giving them an `accessibilityIdentifier` can be
useful in KIF tests to disambiguate distinct instances with similar
element subtrees. Since accessibility identifiers are not consumer
facing, this will have no impact on Voice Over.

For example, in an app that displays a collection of todo lists, two
lists may share the same todo text causing rows with the same
`accessibilityLabel`. Creating a test that could pic a specific todo
item when the text appears across multiple items would be difficult, but
by adding a unique `accessibilityIdentifier` to each todo list a test
could first search for the containing view then only search for the todo
item within that root view.

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -15,10 +15,13 @@ import UIKit
 /// and the search down that branch is also terminated.
 public struct AccessibilityContainer: Element {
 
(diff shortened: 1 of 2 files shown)


kyleve (approved):
Can you update the CHANGELOG.md in the root of the repo as well?

watt (approved):

PR #181, opened 2020-11-18
Set up Github Actions

bencochran:
Travis appears to be getting progressively slower, so let’s give Github Actions a shot. This replicates the build matrix that we had over there. I think to land this we’ll need to bypass the Travis merge requirement and then once Github registers this as a recent check, add it as a requirement.

(Also, this zips up any changed snapshot images as a build artifact. Eventually we should be able to write a script similar to what we have internal at Square to download references images for a given PR. But I’ll leave that part for another time.)


kyleve:
LGTM

kyleve (approved):

BlueprintUICommonControls/Tests/Sources/LabelTests.swift
```diff
@@ -108,7 +108,7 @@ class LabelTests: XCTestCase {
     }
 
     fileprivate func compareSnapshot(identifier: String? = nil, file: StaticString = #file, testName: String = #function, line: UInt = #line, configuration: (inout Label) -> Void) {
-        var label = Label(text: "Hello, world. This is a long run of text that should wrap at some point. Someone should improve this test by adding a joke or something. Alright, it's been fun!")
+        var label = Label(text: "Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.")
```
kyleve, line 111: Intentional?

bencochran: yeah was testing snapshot tests to make sure they built

bencochran: hence the `DRAFT` ;)

kyleve: hrYQS5h

kyleve (approved):

.travis.yml
```diff
@@ -1,30 +0,0 @@
...
-  xcodebuild -scheme "BlueprintUI-Package" test -destination "$DESTINATION"
-
-jobs:
-  include:
-  - name: "Cocoapods BlueprintUI"
-    script: xcodebuild -workspace "SampleApp.xcworkspace" -scheme "BlueprintUI-Unit-Tests" -destination "platform=iOS Simulator,OS=14.0,name=iPhone 8"
-    before_install:
-      - cd SampleApp
-      - bundle install
-      - bundle exec pod repo update
```
kylebshr, line 29: I wonder if we need to repo update in the new script as well - @watt do you know why this is here?

kylebshr (approved):
👏 brilliant, thanks for updating this!

.travis.yml
watt: > I wonder if we need to repo update in the new script as well - @watt do you know why this is here?

If we ever add any external pod dependencies this might be needed? Doesn't seem necessary now, though.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - name: Bundle Install
+      run: |
+        bundle check || bundle install --path .bundle
+
+    - name: Pod Install
+      run: |
+        bundle exec pod install --project-directory=SampleApp/
+
+    - name: Switch Xcode
+      run: sudo xcode-select -s 	/Applications/Xcode_12.app
```
watt, line 80: Should we put the Xcode version in an env variable and use the same version for both jobs?

Also, nit: there is a tab in this command for some reason.

watt (approved):

kyleve:
I'm going to land this since it greatly improves CI times – we can follow up with improvements. The hour+ CI is absolutely killing me rn

PR #182, opened 2020-11-21
Some changes to the ElementPreview type to use it within Market's preview element

kyleve:
I started writing up a `MarketXcodePreview` to let us more easily test Market elements within Xcode's previews, and I had to expose a few previously private values on `ElementPreview` to make this doable. I also added an easier way to convert `SizeConstraint` into `ConstrainedSize`.

### DIFF
--- a/BlueprintUI/Sources/Layout/ConstrainedSize.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedSize.swift
@@ -96,6 +96,26 @@ public extension Element {
     {
         ConstrainedSize(width: width, height: height, wrapping: self)
     }
+    
+    /// Constrains the measured size of the element to the provided `SizeConstraint`.
+    func constrained(to sizeConstraint : SizeConstraint) -> ConstrainedSize
+    {
+        ConstrainedSize(
+            width: {
+                switch sizeConstraint.width {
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -96,6 +96,26 @@ public extension Element {
...
+    /// Constrains the measured size of the element to the provided `SizeConstraint`.
+    func constrained(to sizeConstraint : SizeConstraint) -> ConstrainedSize
+    {
+        ConstrainedSize(
+            width: {
+                switch sizeConstraint.width {
+                case .atMost(let value): return .atMost(value)
+                case .unconstrained: return .unconstrained
+                }
+            }(),
```
watt, line 109: nit: consider an inline func for this instead of the closures

watt (approved):

BlueprintUI/Sources/Layout/ConstrainedSize.swift
kyleve: I've always preferred these sort of closures since it's less characters and nests the relevant value provider at the callsite; though if there was more than two parameters here; I'd agree

PR #189, opened 2020-12-17
Fix empty environment referencing a forever-living MeasurementCache

kyleve:
See the added comment for full details; but I noticed this bug today. TL;DR is because the `MeasurementCache` is a reference type and this was a `static let`, any added keys chill around in memory for the lifetime of the application.

### DIFF
--- a/BlueprintUI/Sources/Environment/Environment.swift
+++ b/BlueprintUI/Sources/Environment/Environment.swift
@@ -34,9 +34,18 @@ import Foundation
 ///         }
 ///     }
 public struct Environment {
+    
     /// A default "empty" environment, with no values overridden.
     /// Each key will return its default value.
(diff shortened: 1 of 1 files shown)


kylebshr (approved):
Perhaps this shouldn't be a property, but a function? Conveys the semantics better. 

`.empty` is nicer sugar though, so I'm fine leaving it.

kyleve:
In theory it doesn't really matter except for this case, which is intentionally a special case for perf, so I'd rather keep as is (measure cache is internal only).

Also goodness it drives me crazy Github doesnt have arbitrary threading...

watt:
> Also goodness it drives me crazy Github doesnt have arbitrary threading...

quote reply is helpful for this sometimes!

PR #190, opened 2020-12-18
Validate elements are structs

kyleve:
This PR introduces validations to ensure that elements are structs – since things will break in weird ways if this is not true unless you are very very careful.

### DIFF
--- a/BlueprintUI/Sources/Element/Element.swift
+++ b/BlueprintUI/Sources/Element/Element.swift
@@ -38,7 +38,7 @@ import UIKit
 /// }
 /// ```
 ///
-public protocol Element {
+public protocol Element : _Blueprint_Elements_Should_Be_Value_Types {
 
     /// Returns the content of this element.
     ///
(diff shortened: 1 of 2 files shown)


bencochran (approved):
Oh how I wish there was a compile-time way to assert this.

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -35,6 +35,8 @@ struct LayoutResultNode {
     var children: [(identifier: ElementIdentifier, node: LayoutResultNode)]
     
     init(element: Element, layoutAttributes: LayoutAttributes, content: ElementContent, environment: Environment) {
+        
+        element.assertIsValueType()
```
kylebshr, line 39: This might be silly, but I've never thought to investigate until now; how expensive is a function call? Worth if-defing here too, since this is a hot path?

kylebshr (approved):

BlueprintUI/Sources/Internal/LayoutResultNode.swift
kylebshr: Didn't profile, but google makes me think it's probably (hopefully?) optimized out

kyleve: So logging this even in debug with validations is fast enough that this is the result from logging the start and end dates is always extremely tiny, eg:

```
1.6239937394857407e-06
1.5599653124809265e-07
5.098991096019745e-08
1.2794043868780136e-07
4.807952791452408e-08
1.2293457984924316e-07
4.400499165058136e-08
6.402842700481415e-08
3.795139491558075e-08
3.8067810237407684e-08
3.702007234096527e-08
4.901085048913956e-08
3.993045538663864e-08
3.702007234096527e-08
3.702007234096527e-08
3.8067810237407684e-08
3.597233444452286e-08
4.7031790018081665e-08
5.098991096019745e-08
3.597233444452286e-08
3.899913281202316e-08
3.8067810237407684e-08
3.795139491558075e-08
```

Eg `0.000000037951`, aka like less than a ten millionth of a second... so I think we're good to go

kyleve:
Ok @bencochran saves the day once again; this is now compile time

BlueprintUI/Sources/Internal/LayoutResultNode.swift
kyleve: Nvm ignore this, ben provided a compile time way to do this

BlueprintUI/Sources/Element/Element.swift
```diff
@@ -64,3 +64,37 @@ public protocol Element {
...
+}
+
+
+public extension Blueprint_Elements_Should_Be_Value_Types {
+    func elements_should_be_value_types_by_overriding_this_method_i_acknowledge_i_am_in_hard_mode() {}
+}
+
+
+public extension Blueprint_Elements_Should_Be_Value_Types where Self : AnyObject {
+    @available(*, unavailable)
```
watt, line 98: You can add a message to this to make the error message better:

```suggestion
    @available(*, unavailable, message: "Elements should be structs")
```

watt (approved):

BlueprintUI/Sources/Element/Element.swift
```diff
@@ -64,3 +64,37 @@ public protocol Element {
...
+    func elements_should_be_value_types_by_overriding_this_method_i_acknowledge_i_am_in_hard_mode() {}
```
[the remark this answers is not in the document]
kyleve: Oh clever! Sure; I always forget that's a thing.

BlueprintUI/Sources/Element/Element.swift
kyleve: Will do! Will commit locally since I'm already adding the underscore.

PR #195, opened 2021-01-11
Add `Transformed` element

zradke:
Adds an element which can manipulate the transform of a wrapped element.


BlueprintUI/Sources/Layout/Transform.swift
```diff
@@ -0,0 +1,134 @@
...
+    private struct Layout: SingleChildLayout {
+        var transform: CATransform3D
+
+        func measure(in constraint: SizeConstraint, child: Measurable) -> CGSize {
+            return child.measure(in: constraint)
+        }
+
+        func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
+            var attributes = LayoutAttributes(size: size)
+            attributes.transform = transform
```
kyleve, line 56: Do we want to do this; or would we want to wrap in another view that has a transform attached to it? Trying to think if this would end up "stomping over" any other transforms, like one from a transition...

kyleve: (Also, maybe Andrew can check me on this, but does this get applied correctly if the wrapped element in this case is also not view-backed, but instead just has several view backed child elements?)

BlueprintUI/Sources/Layout/Transform.swift
```diff
@@ -0,0 +1,134 @@
+import UIKit
+
+/// Changes the transform of the wrapped element.
+public struct Transform: Element {
```
watt, line 4: I'm going to bikeshed just a bit and suggest `Transformed`, to match the style of Aligned, ConstrainedSize, etc.

BlueprintUI/Sources/Layout/Transform.swift
watt: This looks right to me. Attributes set during layout are merged when the tree is flattened, and transition attributes are applied on top of the resolved attributes from layout, so I don't think there's any stomping.

BlueprintUI/Sources/Layout/Transform.swift
```diff
@@ -0,0 +1,134 @@
...
+            wrapping: self
+        )
+    }
+
+    /// Wraps the element in an `Transform` element that rotates the receiver in 2D space.
+    ///
+    /// - parameters:
+    ///   - rotate: The angle measurement to rotate the receiver by.
+    ///
+    func transform(rotate: Measurement<UnitAngle>) -> Transform {
```
watt, line 110: `rotated()` maybe?

BlueprintUI/Sources/Layout/Transform.swift
```diff
@@ -0,0 +1,134 @@
...
+        )
+    }
+
+    /// Wraps the element in an `Transform` element that scales the receiver in 2D space.
+    ///
+    /// - parameters:
+    ///   - scaleX: The X axis scale.
+    ///   - scaleY: The Y axis scale.
+    ///
+    func transform(
```
watt, line 125: and `scaled()`?

BlueprintUI/Sources/Layout/Transform.swift
kyleve: +1

watt (approved):

BlueprintUI/Sources/Layout/Transformed.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        return nil
+    }
+
+    private struct Layout: SingleChildLayout {
+        var transform: CATransform3D
+
+        func measure(in constraint: SizeConstraint, child: Measurable) -> CGSize {
+            return child.measure(in: constraint)
```
kyleve, line 51: Does this measure against the transformed element? I don't think it will; I wonder if we want to apply the transform to the measured size/bounds?

kyleve: But then again, I guess that would affect the frame of reference of the layout, which isn't desirable.

kyleve (approved):
Looks good; thank you for this!

kyleve:
@zradke: I sent you an invite to be a maintainer for the repo, so going forward you'll be able to push and merge in here directly.

PR #198, opened 2021-01-21
Introduce conditionals on Element

kyleve:


mattfaluotico:
Thanks for popping this into it's own PR!

BlueprintUI/Sources/Layout/Conditionals.swift
```diff
@@ -0,0 +1,133 @@
...
+    
+    /// Returns a new element from the provided `modify`
+    /// closure if the provided value is non-nil. Otherwise, the original
+    /// element is returned.
+    /// ```
+    /// myElement.if(let: someValue) { value, element in
+    ///     element.inset(uniform: someValue.padding)
+    /// }
+    /// ```
+    public func `if`<Value>(
```
mattfaluotico, line 74: ❤️

mattfaluotico (approved):

kyleve:
Pls sir.. can I have a review

bencochran (approved):

BlueprintUI/Sources/Layout/Conditionals.swift
```diff
@@ -0,0 +1,133 @@
...
+    /// ```
+    /// myElement.map { element in
+    ///     switch myState {
+    ///     case .small: element.inset(uniform: 5)
+    ///     case .medium: element.inset(uniform: 10)
+    ///     case .large: element.inset(uniform: 15)
+    ///     }
+    /// }
+    /// ```
+    public func map(_ map : (Self) -> Element) -> Element
```
watt, line 129: Didn't we decide _not_ to call this one map, because it's not a container?

watt (approved):

BlueprintUI/Sources/Layout/Conditionals.swift
kyleve: It's been so long. I honestly forget... I figure this is most familiar to people, so, familiarity probably wins out?

kyleve: I could also see `modify`, but that doesn't quite work, since it's not an in-place transform, you're returning a new element – same with `transform`.

n8chur:
Neat!

PR #199, opened 2021-01-31
Protect LayoutWriter against empty frames

kyleve:
- Skip layout when LayoutWriter has no area; these layouts don't mean anything and are likely to cause NaN layouts due to a zero width or height.

### DIFF
--- a/BlueprintUI/Sources/Layout/LayoutWriter.swift
+++ b/BlueprintUI/Sources/Layout/LayoutWriter.swift
@@ -49,7 +49,10 @@ public struct LayoutWriter : Element {
     public var content: ElementContent {
         ElementContent { size, env in
             var builder = Builder()
-            self.build(Context(size: size), &builder)
+            
+            if size.maximum.width > 0 && size.maximum.height > 0 {
+                self.build(Context(size: size), &builder)
+            }
             
             return Content(builder: builder)
         }
--- a/BlueprintUI/Tests/LayoutWriterTests.swift
+++ b/BlueprintUI/Tests/LayoutWriterTests.swift
@@ -34,6 +34,42 @@ class LayoutWriterTests : XCTestCase {
         XCTAssertEqual(buildCount, 2)
     }
     
+    func test_zeroSize_does_no_layout() {
+        
+        // Trying to perform a layout with a zero size doesn't really mean anything.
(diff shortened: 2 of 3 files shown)


kylebshr:
I think it would be good to add a log when this happens, since it might not be intended and might help you debug why your layout isn't visible (like UIKit tells you when constraints break).

BlueprintUI/Tests/LayoutWriterTests.swift
```diff
@@ -34,6 +34,42 @@ class LayoutWriterTests : XCTestCase {
         XCTAssertEqual(buildCount, 2)
     }
     
+    func test_zeroSize_does_no_layout() {
```
kylebshr, line 37:
```suggestion
    func test_zero_size_does_no_layout() {
```
was the snakeCamel_mix unintentional?

kylebshr: `test_buildCount` is failing now btw not sure why 🤔

kylebshr (approved):

BlueprintUI/Sources/Layout/LayoutWriter.swift
```diff
@@ -49,7 +49,10 @@ public struct LayoutWriter : Element {
     public var content: ElementContent {
         ElementContent { size, env in
             var builder = Builder()
-            self.build(Context(size: size), &builder)
+            
+            if size.maximum.width > 0 && size.maximum.height > 0 {
```
watt, line 53: You should probably document this behavior. Most others containers will still include their children even if they're laying out in an empty frame. This change means a zero frame won't even create any children.

BlueprintUI/Tests/LayoutWriterTests.swift
```diff
@@ -34,6 +34,42 @@ class LayoutWriterTests : XCTestCase {
...
+        
+        // Trying to perform a layout with a zero size doesn't really mean anything.
```
watt, line 39: FWIW there are some cases where layout with a zero size is still valid:
- a spacer that is zero in 1 dimension
- children that overflow
- children that are anchored but do not affect layout, e.g. a shadow or a badge

watt (approved):

kyleve:
I'm going to decline this for now, given it can be a valid case sometimes.

PR #201, opened 2021-02-14
Add additional ways to create and mutate an overlay, eg to ease conditional construction

kyleve:


kylebshr (approved):

BlueprintUI/Sources/Layout/Overlay.swift
```diff
@@ -5,15 +5,28 @@ import UIKit
...
 public struct Overlay: Element {
 
+    /// All elements displayed in the overlay.
     public var elements: [Element]
 
-    public init(elements: [Element]) {
+    /// Creates a new overlay with the provided elements.
+    public init(
+        elements: [Element] = [],
+        configure : (inout Overlay) -> () = { _ in}
```
watt, line 18: I kinda feel like this would make more sense as two separate inits.

watt (approved):

BlueprintUI/Sources/Layout/Overlay.swift
kyleve: I thought about that, but they do get matrixed out when autocompleting (eg you get to choose from the one with or without methods), so it felt fine to keep it as one.

PR #205, opened 2021-03-24
Change backing view of `TransitionContainer` to passthrough touches

zradke:
This change modifies the backing view of the `TransitionContainer` to be
a new `PassthroughView` internal type which ignores touches that hit the
view itself but allow those touches to hit subviews for finer grained
control than `userInteractionEnabled`.

### DIFF
--- a/BlueprintUICommonControls/Sources/Internal/PassthroughView.swift
+++ b/BlueprintUICommonControls/Sources/Internal/PassthroughView.swift
@@ -1,0 +1,9 @@
+import UIKit
+
+/// View which does not personally receive touches but allows its subviews to receive touches.
+final class PassthroughView: UIView {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        let result = super.hitTest(point, with: event)
+        return result == self ? nil : result
+    }
+}
--- a/BlueprintUICommonControls/Sources/TransitionContainer.swift
+++ b/BlueprintUICommonControls/Sources/TransitionContainer.swift
@@ -59,7 +59,7 @@ public struct TransitionContainer: Element {
     }
 
     public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
-        return UIView.describe { config in
+        return PassthroughView.describe { config in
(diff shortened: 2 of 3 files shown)


BlueprintUICommonControls/Sources/Internal/PassthroughView.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+/// View which does not personally receive touches but allows its subviews to receive touches.
+class PassthroughView: UIView {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        let result = super.hitTest(point, with: event)
+        return result == self ? nil : result
```
kyleve, line 7: Clever!!

BlueprintUICommonControls/Sources/TransitionContainer.swift
```diff
@@ -59,7 +59,7 @@ public struct TransitionContainer: Element {
     }
 
     public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
-        return UIView.describe { config in
+        return PassthroughView.describe { config in
```
kyleve, line 62: Thinking out loud: Any other view-backed elements that use UIView we should use this new type for? cc @watt

BlueprintUICommonControls/Sources/Internal/PassthroughView.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+/// View which does not personally receive touches but allows its subviews to receive touches.
+class PassthroughView: UIView {
```
kyleve, line 4: Nit: Final?

kyleve (approved):

kyleve:
Thanks for doing this!

BlueprintUICommonControls/Sources/TransitionContainer.swift
watt: > Thinking out loud: Any other view-backed elements that use UIView we should use this new type for? cc @watt

Might be a good base class for the accessibility related elements that are fully transparent containers? Nothing else comes to mind.

PR #208, opened 2021-04-01
[GridRow] Introduce GridRow, a columnar layout element

nononoah:
### Summary
This PR introduces `GridRow` which is described below. 

Once a round of discussion topics are resolved and it receives approval, I plan to open a second PR with tests, changelog updates, and so on.

### Achieving consistent, adaptive column layout
`GridRow` is a new `Element` meant to help users write concise, consistent, and adaptive column-based grids.
​
### Background

The `Stack` API is not well-suited to writing `Rows` that layout data in columns. For such layouts, it is essential that sizing is
  - consistent from `Row` to `Row`
  - adaptive across various device sizes

As is, `Stack` layout convenience is oriented around sizing children based on their content size. Notably, we want to avoid that feature in this case, as it is conflict with our first requirement. 
​
Nevertheless, one can achieve the desired effect by wrapping `Rows` in `GeometryReaders`, calculating available layout width, and applying fractions of it to children wrapped in `ConstrainedSize` elements.
​
This is a rather involved, if reasonable, approach. Since this is a common UI paradigm, however, I believe we should provide a simpler path.
​
### Design
`GridRow` layout will support the following:
  - spacing
  -  vertical alignment
  - children with fixed widths
  - children with fractional widths¹

¹Fractional width in this case always means "the fraction of available layout space, after margins and fixed-width children are laid out."
​
Note the absence of horizontal distribution.
​
`GridRow` will exist separately from `Row` and `Stack`, and house its own layout algorithm. The layout algorithm will be relatively simple:
  - determine available layout width by subtracting margins and fixed-width children
  - layout children from left to right, applying the appropriate width and offset
​
### Usage examples
```swift
GridRow { row in
  row.spacing = 8
  row.verticalAlignment = .fill
  row.add(child: .absolute(author, 50))
  row.add(child: .proportional(body, 0.75))
  row.add(child: .proportional(date, 0.25))
}
```

### Demo
grid-row

### DIFF
--- a/BlueprintUI/Sources/Layout/GridRow.swift
+++ b/BlueprintUI/Sources/Layout/GridRow.swift
@@ -1,0 +1,306 @@
+import UIKit
+
+/// Like `Row`, `GridRow` displays a list of items in a linear horizontal layout. Unlike `Row`, `GridRow` provides
+/// convenience for describing columnar layout.
+///
+/// Horizontally, `GridRow` children are stretched to fill the available space. Vertically, children are aligned
+/// according to the `verticalAlignment` property.
+///
+/// Children may be sized proportionally or absolutely. Proportionally-sized children are granted a proportion of
+/// the total layout space **after** absolutely-sized children and margins have been subtracted.
+///
+/// ## Example:
+///
+/// ```
+/// GridRow { row in
+///   row.verticalAlignment = .fill
+///   row.spacing = 8.0
+///   row.add(width: .proportional(0.75), child: name)
+///   row.add(width: .proportional(0.25), child: number)
+///   row.add(width: .absolute(100), child: status)
+/// }
+/// ```
+///
+/// ## Expected layout:
+///
+/// ```
+/// ┌────────────────────────────┬─┬────────┬─┬──────────────────┐
+/// │            name            │ │ number │ │      status      │
+/// │            (75%)           │8│  (25%) │8│     (100 pts)    │
+/// │                            │ │        │ │                  │
+/// ●──────────── 150 ───────────● ●── 50 ──● ●─────── 100 ──────●
+/// └────────────────────────────┴─┴────────┴─┴──────────────────┘
+/// ●──────────────────────────── 316 ───────────────────────────●
+/// ```
+public struct GridRow: Element {
+    // MARK: - properties -
+    /// How children are aligned vertically. By default, `.fill`.
+    public var verticalAlignment: Row.RowAlignment = .fill
+
+    /// The space between children. By default, 0.
+    public var spacing: CGFloat = 0
+
+    /// The child elements to be laid out. By default, an empty array.
+    public var children: [Child] = []
+
+    // MARK: - initialization -
+    public init(configure: (inout GridRow) -> Void = { _ in }) {
+        configure(&self)
+    }
+
+    // MARK: - mutations -
+    public mutating func add(width: Width, key: AnyHashable? = nil, child: Element) {
+        children.append(Child(width: width, key: key, element: child))
+    }
+
+    // MARK: - GridRow+Element -
+    public var content: ElementContent {
+        ElementContent(layout: GridRowLayout(verticalAlignment: verticalAlignment, spacing: spacing)) { builder in
+            children.forEach {
+                builder.add(traits: $0.width, key: $0.key, element: $0.element)
+            }
+        }
+    }
+
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        nil
+    }
+}
+
+// MARK: - child modeling -
+public extension GridRow {
+    /// A child of a `GridRow`.
+    struct Child {
+        // MARK: - properties -
+        /// The element displayed in the `Grid`.
+        public var element: Element
+        /// A unique identifier for the child.
+        public var key: AnyHashable?
+        // The sizing for the element.
+        public var width: Width
+
+        // MARK: - initialialization -
+        public init(width: Width, key: AnyHashable? = nil, element: Element) {
+            self.element = element
+            self.key = key
+            self.width = width
+        }
+    }
+
+    /// The sizing and content of a `GridRow` child.
+    enum Width: Equatable {
+        /// Assign the child a fixed width equal to the payload.
+        case absolute(CGFloat)
+        /// Assign the child a proportional width of the available layout width. Note that proportional children
+        /// take proportional shares of the available layout width.
+        ///
+        /// ## Example:
+        ///     Available layout width: 100
+        ///     Child A: .proportional(1)  -> 25 (100 * 1/4)
+        ///     Child B: .proportional(3) -> 75 (100 * 3/4)
+        ///
+        /// ## Example:
+        ///     Available layout width: 100
+        ///     Child A: .proportional(0.25)  -> 25 (100 * 1/4)
+        ///     Child B: .proportional(0.75) -> 75 (100 * 3/4)
+        case proportional(CGFloat)
+    }
+}
+
+// MARK: - layout -
+extension GridRow {
+    struct GridRowLayout: Layout {
+        let verticalAlignment: Row.RowAlignment
+        let spacing: CGFloat
+
+        typealias IndexedProportionalItem = (index: Int, proportion: CGFloat, content: Measurable)
+        typealias IndexedSize = (index: Int, size: CGSize)
+
+        static var defaultTraits: GridRow.Width {
+            .absolute(0)
+        }
+
+        func measure(in constraint: SizeConstraint, items: [(traits: Width, content: Measurable)]) -> CGSize {
+            guard items.count > 0 else {
+                return .zero
+            }
+
+            let frames = _frames(in: constraint, items: items)
+
+            // Measure the the row to be as wide as the sum of its children and as tall as its tallest child.
+            let size = frames.reduce(.zero) { size, frame in
+                CGSize(width: frame.maxX, height: max(size.height, frame.height))
+            }
+
+            return size
+        }
+
+        func layout(size: CGSize, items: [(traits: Width, content: Measurable)]) -> [LayoutAttributes] {
+            guard items.count > 0 else {
+                return []
+            }
+
+            let frames = _frames(in: SizeConstraint(size), isExactConstraint: true, items: items)
+            let attributes = frames.map(LayoutAttributes.init)
+
+            return attributes
+        }
+
+        /// Compute child frames in a given layout space.
+        ///
+        /// First, we measure the size of absolutely-sized children. Next, we measure the size of
+        /// proportionally-sized children.
+        ///
+        /// When layout width is constrained, proportionally-sized children receive a portion of the available space.
+        /// When it is not, proportionally-sized children are sized relative to each other.
+        ///
+        /// Finally, we compute the frames of children within the row from left to right, assigning a y-origin
+        /// and height which fulfill the row's vertical alignment.
+        ///
+        /// During layout, the constraint is considered an exact description of the row size (`isExactConstraint`).
+        /// This shadows `Stack.VectorConstraint.exactly` conceptually without introducing the greater complexity of
+        /// this type, which would go mostly unused here.
+        private func _frames(
+            in constraint: SizeConstraint,
+            isExactConstraint: Bool = false,
+            items: [(traits: Width, content: Measurable)]
+        ) -> [CGRect] {
+            var sizes: [CGSize] = Array(repeating: .zero, count: items.count)
+
+            // Group children by their sizing. Maintain child order by also storing index.
+            var absolutelySized: [(index: Int, width: CGFloat, content: Measurable)] = []
+            var proportionallySized: [(index: Int, proportion: CGFloat, content: Measurable)] = []
+            items.enumerated().forEach { (index, item) in
+                switch item.traits {
+                case .absolute(let width):
+                    absolutelySized.append((index, width, item.content))
+                case .proportional(let proportion):
+                    proportionallySized.append((index, proportion, item.content))
+                }
+            }
+
+            // Measure absolutely-sized children.
+            absolutelySized.forEach { (index, width, content) in
+                let fixedWidthConstraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                sizes[index] = CGSize(width: width, height: content.measure(in: fixedWidthConstraint).height)
+            }
+
+            // Measure proportionally-sized children.
+            switch constraint.width {
+            case .atMost(let width):
+                let absoluteItemWidth = sizes.map { $0.width }.reduce(0, +)
+                let availableWidth = width - absoluteItemWidth - spacing * CGFloat(items.count - 1)
+                constrainedProportionalItemSizes(
+                    in: constraint,
+                    availableWidth: availableWidth,
+                    items: proportionallySized
+                )
+                .forEach { (index, size) in
+                    sizes[index] = size
+                }
+            case .unconstrained:
+                unconstrainedProportionalItemSizes(
+                    in: constraint,
+                    items: proportionallySized
+                )
+                .forEach { (index, size) in
+                    sizes[index] = size
+                }
+            }
+
+            // Begin computing frames by reading the vertical alignment. Vertical alignment determines the y-origin
+            // and, in the `.fill` case, the height of the children.
+            //
+            // The tallest child will be used to size the row unless it is a layout pass, in which the row's height
+            // is already known.
+            let rowHeight = isExactConstraint
+                ? constraint.height.maximum
+                : sizes.map { $0.height }.max() ?? 0
+
+            // Define alignment positioning.
+            let yVector = _yVector(rowHeight: rowHeight)
+
+            // Compute frames.
+            var offset: CGFloat = 0
+            return sizes.map { size in
+                defer { offset += size.width + spacing }
+                let (yOrigin, height) = yVector(size.height)
+                return CGRect(x: offset, y: yOrigin, width: size.width, height: height)
+            }
+        }
+
+        /// When the layout width is constrained, allot the children a portion of the available space.
+        ///
+        /// Determine the scale of a portion (points per portion), then use it to apply a width to each child.
+        func constrainedProportionalItemSizes(
+            in constraint: SizeConstraint,
+            availableWidth: CGFloat,
+            items: [IndexedProportionalItem]
+        ) -> [IndexedSize] {
+            guard items.count > 0 else {
+                return []
+            }
+
+            guard availableWidth > 0 else {
+                // There's no room to layout so there's no need to perform any measurement.
+                return items.map { (index, _, _) in (index, .zero) }
+            }
+
+            let portionSum = items.map { $0.proportion }.reduce(0, +)
+            precondition(portionSum > 0, "Proportions of a GridRow must sum to a positive number. Found sum: \(portionSum).")
+            let scale = availableWidth / portionSum
+
+            return items.map { (index, proportion, content) in
+                let width = scale * proportion
+                let fixedWidthConstraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                let size = CGSize(width: width, height: content.measure(in: fixedWidthConstraint).height)
+                return (index, size)
+            }
+        }
+
+        /// When the layout width is unconstrained, size children relative to each other.
+        ///
+        /// Measure each child's width to find the maximum scale (points per portion). The maximum scale
+        /// can be applied to each child to provide widths which satisfy the proportional requirements
+        /// between children and provide enough space to layout each child.
+        ///
+        /// ## Example:
+        ///      - proportions:      1,     2,     3
+        ///      - measured widths:  10,    25,    24
+        ///      - scales:           10,    12.5,  8
+        ///      - max scale:        12.5
+        ///      - widths:           12.5,  25,    37.5
+        private func unconstrainedProportionalItemSizes(
+            in constraint: SizeConstraint,
+            items: [IndexedProportionalItem]
+        ) -> [IndexedSize] {
+            var scale: CGFloat = 0
+            items.forEach { (_, proportion, content) in
+                let width = content.measure(in: constraint).width
+                scale = max(scale, width / proportion)
+            }
+
+            return items.map { (index, proportion, content) in
+                let width = scale * proportion
+                let fixedWidthConstraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                let size = CGSize(width: width, height: content.measure(in: fixedWidthConstraint).height)
+                return (index, size)
+            }
+        }
+
+        private func _yVector(rowHeight: CGFloat) -> (CGFloat) -> (origin: CGFloat, height: CGFloat) {
+            switch verticalAlignment {
+            case .fill:
+                return { _ in (0, rowHeight) }
+            case .align(let id) where id == .top:
+                return { height in (0, height) }
+            case .align(let id) where id == .center:
+                return { height in ((rowHeight - height) / 2.0, height) }
+            case .align(let id) where id == .bottom:
+                return { height in (rowHeight - height,height) }
+            case .align:
+                fatalError("GridRow supports fill, top, center, and bottom alignment.")
+            }
+        }
+    }
+}
--- a/BlueprintUI/Tests/GridRowTests.swift
+++ b/BlueprintUI/Tests/GridRowTests.swift
@@ -1,0 +1,282 @@
+import XCTest
+@testable import BlueprintUI
+
+class GridRowTests: XCTestCase {
+    func test_defaults() {
+        let gridRow = GridRow()
+        XCTAssertTrue(gridRow.children.isEmpty)
+        XCTAssertEqual(gridRow.verticalAlignment, .fill)
+        XCTAssertEqual(gridRow.spacing, 0)
+    }
+
+    func test_measure_unconstrained() {
+        do {
+            // #1: Proportional child measurment
+            //  proportions:      1,     2
+            //  measured widths:  10,    25
+            //  scales:           10,    12.5
+            //  max scale:        12.5
+            //  width:            (1 * 12.5) + (2 * 12.5) = 37.5
+            //  height:           max(5, 10) = 10
+            let gridRow = GridRow { row in
+                row.add(width: .proportional(1), child: TestElement(size: CGSize(width: 10, height: 5)))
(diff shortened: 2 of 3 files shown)


BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,191 @@
...
+///   row.spacing = 8.0
+///   row.add(child: .proportional(name, 0.75))
+///   row.add(child: .proportional(number, 0.25))
+///   row.add(child: .absolute(status, 50))
+/// }
+/// ```
+///
+/// ## Expected layout:
+///
+/// ┌────────────────────────────┬─┬────────┬─┬──────────────────┐
```
kyleve, line 25: Fwiw dropping this into a ``` block will make it render correctly in Xcode's docs preview

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,191 @@
...
+public struct GridRow: Element {
+    // MARK: - properties -
+    /// The space between children. By default, 0.
+    public var spacing: CGFloat = 0
+
+    /// The child elements to be laid out. By default, an empty array.
+    public var children: [Child] = []
+
+    // MARK: - initialization -
+    public init(configure: (inout GridRow) -> Void) {
```
kyleve, line 42: Nit: `= { _ in }` in case someone doesn't want to configure inline for some reason.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,191 @@
...
+            let portionCount = proportionallySized.map { $0.proportion }.reduce(0, +)
+            let portion = availableWidth / portionCount
+            // Assign a proportional width to each proportionally-sized child.
+            proportionallySized.forEach { (index, proportion, content) in
+                let width = portion * proportion
+                let height = content.measure(in: SizeConstraint(width: width)).height
+                sizes[index] = CGSize(width: width, height: height)
+            }
+
+            // Use the tallest child to set the height of all children.
```
kyleve, line 182: Would it make sense to allow a way to align the content top/bottom/centered here?

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // MARK: - mutations -
+    public mutating func add(child: Child) {
+        children.append(child)
+    }
+
+    // MARK: - GridRow+Element -
+    public var content: ElementContent {
+        ElementContent(layout: GridRowLayout(spacing: spacing)) { builder in
```
kyleve, line 53: If you wanted to, you could also use a `LayoutWriter` here to not need a separate layout – up to you! Just a bit less boilerplate w/ a LayoutWriter.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: Oh really? I actually found the existing doesn't render properly because it's not in a block. If I update it properly, I see the following:

Screen Shot 2021-04-01 at 2 35 09 PM

nononoah: Ahh, right! I somehow deleted the top ticks. Thanks for the catch!

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: I'm open to it. As is, one would have to wrap children in `Alignment` elements to achieve this effect, which isn't particularly nice.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: Interesting, let me look into this.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: Fixed.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: Fixed.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: I added this feature in the following commit. The additional layout complexity is actually not too bad:

```swift
// Begin computing frames by reading the vertical alignment. Vertical alignment determines the y-origin
// and, in the `.fill` case, the height of the children.
// The tallest child will be used to size the row.
let rowHeight = sizes.lazy.map { $0.height }.max() ?? 0
let yVector: (CGSize) -> (origin: CGFloat, height: CGFloat)
switch verticalAlignment {
case .fill:
    // Use the tallest child to set the height of all children.
    yVector = { _ in (origin: 0, height: rowHeight) }
case .top:
    yVector = { size in (origin: 0, height: size.height) }
case .center:
    yVector = { size in (origin: (rowHeight - size.height) / 2.0, height: size.height) }
case .bottom:
    yVector = { size in (origin: rowHeight - size.height, height: size.height) }
}
```

Basically, we need to update `yOrigin` and `height` based on alignment. Additionally, we now need to read all frames to determine row height.

I used `lazy` because I hoped it would lead to smaller allocations, but I'm happy to cut it out.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,231 @@
...
+/// ## Expected layout:
+///
+/// ```
+/// ┌────────────────────────────┬─┬────────┬─┬──────────────────┐
+/// │            name            │ │ number │ │      status      │
+/// │            (75%)           │ │  (25%) │ │     (50 pts)     │
+/// │                            │ │        │ │                  │
+/// ●──────────── 75  ───────────● ●── 25 ──● ●─────── 50 ───────●
+/// └────────────────────────────┴─┴────────┴─┴──────────────────┘
+/// ●──────────────────────────── 166 ───────────────────────────●
```
n8chur, line 33: Should this value be 150?

n8chur: It might be helpful to use a value where the proportional parts don't add up to 100 pts to make it more clear how space is distributed.

nononoah: It's subtle, but the extra 16 is attributed to spacing. I'll refine this diagram for clarity.

n8chur: Ah, missed the spacing—makes sense!

nononoah: Cleaned this up a bit by adding numbers to better illuminate spacing and doubling the other figures used.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+
+        // MARK: - initialialization -
+        public init(element: Element, key: AnyHashable? = nil, width: Width) {
+            self.element = element
+            self.key = key
+            self.width = width
+        }
+
+        // MARK: - convenience initialization -
+        public static func absolute(_ element: Element, _ width: CGFloat) -> Child {
```
kyleve, line 103: Nit: I would make the `element` the last parameter here, so the syntax of creating elements inline is cleaner. This is generally the style we try to follow in blueprint. Eg:

```
.absolute(50, Label(text: "...") {
   $0.font = ...
   $0.alignment = ...
})
```

Vs

```
.absolute( Label(text: "...") {
   $0.font = ...
   $0.alignment = ...
}, 50)
```

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+    enum Width: Equatable {
+        /// Assign the child a fixed width equal to the payload.
+        case absolute(CGFloat)
+        /// Assign the child a proportional width of the available layout width. Note that proportional children
+        /// take proportional shares of the available layout width.
+        ///
+        /// ## Example:
+        ///     Layout width: 100
+        ///     Child A: .proportional(1)  -> 25 (100 * 1/4)
+        ///     Child B: .proportaional(3) -> 75 (100 * 3/4)
```
kyleve, line 122: proportaional -> proportional

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+        ///
+        /// ## Example:
+        ///     Layout width: 100
+        ///     Child A: .proportional(0.25)  -> 25 (100 * 1/4)
+        ///     Child B: .proportaional(0.75) -> 75 (100 * 3/4)
```
kyleve, line 127: Same spelling error here

kyleve (approved):

kyleve:
(Also, don't forget to update the CHANGELOG.md!)

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: Great point. I'll update, thanks.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: :bless:

nononoah:
> (Also, don't forget to update the CHANGELOG.md!)

For review clarity, I intended to open a  second PR with CHANGELOG updates, tests, and potentially a demo, then merge them down when we're good to go.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+                return []
+            }
+
+            return _frames(in: SizeConstraint(size), isExactConstraint: true, items: items)
+                .map(LayoutAttributes.init)
+        }
+
+        /// Compute child frames in a given layout space.
+        ///
+        /// First, we measure the size of absolutely-sized  children. Next, we measure the size of
```
nononoah, line 164:
```suggestion
        /// First, we measure the size of absolutely-sized children. Next, we measure the size of
```

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+                sizes[index] = CGSize(width: width, height: content.measure(in: constraint).height)
+            }
+
+            // Begin computing frames by reading the vertical alignment. Vertical alignment determines the y-origin
+            // and, in the `.fill` case, the height of the children.
+            //
+            // The tallest child will be used to size the row unless it is a layout pass, in which the row's height
+            // is already known.
+            let rowHeight = isExactConstraint
+                ? constraint.height.maximum
```
watt, line 221: Again, this is not going to behave well in unconstrained measurement, and probably won't play nicely if it's nested in other containers either (like if it's in a stack with other content).

In general I would recommend avoiding the `maximum` property except for some special uses, like comparing to a range.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+        ///   - constraint: A constraint describing the available layout space.
+        ///   - isExactConstraint: Whether or not the constraint describes the row's size exactly. During measurement,
+        ///   it does not. During layout, it does.
+        ///   - items: Children to layout.
+        private func _frames(
+            in constraint: SizeConstraint,
+            isExactConstraint: Bool = false,
+            items: [(traits: Width, content: Measurable)]
+        ) -> [CGRect] {
+            var sizes: [CGSize] = (0..<items.count).map { _ in .zero }
```
watt, line 181: `Array(repeating: CGSize.zero, count: items.count)` _might_ be slightly better for this?

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+            .absolute(0)
+        }
+
+        func measure(in constraint: SizeConstraint, items: [(traits: Width, content: Measurable)]) -> CGSize {
+            guard items.count > 0 else {
+                return .zero
+            }
+
+            // Measure the the row to fit its tallest child.
+            let height = _frames(in: constraint, items: items).lazy.map { $0.height }.max() ?? 0
```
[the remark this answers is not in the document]
nononoah: Correct, that is the sole advantage.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+        /// and height which fulfill the row's vertical alignment.
+        ///
+        /// - Parameters:
+        ///   - constraint: A constraint describing the available layout space.
+        ///   - isExactConstraint: Whether or not the constraint describes the row's size exactly. During measurement,
+        ///   it does not. During layout, it does.
+        ///   - items: Children to layout.
+        private func _frames(
+            in constraint: SizeConstraint,
+            isExactConstraint: Bool = false,
```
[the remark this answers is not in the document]
nononoah: This is in spirit identical, but meant to avoid the additional type surface area, as it seems to be the only additional bit of information worth encoding.

That said, perhaps I should just use the existing type. Let me look into it.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: I think this should be safe to do in the `.exactly` case, which is what I'm getting at here.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+                absoluteItemWidth += width
+                let constraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                sizes[index] = CGSize(width: width, height: content.measure(in: constraint).height)
+            }
+
+            // Measure proportionally-sized children, allotting them a portion of the available space.
+            let availableWidth = constraint.width.maximum - (spacing * CGFloat(items.count - 1)) - absoluteItemWidth
+            // Determine the number of portions and the width of each portion.
+            let portionCount = proportionallySized.lazy.map { $0.proportion }.reduce(0, +)
+            let portion = availableWidth / portionCount
```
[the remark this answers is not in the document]
nononoah: Great catch, thought it really would be quite odd to use this type without at least one `.proportional` child. 

I'll make sure this is fixed after I update the layout approach.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+            // Measure absolutely-sized children and record the total width they require.
+            var absoluteItemWidth: CGFloat = 0
+            absolutelySized.forEach { (index, width, content) in
+                absoluteItemWidth += width
+                let constraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                sizes[index] = CGSize(width: width, height: content.measure(in: constraint).height)
+            }
+
+            // Measure proportionally-sized children, allotting them a portion of the available space.
+            let availableWidth = constraint.width.maximum - (spacing * CGFloat(items.count - 1)) - absoluteItemWidth
```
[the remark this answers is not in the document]
nononoah: This makes a lot of sense to me, though similarly to what follows below, we want to make sure we account for the `.exactly` case, where it is true that we can use the given value.

Because I was only testing in vertical containers (`ScrollView` and `Column` mostly), I missed the potential flaws in this shortcut.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+
+        func measure(in constraint: SizeConstraint, items: [(traits: Width, content: Measurable)]) -> CGSize {
+            guard items.count > 0 else {
+                return .zero
+            }
+
+            // Measure the the row to fit its tallest child.
+            let height = _frames(in: constraint, items: items).lazy.map { $0.height }.max() ?? 0
+
+            return CGSize(width: constraint.width.maximum, height: height)
```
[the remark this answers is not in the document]
nononoah: Got it. I'll update this to reduce frames into a given size, similar to how `Stack` reduces vectors into the same.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    public func backingViewDescription(bounds: CGRect, subtreeExtent: CGRect?) -> ViewDescription? {
+        nil
+    }
+}
+
+// MARK: - child modeling -
+public extension GridRow {
+    enum VerticalAlignment: Equatable {
```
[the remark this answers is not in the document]
nononoah: I'll look into reuse here. Thanks for the tip.

BlueprintUI/Sources/Layout/GridRow.swift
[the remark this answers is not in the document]
nononoah: I followed this approach, and ended up with the following:

```
var scale: CGFloat = 0
items.forEach { (_, proportion, content) in
    let width = content.measure(in: constraint).width
    scale = max(scale, width / proportion)
}

items.forEach { (index, proportion, content) in
    let width = scale * proportion
    let constraint = SizeConstraint(width: .atMost(width), height: constraint.height)
    sizes[index] = CGSize(width: width, height: content.measure(in: constraint).height)
}
```

I'll work on making this as legible as possible. Thanks for the guidance.

BlueprintUI/Sources/Layout/GridRow.swift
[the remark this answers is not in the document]
nononoah: While I didn't end up reusing the vertical alignment code, I was able to conserve this type without much churn.

BlueprintUI/Sources/Layout/GridRow.swift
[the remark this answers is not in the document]
nononoah: To make full use of `VectorConstraint`, I'd have to open it (and a few related extensions) up a bit. After exploring it, I decided to stick to our own type, which is scoped down to exactly the concerns we have. This type, of course, is nothing more than a `SizeConstraint` and a `Bool` at heart, but it does make the layout code marginally more legible.

Please let me know what you think!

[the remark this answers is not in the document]
nononoah: After playing around a bit with a few approaches, my impression is that this adds a fair bit of mostly unused complexity to `GridRow`.

That said, if we assume the average reader to already be familiar with Row, it's perhaps not a significant burden to add.

i.e.

```swift
- SizeConstraint(width: .atMost(width), height: constraint.height)
+ SizeConstraint(width: .atMost(width), height: constraint.cross.sizeConstraint)
```

```swift
- let width = content.measure(in: constraint).width
+ let width = content.measure(in: constraint.constraint(axis: .horizontal)).width
```

```swift
- let rowHeight = isExactConstraint 
-      ? constraint.height.maximum 
-      : sizes.lazy.map { $0.height }.max() ?? 0
+ let rowHeight: CGFloat
+ switch constraint.cross {
+ case .exactly(height): 
+     rowHeight = height
+ case .atMost, .unconstrained:
+    rowHeight = sizes.lazy.map { $0.height }.max() ?? 0
}
```

nononoah:
@watt, I took a shot at addressing all of your concerns here, and left comments where I didn't quite follow your lead. I'd appreciate another look when you have time.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,303 @@
...
+    /// The child elements to be laid out. By default, an empty array.
+    public var children: [Child] = []
+
+    // MARK: - initialization -
+    public init(configure: (inout GridRow) -> Void = { _ in }) {
+        configure(&self)
+    }
+
+    // MARK: - mutations -
+    public mutating func add(child: Child) {
```
n8chur, line 52: Thought: I think this is probably fine (or even perhaps preferable if we were consistent about the approach), but should the values that make up `Child` be provided as separate parameters here? Maybe just as a separate convenience func?

Alternatively, should we migrate `Stack` to use this style?

No need to take action, but it might be worth driving towards a consistent API overall if there's no technical reason to stray.

nononoah: It's a great question. The main difference, I think, is that `GridRow` uses a `struct`, while `Row` uses a tuple, which makes the spelling of these sites a bit awkward. What's more, `Row` can provide reasonable defaults for `grow` and `shrink` priority, but I can't.

I will replace the existing sugar with a style more in keeping with `Row`. Does something like 

```swift
public mutating func addAbsolute(child: Element, width: CGFloat) { ... }
public mutating func addProportional(child: Element, proportion: CGFloat) { ... }
```

roughly match your expectations?

n8chur: Ah, the lack of reasonable defaults is an interesting point.

I was thinking:
```swift
public mutating func add(width: Width, key: AnyHashable? = nil, child: Element) {
    add(child: .init(width: width, key: key, element: child)
}
```
but what you have above seems fine as well.

Curious what the other Blueprint folks think, but it's probably fine to land as is and provide additional conveniences if they seem helpful 🤷

n8chur (approved):

BlueprintUI/Sources/Layout/GridRow.swift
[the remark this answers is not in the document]
watt: OK, thanks for considering. Sometimes I find that modeling that way can make things easier to understand (see my misunderstanding of the `rowHeight` ternary below), but if it's less cumbersome to write this way, that's fine! We can ensure nothing unexpected happens with adequate test coverage.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,303 @@
...
+            case .fill:
+                return { _ in (0, rowHeight) }
+            case .align(let id) where id == .top:
+                return { height in (0, height) }
+            case .align(let id) where id == .center:
+                return { height in ((rowHeight - height) / 2.0, height) }
+            case .align(let id) where id == .bottom:
+                return { height in (rowHeight - height,height) }
+            case .align:
+                fatalError("GridRow supports fill, top, center, and bottom alignment.")
```
watt, line 299: Hmm, I guess a follow-up for me would be to extract the cross-axis alignment logic from stacks so this is easier to implement.

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -0,0 +1,303 @@
...
+            var scale: CGFloat = 0
+            items.forEach { (_, proportion, content) in
+                let width = content.measure(in: constraint).width
+                scale = max(scale, width / proportion)
+            }
+
+            return items.map { (index, proportion, content) in
+                let width = scale * proportion
+                let constraint = SizeConstraint(width: .atMost(width), height: constraint.height)
+                let size = CGSize(width: width, height: content.measure(in: constraint).height)
```
watt, line 283: Can we skip this second measurement and use the height from the first?

watt (approved):

BlueprintUI/Sources/Layout/GridRow.swift
watt: > ```swift
> public mutating func add(width: Width, key: AnyHashable? = nil, child: Element) {
>     add(child: .init(width: width, key: key, element: child)
> }
> ```
> 
> Curious what the other Blueprint folks think, but it's probably fine to land as is and provide additional conveniences if they seem helpful 🤷

This is probably the closer to the API I would build personally, but I don't have strong feelings about it. Both of them can support new params (e.g. alignment guides) or convenience methods.

nononoah: Thanks for the suggestion. I'll update before merging.

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: The overloading of the `constraint` here makes this a bit less legible, so I intend to fix it.

The first measurement measures the element against an unconstrained constraint. The second measurement occurs against a width-constrained constrained.

Because the height of some elements might grow when the width is constrained, I think we need this second measurement. Please correct me if I've misunderstood!

BlueprintUI/Sources/Layout/GridRow.swift
nononoah: I ended up preferring the organization of a separate type f(or mostly aesthetic reasons) given the relative complexity of the layout here. I'm happy to revisit.

BlueprintUI/Sources/Layout/GridRow.swift
watt: > The first measurement occurs against an unconstrained constraint. The second measurement occurs against a width-constrained constrained.
> 
> Because the height of some elements might grow when the width is constrained, I think we need this second measurement. Please correct me if I've misunderstood!

If I'm understanding right, the width constraint used for the second measurement is guaranteed to be at least as wide as the size that the element reported in the first measurement. So while an element _could_ report a different size under that constraint, I think it would be unusual. Do you have a use case in mind where this would return a different size? I'm not sure it's a behavior we need to support.

(Having just done a performance pass I'm a bit wary of measurement calls.)

nononoah: The use case I had in mind was a label. 

Given an unconstrained width (the first measurement), it would have a height as tall as a single line. 

Given a constrained width (the second measurement), the height will in some cases be taller.

> the width constraint used for the second measurement is guaranteed to be at least as wide as the size that the element reported in the first measurement

I think that is correct, but I'm not sure it accounts for the height concern.

I also want to avoid unnecessary measurement and did try to reason through each call to see if it could be skipped. Still, I'm a novice, and my naivety is worth testing. Please let me know if you think my case is flawed.

kyleve: This seems correct – if the width deviates at all from previous measurements, you will need to re-measure, because things like labels will re-flow.

watt: OK, here's an example with a label that might help.

```swift
let label = UILabel()
label.text = "Hello world"
```

Let's measure the label with no constraint. Here we're asking the label: given an infinite amount of space, what is the size you want to be?

```swift
let unconstrainedSize = label.sizeThatFits(.zero)
print(unconstrainedSize) // (85.5, 20.5)
```

Now let's take that size, and use it as a constraint:

```swift
let constrainedSize = label.sizeThatFits(unconstrainedSize)
print(constrainedSize) // (85.5, 20.5)
```

The label reports the same size. This is because the constraint is large enough that the label does not need to reflow. We know it is large enough, because the label has already told us that this is the size it wants to be. For any well-behaved element, this should hold true.

nononoah: Got it. Apologies for being a bit dense here, this makes sense. I'll open a PR to fix it.

PR #209, opened 2021-04-02
Measurement caching and performance improvements

### LINKED ISSUE #173: Profiling Signposts
We should add [profiling signposts](https://developer.apple.com/documentation/os/logging/recording_performance_data) to the render pass to aid in performance optimization.

watt:
This PR adds a couple of performance-related features. Happy to split this up into multiple PRs if folks prefer that.

# Signpost Logging

Signpost intervals are generated for:
1. Layout
2. View Updates
3. Individual element measurements

The first two are nice to get a high level view of rendering, and the third one provides a pretty nice visualization of the way elements are recursively measured.

Screen Shot 2021-04-01 at 6 45 40 PM

This fixes #173.

# Layout pass measurement caching

This is a new caching mechanism that caches every measured size during a layout pass. The cache structure mirrors the element tree itself, so it is able to cache measurements for every element, without any requirements on the element itself.

This cache's lifetime is only a single layout pass. Because of the way Blueprint repeatedly measures each subtree as it lays out each element, this prevents a lot of duplicate work from re-measuring, especially in deep trees. It does not cache the layout results returned by `performLayout`, because that is only called once per element per layout pass.

The cache is also used during measurement outside of layout (such as from `sizeThatFits` or a direct call to `element.content.measure()`), but it's less useful in that scenario since most elements do not re-measure children.

# Shortcuts

1. When laying out stacks, if the alignment is `fill` and the constraint is `exact`, we can skip the entire second measurement for cross axis sizes.
2. We can skip measurement in `ConstrainedSize` if both axes are `absolute` constraints.

# Potential follow-up work

1. We can't persist this new cache if the element changes, but we might be able to persist it between layout passes on the same element. If we make that change, we should persist the values returned by `performLayout` too.
2. We should consider merging the previous `MeasurementCache` functionality into this one, or removing it.
3. Additional logging during view updates, to visualize view churn.

# Todo

- [x] Tests
- [x] Update changelog
- [x] Move signpost logging into a helper
- [x] doc SignpostToken

### DIFF
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -1,5 +1,4 @@
 import UIKit
-import os.log
 
 /// A view that is responsible for displaying an `Element` hierarchy.
 ///
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -44,6 +44,8 @@ public final class BlueprintView: UIView {
             if oldValue == nil && self.element == nil {
                 return
             }
+
+            logElementAssigned()
             
             setNeedsViewHierarchyUpdate()
             invalidateIntrinsicContentSize()
@@ -125,7 +127,8 @@ public final class BlueprintView: UIView {
         
         return element.content.measure(
             in: measurementConstraint(with: size),
-            environment: self.makeEnvironment()
+            environment: self.makeEnvironment(),
+            cache: CacheFactory.makeCache(name: "sizeThatFits:\(type(of: element))")
         )
     }
 
@@ -151,7 +154,8 @@ public final class BlueprintView: UIView {
         
         return element.content.measure(
             in: constraint,
-            environment: self.makeEnvironment()
+            environment: self.makeEnvironment(),
+            cache: CacheFactory.makeCache(name: "intrinsicContentSize:\(type(of: element))")
         )
     }
 
@@ -473,6 +477,19 @@ extension BlueprintView {
                 log: .blueprint,
                 name: "View Update",
                 signpostID: OSSignpostID(log: .blueprint, object: self)
+            )
+        }
+    }
+
+    private func logElementAssigned() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
+                .event,
+                log: .blueprint,
+                name: "Element assigned",
+                signpostID: OSSignpostID(log: .blueprint, object: self),
+                "Element assigned to %{public}s",
+                name ?? "BlueprintView"
             )
         }
     }
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -52,7 +52,7 @@ public final class BlueprintView: UIView {
         }
     }
 
-    /// A name to help identify this view when profiling or debugging
+    /// An optional name to help identify this view
     public var name: String?
     
     /// Provides performance metrics about the duration of layouts, updates, etc.
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -1,4 +1,5 @@
 import UIKit
+import os.log
 
 /// A view that is responsible for displaying an `Element` hierarchy.
 ///
@@ -48,6 +49,9 @@ public final class BlueprintView: UIView {
             invalidateIntrinsicContentSize()
         }
     }
+
+    /// A name to help identify this view when profiling or debugging
+    public var name: String?
     
     /// Provides performance metrics about the duration of layouts, updates, etc.
     public weak var metricsDelegate : BlueprintViewMetricsDelegate? = nil
@@ -197,6 +201,7 @@ public final class BlueprintView: UIView {
         lastViewHierarchyUpdateBounds = bounds
         
         let start = Date()
+        logLayoutStart()
 
         /// Grab view descriptions
         let viewNodes = element?
@@ -204,7 +209,8 @@ public final class BlueprintView: UIView {
             .resolve() ?? []
         
         let measurementEndDate = Date()
-        
+        logLayoutEnd()
+
         rootController.view.frame = bounds
         
         var rootNode = NativeViewNode(
@@ -215,9 +221,12 @@ public final class BlueprintView: UIView {
 
         let scale = window?.screen.scale ?? UIScreen.main.scale
         rootNode.round(from: .zero, correction: .zero, scale: scale)
-        
+
+        logViewUpdateStart()
+
         rootController.update(node: rootNode, appearanceTransitionsEnabled: hasUpdatedViewHierarchy)
-        
+
+        logViewUpdateEnd()
         let viewUpdateEndDate = Date()
         
         hasUpdatedViewHierarchy = true
@@ -419,3 +428,52 @@ extension BlueprintView {
     }
 }
 
+extension BlueprintView {
+    private func logLayoutStart() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
+                .begin,
+                log: .blueprint,
+                name: "Layout",
+                signpostID: OSSignpostID(log: .blueprint, object: self),
+                "%{public}s",
+                name ?? "BlueprintView"
+            )
+        }
+    }
+
+    private func logLayoutEnd() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
+                .end,
+                log: .blueprint,
+                name: "Layout",
+                signpostID: OSSignpostID(log: .blueprint, object: self)
+            )
+        }
+    }
+
+    private func logViewUpdateStart() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
+                .begin,
+                log: .blueprint,
+                name: "View Update",
+                signpostID: OSSignpostID(log: .blueprint, object: self),
+                "%{public}s",
+                name ?? "BlueprintView"
+            )
+        }
+    }
+
+    private func logViewUpdateEnd() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
+                .end,
+                log: .blueprint,
+                name: "View Update",
+                signpostID: OSSignpostID(log: .blueprint, object: self)
+            )
+        }
+    }
+}
--- a/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
+++ b/BlueprintUI/Sources/Blueprint View/BlueprintView.swift
@@ -45,7 +45,7 @@ public final class BlueprintView: UIView {
                 return
             }
 
-            logElementAssigned()
+            Logger.logElementAssigned(view: self)
             
             setNeedsViewHierarchyUpdate()
             invalidateIntrinsicContentSize()
@@ -205,7 +205,7 @@ public final class BlueprintView: UIView {
         lastViewHierarchyUpdateBounds = bounds
         
         let start = Date()
-        logLayoutStart()
+        Logger.logLayoutStart(view: self)
 
         /// Grab view descriptions
         let viewNodes = element?
@@ -213,7 +213,7 @@ public final class BlueprintView: UIView {
             .resolve() ?? []
         
         let measurementEndDate = Date()
-        logLayoutEnd()
+        Logger.logLayoutEnd(view: self)
 
         rootController.view.frame = bounds
         
@@ -226,11 +226,11 @@ public final class BlueprintView: UIView {
         let scale = window?.screen.scale ?? UIScreen.main.scale
         rootNode.round(from: .zero, correction: .zero, scale: scale)
 
-        logViewUpdateStart()
+        Logger.logViewUpdateStart(view: self)
 
         rootController.update(node: rootNode, appearanceTransitionsEnabled: hasUpdatedViewHierarchy)
 
-        logViewUpdateEnd()
+        Logger.logViewUpdateEnd(view: self)
         let viewUpdateEndDate = Date()
         
         hasUpdatedViewHierarchy = true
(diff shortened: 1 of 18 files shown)


BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -0,0 +1,52 @@
+import CoreGraphics
+import os.log
+
+/// A size cache that also holds subcaches.
+protocol CacheTree: AnyObject {
```
watt, line 5: Considered renaming this to something more generalized, like `RenderContext`

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -419,3 +432,65 @@ extension BlueprintView {
     }
 }
 
+extension BlueprintView {
+    private func logLayoutStart() {
+        if #available(iOS 12.0, *) {
+            os_signpost(
```
kyleve, line 438: Fwiw in the past (and when I took a stab at signpost logging in a draft PR), I wrapped up the logging into a little helper so all the os_signpost stuff was in one place: https://github.com/square/Blueprint/pull/174/files#diff-34f7283a6db50334cae24989cc3fd7656859bae36d059ba5bb7a35206d3017e4R2

BlueprintUI/Sources/Internal/CacheFactory.swift
```diff
@@ -0,0 +1,12 @@
...
+enum CacheFactory {
+    static func makeCache(name: String, signpostRef: AnyObject = SignpostToken()) -> CacheTree {
+        // to disable caching, use this instead:
+        // FakeCache(name: name, signpostRef: signpostRef)
+
+        RenderPassCache(name: name, signpostRef: signpostRef)
+    }
+}
+
+private final class SignpostToken { }
```
kyleve, line 12: Mind commenting what this is for?

BlueprintUI/Sources/Internal/CacheTree.swift
kyleve: This seems good to me – I would've maybe gone with `MeasurementCache` or something

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// The sizes that are contained in this cache, keyed by size constraint.
+    subscript(constraint: SizeConstraint) -> CGSize? { get set }
+
+    /// Gets a subcache identified by the given key, or creates a new one.
+    func subcache(key: SubcacheKey, name: @autoclosure () -> String) -> CacheTree
+}
+
+struct SubcacheKey: RawRepresentable, Hashable {
+    /// A key indicating that this will be the only subcache
+    static let singleton = SubcacheKey(rawValue: -1)
```
kyleve, line 22: Should we do this, or would something like this be clearer?

```
enum SubcacheKey : Hashable {
   case singleton
   case indexed(Int)
}
```

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -0,0 +1,52 @@
...
+        subcache(key: SubcacheKey(rawValue: index), element: element)
+    }
+
+    /// Gets a subcache for an element.
+    func subcache(key: SubcacheKey = .singleton, element: Element) -> CacheTree {
+        subcache(
+            key: key,
+            name: key == .singleton
+                ? "\(self.name).\(type(of: element))"
+                : "\(self.name)[\(key.rawValue)].\(type(of: element))"
```
kyleve, line 49: Curious what sort of cost these strings getting built have. Wonder if we could make them memoized to avoid needing to do it during every traversal.

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -701,6 +705,8 @@ extension StackLayout {
         }
 
         func alignSegments(to alignment: AlignmentID.Type) -> [Segment] {
+            let crossMagnitudes = measureMagnitudes()
```
kyleve, line 708: Does this introduce a second call in some situations, potentially where before we would've only had one before?

kyleve:
This is really smart – well done!

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
```diff
@@ -43,11 +44,16 @@ public final class BlueprintView: UIView {
...
+
+            logElementAssigned()
             
             setNeedsViewHierarchyUpdate()
             invalidateIntrinsicContentSize()
         }
     }
+
+    /// A name to help identify this view when profiling or debugging
+    public var name: String?
```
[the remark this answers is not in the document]
watt: > Nit: `debugIdentifier`?

Well, I see this potentially coming from a workflow screen, and I kind of want to avoid "identifier" since it's just a display thing and not used for identity checks.

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
watt: > Fwiw in the past (and when I took a stab at signpost logging in a draft PR), I wrapped up the logging into a little helper so all the os_signpost stuff was in one place: https://github.com/square/Blueprint/pull/174/files#diff-34f7283a6db50334cae24989cc3fd7656859bae36d059ba5bb7a35206d3017e4R2

Sure, I can centralize these.

BlueprintUI/Sources/Internal/CacheTree.swift
watt: > Should we do this, or would something like this be clearer?
> 
> ```
> enum SubcacheKey : Hashable {
>    case singleton
>    case indexed(Int)
> }
> ```

Maybe? I was torn on how much to optimize this one. The singleton case is only used to improve the element name emitted to signpost, so we could also get rid of this type entirely and just pass in a hint along with the name.

BlueprintUI/Sources/Internal/CacheTree.swift
watt: > Curious what sort of cost these strings getting built have. Wonder if we could make them memoized to avoid needing to do it during every traversal.

Yeah, I had the same concern, and that's the reason for the autoclosure. It's only hit once per element per traversal, so I think it's pretty small compared to the amount of other work we do during layout. Dunno if we can do much better.

BlueprintUI/Sources/Layout/Stack.swift
watt: > Does this introduce a second call in some situations, potentially where before we would've only had one before?

I don't think so. The idea here is that we previously did it in every call to `_crossSegments`, and now it's skipped in some cases.

BlueprintUI/Sources/Internal/CacheTree.swift
kyleve: Ahh missed the autoclosure – and it's only used for signpost logging, meaning we won't hit it in release, right?

BlueprintUI/Sources/Layout/Stack.swift
kyleve: 👍

kyleve (approved):

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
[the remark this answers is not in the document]
kyleve: True – I guess when I see "name" I think of it as something that's potentially user-facing without a "debug" prefix.

kyleve:
Ok dumb question – I am testing this out locally to see what effect removing the measurement caching key stuff has on perf, and switching between the fake cache and the real one isn't giving me much perf difference:

```
Iterations: 71, Average Time: 0.14124150679145062
Iterations: 68, Average Time: 0.14753383923979366
```
I'm getting a small boost, but not much – I would've expected more here, but maybe not?

kyleve:
I removed the measurement cache key machinery to see what this gets us in a plain environment, it still seems slow given it should be caching at every node – maybe that's not happening? With cache:
```
Iterations: 12, Average Time: 0.9106179177761078
```

Fake cache:
```
Iterations: 9, Average Time: 1.1527716716130574
```

So it does help a bit – but not nearly as much as I'd expect w/ caching the layout at each node in the tree. Note that this is with the `test_deep_element_hierarchy` test.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -287,8 +301,27 @@ extension ElementContent {
             return result
         }
 
-        private func layoutItems(in environment: Environment) -> [(LayoutType.Traits, Measurable)] {
-            return children.map { ($0.traits, $0.content.measurable(in: environment)) }
+        private func layoutItems(
+            in environment: Environment,
+            cache: CacheTree
+        ) -> [(LayoutType.Traits, Measurable)] {
+            return zip(children.indices, children).map { index, child in
```
kyleve, line 308: Nit: When I was profiling, `zip` is for some reason slower than just doing `foo.enumerated()`, I'd recommend swapping out to that throughout the PR.

watt:
> I removed the measurement cache key machinery to see what this gets us in a plain environment, it still seems slow given it should be caching at every node – maybe that's not happening? With cache:
> 
> ```
> Iterations: 12, Average Time: 0.9106179177761078
> ```
> 
> Fake cache:
> 
> ```
> Iterations: 9, Average Time: 1.1527716716130574
> ```
> 
> So it does help a bit – but not nearly as much as I'd expect w/ caching the layout at each node in the tree. Note that this is with the `test_deep_element_hierarchy` test.

I think there are a couple of reasons why `test_deep_element_hierarchy` doesn't see much improvement from this.

1. `test_deep_element_hierarchy` is much wider than it is deep. It's only 4 elements deep, but has 1000 leaf labels. As each stack layer measures its children to divvy up the space, the leaves can be measured up to 8 times with a unique size constraint:

Screen Shot 2021-04-05 at 7 30 40 PM

The caching doesn't help when there are so many unique measurements.

2. It looks like floating point rounding error in the stack layout that can cause cache misses with values that are _almost_ the same. Just by increasing the width constraint from `1000` to `10000` I saw a big improvement in caching for this test.

kyleve:
Oh, the cache misses are interesting... huh. Wonder if we should round those measurements somehow? Though not sure how to do that super intelligently..

BlueprintUI/Sources/Blueprint View/BlueprintView.swift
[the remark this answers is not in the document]
n8chur: `debugName`?

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -0,0 +1,47 @@
...
+    var signpostRef: AnyObject { get }
+
+    /// The sizes that are contained in this cache, keyed by size constraint.
+    subscript(constraint: SizeConstraint) -> CGSize? { get set }
+
+    /// Gets a subcache identified by the given key, or creates a new one.
+    func subcache(key: SubcacheKey, name: @autoclosure () -> String) -> CacheTree
+}
+
+typealias SubcacheKey = Int
```
n8chur, line 20: Nit: Should this be scoped under `CacheTree`?

BlueprintUI/Tests/Element+Testing.swift
```diff
@@ -22,4 +22,14 @@ extension ElementContent {
     func measure(in constraint: SizeConstraint) -> CGSize {
         return measure(in: constraint, environment: .empty)
     }
+
+    /// A convenience wrapper to perform layout during testing, using an default `Environment` and
```
n8chur, line 26:
```suggestion
    /// A convenience wrapper to perform layout during testing, using a default `Environment` and
```

n8chur (approved):
Great work!

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -37,21 +38,36 @@ public struct ElementContent {
...
     /// - returns: The layout size needed by this content.
     public func measure(in constraint: SizeConstraint, environment: Environment) -> CGSize {
+        return measure(
+            in: constraint,
+            environment: environment,
+            cache: CacheFactory.makeCache(name: "ElementContent")
+        )
+    }
+
+    func measure(in constraint: SizeConstraint, environment: Environment, cache: CacheTree) -> CGSize {
```
n8chur, line 48: Should we combine these two measure functions into one with a default value for the cache? E.g. 
```swift
cache: CacheTree = CacheFactory.makeCache(name: "ElementContent")
```

BlueprintUI/Sources/Internal/CacheTree.swift
watt: > Ahh missed the autoclosure – and it's only used for signpost logging, meaning we won't hit it in release, right?

It _will_ be hit in release, since profiling is typically done against a release build.

watt:
I think I've addressed all feedback so far.

BlueprintUI/Sources/Internal/CacheTree.swift
kyleve: Thought / ask: Worth adding a way to easily short circuit the signpost logging if we want to do a pure perf trace without the signpost logging appearing? When I was testing, this was always visible in the traces, which while easy to ignore at a top level if inverting the call tree, using something like top functions means it can be a bit harder to realize that the stuff you're trying to optimize is actually just from the signpost logging.

Also, any idea if this stuff fires in device beta or app store builds? When I was testing w/ this in the checkout applet, of the ~15s of so of CPU time, the signpost logging took ~600ms, so its definitely not free.

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+import os.log
+
+extension OSLog {
```
kyleve, line 4: Nit: Could this go into the `Logger` file now?

BlueprintUI/Tests/ElementContentTests.swift
```diff
@@ -70,7 +70,80 @@ class ElementContentTests: XCTestCase {
             container.measure(in: SizeConstraint(CGSize.zero), environment: .empty),
             CGSize(width: 600, height: 800))
     }
-    
+
+    func test_cacheTree() {
```
kyleve, line 74: Worth writing a test to ensure that for a given size constraint during a full layout pass, the actual measure function is only called once? Eg, to verify the original bug that caused re-measurements due to the singleton key not being used in all cases.

kyleve (approved):

BlueprintUI/Sources/Internal/CacheTree.swift
watt: > Thought / ask: Worth adding a way to easily short circuit the signpost logging if we want to do a pure perf trace without the signpost logging appearing? When I was testing, this was always visible in the traces, which while easy to ignore at a top level if inverting the call tree, using something like top functions means it can be a bit harder to realize that the stuff you're trying to optimize is actually just from the signpost logging.

Sure. The [WWDC video](https://developer.apple.com/videos/play/wwdc2018/405/?time=863) on this talks about how they're meant to be super lightweight, but also gives a nice way to disable them.

> Also, any idea if this stuff fires in device beta or app store builds? When I was testing w/ this in the checkout applet, of the ~15s of so of CPU time, the signpost logging took ~600ms, so its definitely not free.

An app store build is still just a release build, so I would guess yes? But I would also guess that in a release build it's much faster than that.

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
watt: I wound up leaving it in a separate file along with the config to enable/disable.

BlueprintUI/Sources/Internal/CacheTree.swift
watt: Config added! This is now off by default. To enable it, there is a global config at `BlueprintLogging.enabled`.

PR #214, opened 2021-04-06
[DNR WIP] Built In Coordinate Space Reader

kyleve:


BlueprintUI.podspec
```diff
@@ -9,7 +9,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = '5.1'
 
-  s.ios.deployment_target = '11.0'
+  s.ios.deployment_target = '12.0'
```
kyleve, line 12: Bumping this lets me access `UIView.coordinateSpace`, which removes the implicit conversion to the coordinate space, which is nice, so consumers don't try this conversion randomly.

kyleve: (This is already the SPOS min deployment target; so this is basically a no-op.)

kyleve:
Hey @kylebshr @n8chur @watt thoughts on this? Lets us add coordinate space reading to any arbitrary element; which is useful for, eg, `MarketButton` and still let people put it into a button group. Also figured integrating more deeply than just an element would be good so we can allow people to hook into / operate alongside the inherited blueprint animations and transitions.

kylebshr:
- It does seem nice to have this built-in
- I'm a little concerned about the performance of using a CADisplayLink (maybe that's unwarranted, but I feel like we should profile - but my biggest concern is _what will be done_ in those callbacks, e.g., folks will probably try to re-render the whole blueprint tree based on an updated position?) In most cases emitting only when a blueprint layout pass is done is fine, could we have a special reader type for when you need `CADisplayLink` updates?
- RE: reading buttons in a button group, that's still something we'd need to add to the button API right? But the view description changes enable that?

PR #220, opened 2021-04-21
Add result builders for StackElement (Row & Column), EqualStack, Grid Row. 

ldstreet:
This adds function builders to Stacks and Grid Rows.

Example: 

StackElement (Row/Column):
```swift
Row {
    Element1()
    Element2()
        .stackChild(growPriority: 0.0)
    Element3()
} configure: {
    $0.horizontalUnderflow(.center)
}
```

EqualStack:

```swift
EqualStack(.vertical) {
    Element1()
    Element2()
    Element3()
} configure: {
    $0.spacing = 4.0
}
```

GridRow:

```swift
GridRow {
    Element1()
        .gridRowChild(width: .proportional(1))
    Element2()
        .gridRowChild(width: .proportional(2))
} configure: {
    $0.spacing = 5
}
```

To consider:

I carried forward the concept of a `configure` closure here for consistency, but it does seem odd that you could technically use the result builder and then `$0.add(...)` inside the configure block. An alternate approach could be to create helper functions on these types that copies, modifies, and returns the element so that it looks like this:

```swift
Row {
    Element1()
    Element2()
        .stackChild(growPriority: 0.0)
    Element3()
}
.horizontalUnderflow(.center)
```

Also, I am not sure if using `.proportional(1)` is the right move. I am a little less familiar with this `Element`. If defining this is always required and there is no sensible default, we could modify the init to look like this:

```swift
    public init(@Builder<GridRow.Child> _ children: () -> [GridRow.Child], configure: (inout Self) -> Void = { _ in }) {
        self.children = children()
        configure(&self)
    }

    GridRow {
        .init(width: .proportional(1),  element: Element1())
        .init(width: .proportional(2),  element: Element2())
    } configure: {
        $0.spacing = 5
    }
```


BlueprintUI/Sources/Layout/Builder/Builder.swift
```diff
@@ -0,0 +1,10 @@
+/// Generic result builder for converting blocks of `Child...` into `[Child]`.
+@_functionBuilder // change to @resultBuilder when we upgrade to Swift 5.4.
+public struct Builder<Child> {
```
kyleve, line 3: It would be nice to include the other builders here too – eg buildEither for ifs, etc..

BlueprintUI/Sources/Layout/Builder/EqualStack+Builder.swift
```diff
@@ -0,0 +1,15 @@
+extension EqualStack {
```
kyleve, line 1: I would just add these inits in `EqualStack.swift` (and same for the other files) – no reason to split it out, and it makes the API harder to find if you're just browsing that file otherwise.

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
```diff
@@ -0,0 +1,115 @@
...
+/// By default, elements will default to a nil key and the default `StackLayout.Traits` initializer.
+/// `@StackElementBuilder` will check every child to see if it can be type cast to a `StackChild`
+/// and then pull of the given traits and key and then apply those to the stack
+public struct StackChild: ProxyElement {
+    public var traits: StackLayout.Traits
+    public var key: AnyHashable?
+    private let wrappedElement: Element
+
+    public init(
+        wrappedElement: Element,
```
kyleve, line 14: Nit: Just `wrapping` for this – Swift style is to leave off the value type (`Element`), since that is clear from the type itself

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
```diff
@@ -0,0 +1,115 @@
...
+        key: AnyHashable? = nil
+    ) {
+        self.wrappedElement = wrappedElement
+        self.traits = traits
+        self.key = key
+    }
+    
+    public init(
+        wrappedElement: Element,
+        growPriority: CGFloat = 1.0,
```
kyleve, line 25: Nit: We generally push people towards the existing `addFixed` and `addFlexible` methods on the Stack – should we use that paradigm here too? Eg thinking of maybe an enum we could use or something, like:

```
    public init(
        wrappedElement: Element,
        sizing: Sizing,
        ...
    )

```

Where `Sizing` is like:

```
struct Sizing {
   static let fixed : Self = .init(0, 0)
   static let flexible : Self = .init(1, 1)

   var grow : CGFloat
   var shrink : CGFloat
}
```

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+extension StackElement {
+    /// Initializer using result builder to declaritively build up a stack.
+    /// - Parameters:
+    ///   - children: A block containing all elements to be included in the stack.
+    ///   - configure: A closure used to modify the stack.
+    /// - Note: If element is a StackChild, then traits and key will be pulled from the element, otherwise defaults are passed through.
+    public init(@Builder<Element> elements: () -> [Element], configure: (inout Self) -> Void = { _ in }) {
```
kyleve, line 87: I'd put the `configure` block first, since it's usually a lot shorter – so harder to lose after lots of content building. Thoughts?

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
```diff
@@ -0,0 +1,115 @@
...
+    ///   - key: A key used to disambiguate children between subsequent updates of the view
+    ///     hierarchy.
+    /// - Returns: A wrapped element with additional layout information for the `StackElement`.
+    public func stackChild(traits: StackLayout.Traits = .init(), key: AnyHashable? = nil) -> StackChild {
+        .init(wrappedElement: self, traits: traits, key: key)
+    }
+}
+
+
+extension StackElement {
```
kyleve, line 81: Thought: The two options presented for configuring the stack's options have so far been:

1) The `configure` block option added here
2) The option you described in the PR description, where we add "builder-like" callbacks that let you do `.horizontalUnderflow(.center)` on the stack to configure these options.

A third option that in retrospect, seems really obvious, is just including these params in the `init` as optional before the builder closure, eg:

```
public init(
   horizontalUnderflow : ... = .default,
   horizontalOverflow : ... = .default,
   @Builder<Element> elements: () -> [Element],
)
```

BlueprintUI/Sources/Layout/Builder/Builder.swift
ldstreet: I'm not exactly sure what you mean here. Could you give an example of what this would look like when used? Can't you achieve the same syntax for an `if` with normal closures?

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
ldstreet: Nice! I like this approach.

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
ldstreet: I started with the configure block first, but switched to make it second for a couple reasons:
1. Consistency - the other elements I have used all seem to place the configure block last
2. When I am building a layout I start with building the hierarchy and then tweaking the containers and elements. Like this:

```swift
// 1)
Column {
    Element1()
    Text("Hello, world!")
    Element2()
}

// 2)
Column {
    Element1()
    Text("Hello, world!") { 
        $0.lineLimit = 1
    }
    Element2()
} configure: {
    $0.horizontalUnderflow(.center)
}
```

By having the configure block come second I can simply append any modifications rather whereas if I were to reverse the parameters then I would need to insert the `elements` label into the code. Its nothing major but is (imo) a small quality of life improvement.

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
ldstreet: I like this idea, but I wonder about scalability. We would likely need to put every instance variable into the initializer. Is that ok? As the property list grows the init could become unwieldy, though a lot of that will be hidden via default values. 

We could also consider a hybrid approach...add the most commonly used properties to the init and then use the configure block or builder-functions to customize the rest.

As a separate note: one thing that detracts me from the configure block is the possibility of mixing the `@Builder` approach with `add(Element:)`. This could lead to an unclear API (which gets called first? configure or the builder?).

ldstreet: Coming back to this after paternity leave and I think leaning on the initializer makes the most sense...default values help keep it clean and it avoids the boiler plate/maintenance of adding all of the modifiers. I will update the PR so we can see how it looks in practice.

kyleve: Welcome back!

BlueprintUI/Sources/Layout/Builder/Builder.swift
```diff
@@ -0,0 +1,35 @@
...
+    public static func buildEither(first: [Element]) -> Element {
+        ElementContainer(elements: first)
+    }
+
+    public static func buildEither(second: [Element]) -> Element {
+        ElementContainer(elements: second)
+    }
+}
+
+private struct ElementContainer: ProxyElement {
```
ldstreet, line 30: @kyleve I'm curious to hear your thoughts here. The function builder needs the types to be homogenous so it doesn't allow for me to mix arrays of Elements and single Elements . To solve for it I created this private wrapper Element that should never render and simply holds an array of elements. Then in `buildBlock` I check to see if the element can be cast as an ElementContainer and if so I extract out the contained elements. I got the idea for ElementContainer from SwiftUI's ForEach but there those views can actually be rendered and have uses beyond what I am doing here.

BlueprintUI/Sources/Layout/Builder/StackElement+Builder.swift
ldstreet: Ok, I updated the syntax to remove configure blocks and instead add configurations to the initializer. An example looks like this:

````swift
Row(horizontalUnderflow: .center) {
    Element1()
    Element2()
        .stackChild(growPriority: 0.0)
    Element3()
}
````

File: BlueprintUI/Sources/Layout/Builder/StackChild.swift [hunk 8]
```diff
@@ -0,0 +1,87 @@
+import UIKit
+
+/// `StackChild` is a wrapper which allows an element to define its StackLayout traits and Keys.
+/// This struct is particularly useful when working with `@StackElementBuilder`.
+/// By default, elements will default to a nil key and the default `StackLayout.Traits` initializer.
+/// `@StackElementBuilder` will check every child to see if it can be type cast to a `StackChild`
+/// and then pull of the given traits and key and then apply those to the stack
+public struct StackChild: ProxyElement {
```
watt, line 8: It feels weird to me that this is an element. I would suggest inverting these, making `StackChild` the currency of the builder, and handling plain elements with a `buildExpression` func.

Something like:

```swift
struct StackChild {
    var element: Element
    var traits: Traits = .defaultTraits
}

@_functionBuilder
struct StackBuilder {
    typealias Component = [StackChild]

    static func buildBlock(_ children: Component...) -> Component {
        children.flatMap { $0 }
    }

    static func buildExpression(_ element: Element) -> Component {
        // promote naked element to a child with default traits
        [StackChild(element: element)]
    }
}

extension Element {
    // or perhaps separate modifiers for priority, key, etc
    func stackTraits(
        growPriority: CGFloat = 1,
        shrinkPriority: CGFloat = 1,
        key: AnyHashable? = nil
    ) -> StackChild {
        StackChild(
            element: self,
            traits: Traits(
                growPriority: growPriority,
                shrinkPriority: shrinkPriority,
                key: key
            )
        )
    }
}
```

ldstreet: I do like the idea of changing the currency of the builder...but one downside of changing to be a StackChild is that we will now need new builders for different container based views such as EqualStacks and GridRows. What seems odd about encapsulating stack information in an Element? It seems like many elements in the library simply carry along extra information to handle layout, styling, interaction, etc.

BlueprintUI/Sources/Layout/Builder/Builder.swift
ldstreet: Never mind here....I was able to get rid of it. Just needed the weekend to clear my head 😄

BlueprintUI/Sources/Layout/Builder/StackChild.swift
watt: > one downside of changing to be a StackChild is that we will now need new builders for different container based views such as EqualStacks and GridRows

Yeah, but that seems like the right approach to me. Each of those containers has a different child type, with different traits. With the approach you have now, each container has to do some version of an `element as? StackChild` downcast to get the type it wants, and that goes away if we use the right currency for each container.

> It seems like many elements in the library simply carry along extra information to handle layout, styling, interaction, etc.

Elements carry this information but do not share it; they are generally self-contained and loosely coupled. They work together by composition, while remaining opaque to each other.

ldstreet: This makes sense to me. Thanks for explaining! I will go ahead and make these changes 👍

ldstreet: @watt I created two new builders that use StackChild and GridRowChild as a currency (see latest commit). I definitely like not having to cast Elements to StackChild anymore!

BlueprintUI/Sources/Layout/Builder/ElementBuilder.swift
```diff
@@ -0,0 +1,29 @@
+/// Generic result builder for converting blocks of `Child...` into `[Child]`.
+@_functionBuilder // change to @resultBuilder when we upgrade to Swift 5.4.
+public struct ElementBuilder {
```
kyleve, line 3: Thinking out loud: It would be nice if we could avoid triple-duplicating these; since they're all nearly the same. Could we introduce generics here around the content type, and defer to that content type for the bottom two methods:

```
    public static func buildExpression(_ element: Element) -> Component {
        [element.gridRowChild(width: .proportional(1))]
    }

    public static func buildExpression(_ elements: [Element]) -> Component {
        elements.map { $0.gridRowChild(width: .proportional(1)) }
    }
```

Which seem to be the two implementations which diverge across the builder types?

(Or maybe this is basically just boilerplate that will never change; and having three is fine! Thoughts?)

ldstreet: Love this idea...adding a generic builder definitely cleans up the code. I created a protocol to factor out the differing logic.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+    public var key: AnyHashable?
+    
+    public struct Sizing {
+        public static let fixed: Self = .init(growPriority: 0, shrinkPriority: 0)
+        public static let flexible: Self = .init(growPriority: 1, shrinkPriority: 1)
+
+        public var growPriority: CGFloat
+        public var shrinkPriority: CGFloat
+        
+        public init(growPriority: CGFloat = 1, shrinkPriority: CGFloat = 1) {
```
kyleve, line 20: I would leave the defaults off here. This API is a super big footgun (most people don't understand it at all), so I'd make them pass this explicitly.

Going further; since this is an API split; I wonder if we should take this opportunity to make the sizing for result builders only be able to take in Bools for these.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+/// By default, elements will default to a nil key and the default `StackLayout.Traits` initializer.
+/// `@StackElementBuilder` will check every child to see if it can be type cast to a `StackChild`
+/// and then pull of the given traits and key and then apply those to the stack
+public struct StackChild {
+    public let element: Element
+    public var traits: StackLayout.Traits
+    public var key: AnyHashable?
+    
+    public struct Sizing {
+        public static let fixed: Self = .init(growPriority: 0, shrinkPriority: 0)
```
kyleve, line 14: Two others we could add here would be `grows` and `shrinks` with 0 and 1 as their values (and vice-versa).

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+    
+    /// Wraps an element with a `StackChild` in order to customize `StackLayout.Traits` and the key.
+    /// - Parameters:
+    ///   - sizing: Controls the amount of extra space distributed to this child during underflow and overflow
+    ///   - alignmentGuide: Allows for custom alignment of a child along the cross axis.
+    ///   - key: A key used to disambiguate children between subsequent updates of the view
+    ///     hierarchy.
+    /// - Returns: A wrapped element with additional layout information for the `StackElement`.
+    public func stackChild(
+        sizing: StackChild.Sizing = .flexible,
```
kyleve, line 64: IMO I wouldn't given this a default; since you have a 50% chance of getting what you want wrong.

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -43,13 +43,48 @@ public struct Column: StackElement {
...
+    ///   - alignment: Determines the cross-axis layout
+    ///   - verticalUnderflow: Determines the layout when there is extra free space available.
+    ///   - verticalOverflow: Determines the layout when there is not enough space to fit all children as measured.
+    ///   - horizontalAlignment: Specifies how children will be aligned horizontally.
+    ///   - minimumVerticalSpacing: Spacing in between elements.
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a StackChild, then traits and key will be pulled from the element, otherwise defaults are passed through.
+    public init(
+        alignment: StackLayout.Alignment = StackLayout.defaultColumn.alignment,
+        verticalUnderflow: StackLayout.UnderflowDistribution = StackLayout.defaultColumn.underflow,
```
kyleve, line 74: API design question: I wonder if we should remove "vertical" and "horizontal" from these; since that's implicit from being a Column.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: Which defaults would you give to an element that does not specify its stack traits in the builder? I can leave off the defaults here, but then I need to add them back when implicitly converting from `Element` ->  `StackChild`. Or we could get rid of the implicit conversion, but that definitely makes the syntax more cumbersome.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: Hm granted I'm not as familiar with the API, but don't these priorities allow you to balance priorities among multiple elements similar to `contentHugging` and `compressionResistence` in UIKit? Wouldn't a bool only allow comparison between two elements? Unless you mean to change it to `allowShrinking: Bool` and `allowGrowing: Bool`.

watt: > don't these priorities allow you to balance priorities among multiple elements similar to `contentHugging` and `compressionResistence` in UIKit? 

Kind of, but in practice, the priority algorithm (documented on `StackElement.add`) doesn't do what you probably expect, and so we often recommend sticking to either 0/0 for "fixed" items and 1/1 for "flexible" items. The `addFixed` and `addFlexible` methods on stacks were added to encourage that.

> Wouldn't a bool only allow comparison between two elements? Unless you mean to change it to `allowShrinking: Bool` and `allowGrowing: Bool`.

I think what Kyle is suggesting is to limit this API to 2 discrete cases (fixed & flexible) rather than exposing the numbers at all. And I agree! The underlying values are still stored on the traits if we want to expose them later.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+extension Element {
+    
+    /// Wraps an element with a `StackChild` in order to customize `StackLayout.Traits` and the key.
+    /// - Parameters:
+    ///   - sizing: Controls the amount of extra space distributed to this child during underflow and overflow
+    ///   - alignmentGuide: Allows for custom alignment of a child along the cross axis.
+    ///   - key: A key used to disambiguate children between subsequent updates of the view
+    ///     hierarchy.
+    /// - Returns: A wrapped element with additional layout information for the `StackElement`.
+    public func stackChild(
```
watt, line 63: I'm going to bikeshed for `stackTraits` here. As a modifier, this adds traits to a child; every item without it is already a child.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
watt: I think a default is safe as long as it matches the default you'd get without using this modifier at all. Might be nice for cases where you just want to add the key or alignment guide.

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -43,13 +43,48 @@ public struct Column: StackElement {
...
+    ///   - minimumVerticalSpacing: Spacing in between elements.
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a StackChild, then traits and key will be pulled from the element, otherwise defaults are passed through.
+    public init(
+        alignment: StackLayout.Alignment = StackLayout.defaultColumn.alignment,
+        underflow: StackLayout.UnderflowDistribution = StackLayout.defaultColumn.underflow,
+        overflow: StackLayout.OverflowDistribution = StackLayout.defaultColumn.overflow,
+        alignment: ColumnAlignment = StackLayout.defaultColumn.columnAlignment,
+        minimumSpacing: CGFloat = StackLayout.defaultColumn.minimumSpacing,
+        @ElementBuilder<StackChild> elements: () -> [StackChild]
```
watt, line 78: I'm not in favor of this, but I'll mention that there's an alternative to passing all these params in the initializer: you can pass the stack into the builder closure as an `inout` param like the existing configure blocks do, to get a hybrid approach.

```swift
Row { row in
  row.alignment = .leading

  Foo()
  Bar()
}
```

It might require a builder method like this, but it seems to work!
```swift
static func buildExpression(_: Void) -> Component {
  []
}
```

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+        sizing: StackChild.Sizing = .flexible,
+        alignmentGuide: ((ElementDimensions) -> CGFloat)? = nil,
+        key: AnyHashable? = nil
+    ) -> StackChild {
+        .init(
+            element: self,
+            sizing: sizing,
+            alignmentGuide: alignmentGuide, key: key
```
n8chur, line 71:
```suggestion
            alignmentGuide: alignmentGuide,
            key: key
```

BlueprintUI/Sources/Layout/Builder/StackChild.swift
n8chur: `stackItem`? e.g.:

```swift
Column {
    Label(...)
        .stackItem(sizing: .flexible)
}
```

I am kind of a fan of naming this after the type it is converted into, since it's in the global `Element` scope, but I don't feel too strongly about it.

BlueprintUI/Sources/Layout/Column.swift
n8chur: This is clever, although as we discussed on our call let's start with enumerating everything in the initializer and re-evaluate if that ends up not working out for us!

I wish this proposed solution didn't require `_ in` when you don't need it, although I suppose we could provide a second initializer that doesn't pass anything in!

n8chur (commented):
Thanks again for the contribution! This is great!

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: Does this look good to you? 
image

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
...
+/// This struct is particularly useful when working with `@ElementBuilder`.
+/// By default, elements will default to a nil key and the default `StackLayout.Traits` initializer.
+/// `@StackElementBuilder` will check every child to see if it can be type cast to a `StackChild`
+/// and then pull of the given traits and key and then apply those to the stack
+public struct StackChild {
+    public let element: Element
+    public var traits: StackLayout.Traits
+    public var key: AnyHashable?
+    
+    public struct Sizing {
```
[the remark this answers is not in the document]
ldstreet: image

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: I'm thinking that stackTraits might get confused with StackLayout.Traits. Also, this struct is closer to a `child` since `children` in StackElement is defined as` [(element: Element, traits: StackLayout.Traits, key: AnyHashable?)]` which is pretty much the shape of StackChild. I would almost be tempted to change `children` from an array of tuples to `[StackChild]` but that would cause a lot of regressions.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,91 @@
+import UIKit
+
+/// `StackChild` is a wrapper which allows an element to define its StackLayout traits and Keys.
+/// This struct is particularly useful when working with `@ElementBuilder`.
+/// By default, elements will default to a nil key and the default `StackLayout.Traits` initializer.
+/// `@StackElementBuilder` will check every child to see if it can be type cast to a `StackChild`
+/// and then pull of the given traits and key and then apply those to the stack
+public struct StackChild {
```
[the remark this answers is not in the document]
ldstreet: We could but I think it more relates to `StackElement` then it does `StackLayout`. Unfortunately we can't use protocols as a namespace. I am still open to moving it but not sure I see a strong enough connection between the two.

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -43,13 +43,48 @@ public struct Column: StackElement {
...
+    /// - Parameters:
+    ///   - alignment: Determines the cross-axis layout
+    ///   - verticalUnderflow: Determines the layout when there is extra free space available.
+    ///   - verticalOverflow: Determines the layout when there is not enough space to fit all children as measured.
+    ///   - horizontalAlignment: Specifies how children will be aligned horizontally.
+    ///   - minimumVerticalSpacing: Spacing in between elements.
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a StackChild, then traits and key will be pulled from the element, otherwise defaults are passed through.
+    public init(
+        alignment: StackLayout.Alignment = StackLayout.defaultColumn.alignment,
```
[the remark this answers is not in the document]
ldstreet: Good call, I realized I had duplicate properties in here...oops!

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
```diff
@@ -0,0 +1,42 @@
+/// `GridRowChild` is a wrapper which allows an element to define its `GridRow.Width` and `key`.
+/// This struct is particularly useful when working with a result builder initializer that uses `@ElementBuilder`.
+/// By default, elements will be set to a nil key and `.proportional(1)` initializer.
+/// The initializer will check every child to see if it can be type cast to a `GridRowChild`
+/// and then pull out the given width and key and then apply those to the row.
+public struct GridRowChild {
+    public let element: Element
+    public var width: GridRow.Width
```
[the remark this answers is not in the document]
ldstreet: Hm, good call. I think it must have been copy and paste error. I always like to start with `let` until a need for mutability is proven.

BlueprintUI/Sources/Layout/Builder/ElementBuilder.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// Allow Elements to be implicitly converted into `Child`.
+    public static func buildExpression(_ elements: [Element]) -> Children {
+        elements.map(Child.init)
+    }
+}
+
+/// Defines a way for an`Element` to be implicitly converted into the conforming type (the child of a container).
+/// In practive, this allows us to pass an `Element` directly into the result builder without manually converting to `Child` (i.e. Converting `Element` -> `StackChild`.
+public protocol ElementBuilderChild {
+    init(from element: Element)
```
n8chur, line 45: Could this just be `init(element: Element)`?

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,105 @@
+import UIKit
+
+
+/// `StackChild` is a wrapper which holds an element along with  its StackLayout traits and Keys.
+/// This struct is particularly useful when working with `@ElementBuilder<StackChild>`.
+/// By default, elements inside of `ElementBuilder<StackChild>` will be implicitly converted to a StackChild
+/// with a nil key and the default `StackLayout.Traits` initializer. But when given a `StackChild` via `Element.stackChild(...)` modifier or initialized directly, `@ElementBuilder<StackChild>`pull out the given traits and key and then apply those to the stack
```
n8chur, line 7:
```suggestion
/// with a nil key and the default `StackLayout.Traits` initializer. But when given a `StackChild` via 
/// `Element.stackChild(...)` modifier or initialized directly, `@ElementBuilder<StackChild>`pull out the 
/// given traits and key and then apply those to the stack
```

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -175,6 +199,28 @@ extension StackElement {
...
+
+    /// Converts alignment to `RowAlignment`.
+    public var rowAlignment: Row.RowAlignment {
+        .init(alignment)
+    }
+
+    /// Converts alignment to `ColumnAlignment`.
+    public var columnAlignment: Column.ColumnAlignment {
+        .init(alignment)
+    }
```
n8chur, line 222: Should we move these into extensions defined at the bottom of `Row.Swift`/`Column.swift`?

n8chur (approved):
LGTM overall! Would love to get a couple more stamps from other UIS folks on this before merging, though.

Thanks again for taking this on!

BlueprintUI/Sources/Layout/Builder/ElementBuilder.swift
ldstreet: or maybe even `init(_ element: Element)` since the type itself is pretty explanatory

BlueprintUI/Sources/Layout/Stack.swift
ldstreet: I have no opinion on this so I'll go ahead and move them 👍

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
```diff
@@ -0,0 +1,17 @@
+/// Wraps around an Element for consistency with `StackChild` and `GridRowChild`.
+/// In the future this struct could hold traits used for laying out inside an EqualStack
+extension EqualStack {
+    public struct Child {
```
watt, line 4: Can you move this file's contents into `EqualStack.swift`? It's very much part of that implementation, I think.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
```diff
@@ -0,0 +1,58 @@
+import UIKit
+
+// MARK: - child modeling -
+extension GridRow {
+    /// A child of a `GridRow`.
+    public struct Child {
```
watt, line 6: Same request here re: moving into the parent file. The Child type is an inseparable part of the implementation of GridRow.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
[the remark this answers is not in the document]
watt: I would argue they are quite strongly connected! StackElement, StackLayout, and this child type are all part of the stack implementation and won't be used elsewhere.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
watt: > I would almost be tempted to change `children` from an array of tuples to `[StackChild]` but that would cause a lot of regressions.

I'd encourage this — most stacks are built with the `add` methods, and don't access the `children` directly.

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -43,13 +43,46 @@ public struct Column: StackElement {
...
+    ///   - overflow: Determines the layout when there is not enough space to fit all children as measured.
+    ///   - horizontalAlignment: Specifies how children will be aligned horizontally.
+    ///   - minimumSpacing: Spacing in between elements.
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a plain Element, then that Element will be implicitly converted into a `StackChild` with default values
+    public init(
+        alignment: ColumnAlignment = StackLayout.defaultColumn.columnAlignment,
+        underflow: StackLayout.UnderflowDistribution = StackLayout.defaultColumn.underflow,
+        overflow: StackLayout.OverflowDistribution = StackLayout.defaultColumn.overflow,
+        minimumSpacing: CGFloat = StackLayout.defaultColumn.minimumSpacing,
```
watt, line 76: How do these default values show up in Xcode's autocomplete? It might be nice to put the defaults in here explicitly, even if there's some duplication, to make it easy to see which properties you need to override and which you can skip.

The doc comments should say what the defaults are, either way.

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -43,13 +43,46 @@ public struct Column: StackElement {
...
+    ///   - layout: A StackLayout describing the column's layout
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a StackChild, then traits and key will be pulled from the element, otherwise defaults are passed through.
+    public init(
+        layout: StackLayout = .defaultColumn,
+        @ElementBuilder<StackChild> elements: () -> [StackChild]
+    ) {
+        self.init(elementsBuilder: elements)
+        self.layout = layout
+    }
```
watt, line 61: Do we need this init? You should never need to create a StackLayout directly.

BlueprintUI/Sources/Layout/Row.swift
```diff
@@ -39,7 +39,7 @@ public struct Row: StackElement {
             }
         }
 
-        var stackAlignment: StackLayout.Alignment {
+        public var stackAlignment: StackLayout.Alignment {
```
watt, line 42: Intentional? It's not public on `ColumnAlignment`.

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -33,6 +33,30 @@ extension StackElement {
...
+    ///   - items: The array of items to transform.
+    ///   - transform: Closure to be applied to each item `T` which transforms that item into an `Element`.
+    ///   - configure: A closure used to modify the stack.
+    public init<T>(_ items: [T], transform: (T) -> StackChild, configure: (inout Self) -> Void = { _ in }) {
+        self.init()
+        children = items.map(transform).map { stackChild in
+            (stackChild.element, stackChild.traits, stackChild.key)
+        }
+        configure(&self)
+    }
```
watt, line 58: Do these need to be public?

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,107 @@
...
+        )
+    }
+
+    /// Wraps an element with a `StackChild` in order to customize `StackLayout.Traits` and the key.
+    /// - Parameters:
+    ///   - traits: Contains traits that affect the layout of individual children in the stack.
+    ///   - key: A key used to disambiguate children between subsequent updates of the view
+    ///     hierarchy.
+    /// - Returns: A wrapper containing this element with additional layout information for the `StackElement`.
+    public func stackChild(traits: StackLayout.Traits = .init(), key: AnyHashable? = nil) -> StackChild {
```
watt, line 97: This one feels unnecessary to me — it's very rare to create traits directly.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
```diff
@@ -0,0 +1,107 @@
...
+public struct StackChild {
+    public let element: Element
+    public let traits: StackLayout.Traits
+    public let key: AnyHashable?
+
+    public enum Priority {
+        case fixed
+        case flexible
+        case grows
+        case shrinks
```
watt, line 19: I worry that `grows` and `shrinks` are more likely to be used incorrectly than to be useful.

watt (commented):
Getting close now!

BlueprintUI/Sources/Layout/Builder/StackChild.swift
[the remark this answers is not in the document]
ldstreet: Ok sure, I also will rename the modifier from `stackChild(...)` to `stackLayoutChild()` for consistency. Does that sound ok?

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: Hm, even though it' not widely used, wouldn't this still be a breaking change? I could make a new `stackChildren: [StackLayout.Child]` (naming? _children? something else?) and then keep `children: [(element: Element, traits: StackLayout.Traits, key: AnyHashable?)]` as a computed variable that derives from [StackLayout.Child]. Then I could add a deprecation warning on top.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: I'll remove 👍

BlueprintUI/Sources/Layout/Column.swift
ldstreet: This makes a lot of sense and lets me delete the defaultRow code 👍

BlueprintUI/Sources/Layout/Row.swift
ldstreet: I needed this at one point, but seems like it is no longer necessary. Good catch

BlueprintUI/Sources/Layout/Stack.swift
ldstreet: Probably not since Row and Column have inits with all default values. Also, I don't think the second one is necessary at all anymore since I added the ability for result builders to take in arrays of children. The same thing can be accomplished with a simple `map` on the items you are trying to convert to Elements.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
kyleve: (Yes! Sorry for the slow reply..)

BlueprintUI/Sources/Layout/Column.swift
kyleve: Do builders allow passing parameters? I thought they didn't; or at least last I checked this didn't work..

BlueprintUI/Sources/Layout/Stack.swift
kyleve: +1 agree with Westin; thanks!

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
kyleve: Bump here

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
```diff
@@ -0,0 +1,17 @@
...
+        public let element: Element
+
+        public init(element: Element) {
```
kyleve, line 7: I'd maybe make this `init(_ element : ...`; thoughts?

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
```diff
@@ -0,0 +1,17 @@
...
+            self.element = element
+        }
+    }
+}
+
+extension EqualStack.Child: ElementBuilderChild {
+    public init(_ element: Element) {
```
kyleve, line 14: This doesn't seem super useful / I'd just make the `EqualStack.Child` implement this, and make it conform to this directly.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
kyleve: Bump here too

BlueprintUI/Sources/Layout/Builder/StackChild.swift
kyleve: IMO it's more useful / more clear than having people pass `0` and `1` to the individual params. What about more verbose names?

`growsOnUnderflow`
`shrinksOnOverflow`

BlueprintUI/Sources/Layout/Builder/StackChild.swift
kyleve: It would be a breaking changes - but we (UIS) can handle applying this change.

BlueprintUI/Sources/Layout/Column.swift
kyleve: Inversely; its nice to be able to put these in one place. Hardcoding breaks that. I'd prefer leaving this as is, IMO? If someone _really_ cares; they can do look at the value.

BlueprintUI/Sources/Layout/Column.swift
kyleve: (If this does work, it's kinda nice, IMO – as mentioned we can provide two inits!)

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -33,6 +33,30 @@ extension StackElement {
...
+            (stackChild.element, stackChild.traits, stackChild.key)
+        }
+    }
+
+    /// Initializes stack transforming an array of some item into an array of elements.
+    /// - Parameters:
+    ///   - items: The array of items to transform.
+    ///   - transform: Closure to be applied to each item `T` which transforms that item into an `Element`.
+    ///   - configure: A closure used to modify the stack.
+    public init<T>(_ items: [T], transform: (T) -> StackChild, configure: (inout Self) -> Void = { _ in }) {
```
kyleve, line 52: Minor nit: Swift has moved away from `T`, `U`, etc, to names that reflect the actual value of the generic (eg `Element`); might be nice to do that here too.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+extension Element {
+    /// Wraps an element with a `GridRowChild` in order to provide meta information that a `GridRow` can aply to its layout.
+    /// - Parameters:
+    ///   - key: A unique identifier for the child.
+    ///   - width: The sizing for the element.
+    /// - Returns: `GridRowChild`
+    public func gridRowChild(key: AnyHashable? = nil, width: GridRow.Width) -> GridRow.Child {
```
kyleve, line 49: I'm curious for others' thoughts here, but I wonder if we want to _not_ add these kinds of helpers; since they're only valid in a given context (you using the result in a grid row (etc). If you use the wrong helper, or an element uses the helper is presented in a different context; it basically is a silent failure. 

Eg, maybe we just want to force people to explicitly return a `GridRow.Child`; but then again, you can still fuck that up in the same exact ways, since everything is type erased anyways... So maybe this is fine! Thoughts?

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
ldstreet: Done

BlueprintUI/Sources/Layout/Builder/EqualStackChild.swift
ldstreet: Agreed.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
ldstreet: Done.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
ldstreet: So because we changed `ElementBuilder` to use a corresponding `Child` (i.e. GridRow.Child) as its currency and that `Child` no longer conforms to `Element`, that means this modifier can really only be used inside of its supported container. If you use it in a mismatching container, or an API that takes in an element, you will get a compile time error.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: I like `growsOnUnderflow` and `shrinksOnOverflow` but then I would expect something like `growsAndShrinks` instead of `flexible` for consistent semantics. Otherwise I wouldn't immediate catch that `flexible` is a combination of both `grows` and `shrinks`.

BlueprintUI/Sources/Layout/Builder/StackChild.swift
ldstreet: Cool! Then I'll go ahead and make it.

BlueprintUI/Sources/Layout/Column.swift
ldstreet: one issue here is API confusion on what happens if you call row.add(...) inside the block? What order will the element be added?

BlueprintUI/Sources/Layout/Stack.swift
ldstreet: I actually removed this init since I added the ability for the builder to take in `[Element]` and `[Child]` and its just as easy to call `items.map(...)` inside the block.

BlueprintUI/Sources/Layout/Column.swift
n8chur: IMO let's leave this out of this first iteration. If we find the initializer parameters to be too cumbersome we can consider adding a second init with this syntax (or the other alternatives we discussed).

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
n8chur: I think what we have here is worth the trade offs (e.g. less code/nesting for common use case), and as @ldstreet pointed out, you can't really use it in the wrong context since it's not an `Element`.

BlueprintUI/Sources/Layout/Column.swift
kyleve: > one issue here is API confusion on what happens if you call row.add(...) inside the block? What order will the element be added?

Very good point!

BlueprintUI/Sources/Layout/Builder/StackChild.swift
kyleve: IMO they're special enough that I think it's fine to have slightly different naming than the two common use cases.

BlueprintUI/Sources/Layout/Builder/GridRowChild.swift
kyleve: Ah cool! Sounds good.

kyleve (approved):
Looks good! Really nice work; and thank you!

kyleve:
(Also, before merging; let's also ensure to update the CHANGELOG.md)

kyleve:
(Oops, wrong button)

kyleve:
(We can handle this after this initial PR: We should probably add result builders for Overlay, too – UI Sys can handle this one)

BlueprintUI/Sources/Layout/Column.swift
```diff
@@ -41,15 +41,39 @@ public struct Column: StackElement {
...
     public init() {}
 
+    /// Initializer using result builder to declaritively build up a stack.
+    /// - Parameters:
+    ///   - alignment: Specifies how children will be aligned horizontally. Default: .leading
+    ///   - underflow: Determines the layout when there is extra free space available. Default: .spaceEvenly
+    ///   - overflow: Determines the layout when there is not enough space to fit all children as measured. Default: .condenseProportionally
+    ///   - minimumSpacing: Spacing in between elements. Default: 0
+    ///   - elements: A block containing all elements to be included in the stack.
+    /// - Note: If element is a plain Element, then that Element will be implicitly converted into a `StackLayout.Child` with default values
```
watt, line 60: Since we are introducing radically new API here, I think it would be helpful to have more in the docs, particularly explaining the way to add child traits.

```suggestion
    /// Creates a Column, using result builder syntax. Columns display a list of items in a vertical
    /// stack.
    ///
    ///     Column {
    ///         Label(text: "Welcome")
    ///
    ///         TextField(text: username)
    ///         TextField(text: password)
    ///
    ///         Button(
    ///             onTap: handleSignIn,
    ///             wrapping: Label(text: "Sign In")
    ///         )
    ///     }
    ///
    /// By default, each item in the column will be stretched or compressed with equal priority in
    /// the event of overflow or underflow. You can control this behavior by adding a
    /// `stackLayoutChild` modifier to an item.
    ///
    ///     Column {
    ///         ImportantHeader()
    ///             .stackLayoutChild(priority: .fixed)
    ///
    ///         LessImportantContent()
    ///     }
    ///
    /// You can also use this modifier to add keys and alignment guides. See `StackElement.add` for
    /// more information.
    ///
    /// ## See Also
    /// [StackElement.add()](x-source-tag://StackElement.add)
    ///
    /// - Parameters:
    ///   - alignment: Specifies how children will be aligned horizontally. Default: `.leading`
    ///   - underflow: Determines the layout when there is extra free space available. Default:
    ///     `.spaceEvenly`
    ///   - overflow: Determines the layout when there is not enough space to fit all children as
    ///     measured. Default: `.condenseProportionally`
    ///   - minimumSpacing: Spacing in between elements. Default: `0`
    ///   - elements: A block containing all elements to be included in the stack.
```

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -965,3 +974,101 @@ extension SizeConstraint.Axis {
...
+    public struct Child {
+        public let element: Element
+        public let traits: StackLayout.Traits
+        public let key: AnyHashable?
+
+        public enum Priority {
+            case fixed
+            case flexible
+            case growsOnUnderflow
+            case shrinksOnOverflow
```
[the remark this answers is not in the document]
kyleve: So where I struggle is: If you want this, this is nicer than specifying the values. But also, no one ever really wants this, I guess. So yeah, maybe we should just do fixed and flexible and call it a day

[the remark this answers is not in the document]
ldstreet: Removed.

BlueprintUI/Tests/GridRowTests+ResultBuilders.swift
```diff
@@ -0,0 +1,279 @@
...
+
+class GridRowTestsResultBuilders: XCTestCase {
+    func test_defaults() {
+        let gridRow = GridRow()
+        XCTAssertTrue(gridRow.children.isEmpty)
+        XCTAssertEqual(gridRow.verticalAlignment, .fill)
+        XCTAssertEqual(gridRow.spacing, 0)
+    }
+
+    func test_measure_unconstrained() {
```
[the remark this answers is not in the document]
ldstreet: Good call, adding some more specific tests now. I noticed that you expect the default width for TestElement() to be .absolute(0). Currently I have .proportional(1). Is that incorrect?

n8chur:
Register, Market, and Blueprint are all on Xcode 12.5.1 now so we should be good to move forward when this is ready!

kyleve:
> Register, Market, and Blueprint are all on Xcode 12.5.1 now so we should be good to move forward when this is ready!

dj-blob-cat

BlueprintUI/Tests/GridRowTests+ResultBuilders.swift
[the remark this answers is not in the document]
watt: > Good call, adding some more specific tests now. I noticed that you expect the default width for TestElement() to be .absolute(0). Currently I have .proportional(1). Is that incorrect?

I got `.absolute(0)` from `GridRowLayout.defaultTraits`, but I don't think that value is actually used. I like your default better.

watt (approved):
Looks ready to me! Thanks for going through all this feedback, I think it came out really nice!

PR #224, opened 2021-05-21
Add scroll to top trigger for ScrollView

alleywalley:


BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -233,6 +234,15 @@ fileprivate final class ScrollerWrapperView: UIView {
     }
 
     private var refreshAction: () -> Void = { }
+    
+    private var scrollToTopTrigger: ScrollView.Trigger? {
+        didSet {
+            oldValue?.action = { }
+            scrollToTopTrigger?.action = { [weak self] in
+                self?.scrollView.setContentOffset(CGPoint(x: 0, y: 0), animated: false)
```
kyleve, line 242: I think we should allow specifying an animation param here; probably by passing it to the `fire()` method.

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -485,3 +499,21 @@ extension ScrollerWrapperView : KeyboardObserverDelegate {
...
+
+    final public class Trigger {
+
+        var action: () -> Void
+
+        public init() {
+            action = { }
+        }
+
+        public func fire() {
```
kyleve, line 513: I wonder if we should:
a) Take in an `animated` param here
b) Add an `enum` that could control various scrolling options: Scroll to top, or bottom, etc.

watt: > I wonder if we should:
> a) Take in an `animated` param here
> b) Add an `enum` that could control various scrolling options: Scroll to top, or bottom, etc.

Yeah, I agree. I could see something like this being real handy:

```swift
public struct ContentOffset {
    public struct ScrollingContext {
        public var contentSize: CGSize
        public var scrollViewBounds: CGRect
        public var contentInsets: UIEdgeInsets
    }

    public typealias Provider = (ScrollingContext) -> CGPoint

    public static let top = ContentOffset { _ in .zero }
    public static let bottom = ContentOffset { context in
        CGPoint(x: 0, y: context.contentSize.height - context.scrollViewBounds.size.height)
    }

    var provider: Provider

    public init(provider: @escaping Provider) {
        self.provider = provider
    }
}
```

And then instead of just `fire()`, you could call it like:
```swift
scrollTrigger.scroll(to: .top, animated: true)
```

watt (commented):
Mind also adding a bullet to the changelog?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -32,6 +32,7 @@ public struct ScrollView: Element {
 
     public var delaysContentTouches: Bool = true
 
+    public var scrollToTopTrigger: ScrollTrigger?
```
allendeleon, line 35: Need to rename this

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -329,6 +346,10 @@ fileprivate final class ScrollerWrapperView: UIView {
         if self.scrollView.contentInsetAdjustmentBehavior != scrollView.contentInsetAdjustmentBehavior {
             self.scrollView.contentInsetAdjustmentBehavior = scrollView.contentInsetAdjustmentBehavior
         }
+        
+        if let trigger = scrollView.setContentOffsetTrigger {
```
kyleve, line 350: I don't think you want this to be an `if let`, just a normal set; so if the value is cleared, it will be removed, right?

kyleve (approved):

kyleve:
One comment, but LGTM – thank you!

BlueprintUICommonControls/Sources/ScrollView.swift
allendeleon: - [ ] remove `if let` before setting trigger

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -485,3 +504,41 @@ extension ScrollerWrapperView : KeyboardObserverDelegate {
...
+            public var contentSize: CGSize
+            public var scrollViewBounds: CGRect
+            public var contentInsets: UIEdgeInsets
+        }
+
+        public typealias Provider = (ScrollingContext) -> CGPoint
+
+        public static let top = ContentOffset { _ in .zero }
+        public static let bottom = ContentOffset { context in
+            CGPoint(x: 0, y: context.contentSize.height - context.scrollViewBounds.size.height)
```
watt, line 535: Ah, I forgot to include insets my other comment — the y value here should probably include `+ context.contentInsets.bottom`

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -32,6 +32,7 @@ public struct ScrollView: Element {
 
     public var delaysContentTouches: Bool = true
 
+    public var setContentOffsetTrigger: ScrollTrigger?
```
watt, line 35: Could probably just call this `contentOffsetTrigger` to avoid the awkwardness of a var that starts with "set".

watt (approved):

PR #225, opened 2021-05-22
Add scroll indicator insets

kylebshr:


BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -330,6 +331,10 @@ fileprivate final class ScrollerWrapperView: UIView {
             self.scrollView.contentInsetAdjustmentBehavior = scrollView.contentInsetAdjustmentBehavior
         }
 
+        if self.scrollView.scrollIndicatorInsets != scrollView.scrollIndicatorInsets {
+            self.scrollView.scrollIndicatorInsets = scrollView.scrollIndicatorInsets
```
kyleve, line 335: Is this going to fight the changes in `updateBottomContentInsetWithKeyboardFrame`? We might want to flow all the inset updates through one method. This is what I ended up doing in Listable: https://github.com/kyleve/Listable/blob/21c3cabc5bcd7df2351405ca8b18117dfb54d9e9/ListableUI/Sources/ListView/ListView.swift#L287

kylebshr: ah crap yeah

PR #229, opened 2021-06-04
Improve ConstrainedAspectRatio content modes

kylebshr:
This PR improves the content modes of `ConstrainedAspectRatio` by renaming `fill` and `fit` to `fitContent` and `shrinkContent` respectively (which more accurately describe their behavior) and adding a `fillParent` and `fitParent` content mode, which are a common use case. The default content mode is `fitContent` ( e.g., the same behavior as it was before).

### DIFF
--- a/BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
@@ -8,31 +8,31 @@ public struct ConstrainedAspectRatio: Element {
         /// The content will be sized to fill its parents largest dimension, while maintaining the
         /// specified aspect ratio.
         ///
-        /// If the parent has unconstrained dimensions, the content size will be used for that
-        /// dimension. This means that in an unconstrained measurement, this is equivalent to
-        /// `fitContent`.
+        /// If the parent is unconstrained in all dimensions, the content size will be used for
+        /// measurement and will behave like `fitContent`. If the parent is unconstrained in one
+        /// dimension, the element will fill the constrained dimension.
         case fillParent
 
         /// The content will be sized to fit its parents smallest dimension, while maintaining the
         /// specified aspect ratio.
         ///
-        /// If the parent has unconstrained dimensions, the content size will be used for that
-        /// dimension. This means that in an unconstrained measurement, this is equivalent to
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -85,12 +166,11 @@ public extension Element {
...
     /// - parameters:
     ///   - aspectRatio: The aspect ratio that the content size should match.
-    ///   - contentMode: Whether the aspect ratio should be reached by expanding the content
-    ///     element's size to fill its parent or shrinking it to fit.
+    ///   - contentMode: How the content should size itself relative to its parent.
     ///
     func constrainedTo(
         aspectRatio: AspectRatio,
-        contentMode: ConstrainedAspectRatio.ContentMode = .fill
+        contentMode: ConstrainedAspectRatio.ContentMode = .fitContent
```
kyleve, line 173: This doesn't change the default behaviour, right?

kyleve (approved):

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -2,35 +2,113 @@ import UIKit
...
+                availableWidth = contentSize.width
+            }
+
+            let constrainedHeight = aspectRatio.height(forWidth: availableWidth)
+            let constrainedWidth = aspectRatio.width(forHeight: availableHeight)
+
+            switch self {
+            case .fillParent:
+                if constraint.width.constrainedValue == nil &&
+                    constraint.height.constrainedValue == nil {
```
kyleve, line 62: Should this be an `&&`, or an `||` in terms of the fallback? Though I guess if it was an `||`, unconstrained parents like scroll views wouldn't work.

kylebshr: The cases for a single dimension are handled right below this statement, so it should be `&&`

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
kylebshr: Correct!

PR #230, opened 2021-06-21
Use structs within the ViewDescription config; it's a bit faster.

kyleve:
Happened onto this while looking at some other code. Turns out, moving this to use structs makes things a little bit faster – switching to struct for value bindings, and avoiding the AnyHashable wraps. This seems to be about a ~10% perf gain for view descriptions. Not huge, but may as well.


BlueprintUI/Sources/ViewDescription/ViewDescription.swift
```diff
@@ -220,39 +217,40 @@ extension ViewDescription.Configuration {
...
+        var value: Value
 
         init(keyPath: ReferenceWritableKeyPath<View, Value>, value: Value) {
             self.value = value
             self.keyPath = keyPath
         }
 
-        override func apply(to view: View) {
+        func apply(to anyView: UIView) {
+            let view = anyView as! View
```
n8chur, line 252: Nit: Is there any way to avoid this cast? Could `AnyValueBinding` become a `struct` that's generic on the `View` perhaps?

n8chur (approved):

BlueprintUI/Sources/ViewDescription/ViewDescription.swift
kyleve: I might be able to get rid of the protocol via some clever closure stuff in init, let me see.

kyleve: Hmm, actually, no, that won't work, since we still need to access the typed value.

kyleve: Yeah I can't figure out how to do this.

n8chur: [This works](https://github.com/square/Blueprint/compare/kve/view-description-structs-only...westin/type-erasure?expand=1) but I'm not convinced it's much better...

n8chur: What you have seems fine!

kyleve: Yeah that's kinda where I had ended up – had to cast the value instead of the view; but one of them has to be cast. I think I'll stick with this

BlueprintUI/Tests/PerformancePlayground.swift
```diff
@@ -24,7 +24,19 @@ class PerformancePlayground : XCTestCase
```
kyleve, line 24: Revert this file

kylebshr (approved):

PR #234, opened 2021-06-25
Automatically propagate inherited environment to child blueprint views

kyleve:
This PR introduces automatic propagation of `Environment` through to arbitrarily nested `BlueprintViews`. This is useful, eg in `Market`, where in many places we nest additional `BlueprintViews` within stateful controls such as buttons – previously `Environment` values either would not flow down to these nested views, or we would have to manually flow the `Environment` down to these views.

We accomplish this propagation through updates to `NativeViewNode` and `NativeViewController`. When we create a `NativeViewNode`, we now record its `Environment`. Then, when the `NativeViewController` goes to update the `node`'s associated `UIView`, we assign the `UIView`'s `nativeViewNodeBlueprintEnvironment` (added via associated objects). This means that the `UIView` in the view hierarchy now has the correct `Environment`, which corresponds to its `Element`'s environment at measurement / layout time. 

To fetch our inherited `Environment`, when it comes time for a `BlueprintView` to layout its `Element` hierarchy; it asks for its own `inheritedBlueprintEnvironment` – this walks up the superview hierarchy until we find a `nativeViewNodeBlueprintEnvironment`, signaling that we are inside a parent `BlueprintView`; and uses this `Environment` to seed the `makeEnvironment()` function.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -36,9 +36,34 @@ public final class BlueprintView: UIView {
 
     private let rootController: NativeViewController
 
-    /// A base environment to render elements with.
-    /// Some keys may be overriden based on the traits of the view itself.
+    /// A base environment used when laying out and rendering the element tree.
+    ///
+    /// Some keys will be overridden with the traits from the view itself. Eg, `windowSize`, `safeAreaInsets`, etc.
+    ///
+    /// If this blueprint view is within another blueprint view, the environment of the parent view
+    /// will be inherited by this view if `automaticallyInheritsEnvironmentFromContainingBlueprintViews`
+    /// is enabled. In the case of matching keys in both the inherited environment and the provided
+    /// environment, the values from this environment will take priority over the inherited environment.
     public var environment: Environment {
+        didSet {
+            setNeedsViewHierarchyUpdate()
+            invalidateIntrinsicContentSize()
+        }
+    }
+    
+    ///
+    /// If `true`, then Blueprint will automatically inherit the  ``Environment`` from the nearest
+    /// parent ``BlueprintView`` in the view hierarchy.
+    ///
+    /// If `false`, then only the values from ``BlueprintView/environment`` will be used to
+    /// seed the environment passed to the element hierarchy.
+    ///
+    /// This property is recursive – if the nearest parent ``BlueprintView`` also sets this property to
+    /// true, then you will inherit the ``Environment`` from that view's parent ``BlueprintView``, and so on.
+    ///
+    /// Defaults to `true`.
+    ///
+    public var automaticallyInheritsEnvironmentFromContainingBlueprintViews : Bool = true {
         didSet {
             setNeedsViewHierarchyUpdate()
             invalidateIntrinsicContentSize()
@@ -78,8 +103,13 @@ public final class BlueprintView: UIView {
         rootController = NativeViewController(
             node: NativeViewNode(
                 content: UIView.describe() { _ in },
+                // Because no layout update occurs here, passing an empty environment is fine;
+                // the correct environment will be passed during update.
+                environment: .empty,
                 layoutAttributes: LayoutAttributes(),
-                children: []))
+                children: []
+            )
+        )
     
         super.init(frame: CGRect.zero)
         
@@ -216,10 +246,12 @@ public final class BlueprintView: UIView {
         
         let start = Date()
         Logger.logLayoutStart(view: self)
+        
+        let environment = self.makeEnvironment()
 
         /// Grab view descriptions
         let viewNodes = element?
-            .layout(layoutAttributes: LayoutAttributes(frame: bounds), environment: makeEnvironment())
+            .layout(layoutAttributes: LayoutAttributes(frame: bounds), environment: environment)
             .resolve() ?? []
         
         let measurementEndDate = Date()
@@ -229,6 +261,7 @@ public final class BlueprintView: UIView {
         
         var rootNode = NativeViewNode(
             content: UIView.describe() { _ in },
+            environment: environment,
             layoutAttributes: LayoutAttributes(frame: bounds),
             children: viewNodes
         )
@@ -238,7 +271,12 @@ public final class BlueprintView: UIView {
 
         Logger.logViewUpdateStart(view: self)
 
-        rootController.update(node: rootNode, appearanceTransitionsEnabled: hasUpdatedViewHierarchy)
+        rootController.update(
+            node: rootNode,
+            context: .init(
+                appearanceTransitionsEnabled: hasUpdatedViewHierarchy
+            )
+        )
 
         Logger.logViewUpdateEnd(view: self)
         let viewUpdateEndDate = Date()
@@ -268,7 +306,19 @@ public final class BlueprintView: UIView {
     }
     
     private func makeEnvironment() -> Environment {
-        var environment = self.environment
+                
+        let inherited : Environment = {
+            if
+                self.automaticallyInheritsEnvironmentFromContainingBlueprintViews,
+                let inherited = self.inheritedBlueprintEnvironment
+            {
+                return inherited
+            } else {
+                return .empty
+            }
+        }()
+        
+        var environment = inherited.merged(prioritizing: self.environment)
 
         if let displayScale = window?.screen.scale {
             environment.displayScale = displayScale
@@ -312,7 +362,6 @@ extension BlueprintView {
     final class NativeViewController {
 
         private var viewDescription: ViewDescription
-
         private var layoutAttributes: LayoutAttributes
         
         private (set) var children: [(ElementPath, NativeViewController)]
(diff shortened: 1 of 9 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -37,13 +37,23 @@ public final class BlueprintView: UIView {
...
             invalidateIntrinsicContentSize()
         }
     }
+    
+    /// A base environment inherited when this ``BlueprintView`` is within the view hierarchy of another ``BlueprintView``.
+    /// Set during the view layout portion of a layout cycle.
+    /// Some keys may be overridden based on the traits of the view itself, or based on the explicit environment set on `environment`.
+    public internal(set) var inheritedEnvironment : Environment? {
+        didSet {
+            setNeedsViewHierarchyUpdate()
```
kyleve, line 53: Actually; I don't think I need this; since the layout pass itself should do this?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -403,7 +434,7 @@ extension BlueprintView {
                             contentView.insertSubview(controller.view, at: index)
                         }
 
-                        controller.update(node: child, appearanceTransitionsEnabled: true)
+                        controller.update(node: child, context: context)
```
kyleve, line 437: Passing `true` all the time here was wrong; but also a no-op kind of wrong; since we'd never have any children to update if we hadn't done at least one update path.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -441,3 +474,21 @@ extension BlueprintView {
         }
     }
 }
+
+extension BlueprintView.NativeViewController {
+    
+    /// A context value passed to the ``BlueprintView/NativeViewController`` instance as the view tree is updated by blueprint.
+    /// Contains information relevant to the correct construction and management of the view tree.
+    struct UpdateContext {
```
kyleve, line 482: This isn't a needed change; but I did it since it shortens up the callsites and makes it easier to add additional information here without appending tons of parameters (like animation state, which I was planning on looking at reviving soon).

SampleApp/Podfile.lock
```diff
@@ -2,7 +2,7 @@ PODS:
   - BlueprintUI (0.27.0)
   - BlueprintUI/Tests (0.27.0)
   - BlueprintUICommonControls (0.27.0):
-    - BlueprintUI
+    - BlueprintUI (= 0.27.0)
```
kyleve, line 5: Out of date on `main`.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -69,8 +69,50 @@ public struct Environment {
...
+
+
+extension UIView {
+    
+    /// The ``Environment`` for the ``Element`` that this view represents in a Blueprint element tree,
+    /// or if the view is not explicitly managed by Blueprint, the ``Environment`` of
+    /// the nearest superview that is managed by Blueprint.
+    ///
+    /// If no views in the superview hierarchy are managed by Blueprint, this property returns nil.
+    var inheritedBlueprintEnvironment : Environment? {
```
kyleve, line 95: I kept this internal for now; but would this be helpful at all, eg in the Market view layer; would it simplify anything?

BlueprintUI/Tests/EnvironmentTests.swift
```diff
@@ -132,13 +137,26 @@ class EnvironmentTests: XCTestCase {
...
+    func test_merged() {
+        var lhs = Environment.empty
+        lhs[TestKey.self] = .wrong
+        
+        var rhs = Environment.empty
+        rhs[TestKey.self] = .right
+        
+        let output = lhs.merged(prioritizing: rhs)
+        
+        XCTAssertEqual(output[TestKey.self], .right)
```
n8chur, line 158: Would it also be worth testing that setting values for properties exclusively on each side are merged correctly? E.g. only set `KeyA` on `lhs` and `KeyB` on `rhs` and validate that both values exist on the merged object.

BlueprintUI/Tests/EnvironmentTests.swift
```diff
@@ -163,6 +181,47 @@ class EnvironmentTests: XCTestCase {
...
+        
+        XCTAssertEqual(first.inheritedBlueprintEnvironment?[TestingKey1.self], 1)
+        XCTAssertEqual(second.inheritedBlueprintEnvironment?[TestingKey1.self], 1)
+        XCTAssertEqual(third.inheritedBlueprintEnvironment?[TestingKey1.self], 1)
+        
+        first.nativeViewNodeBlueprintEnvironment = nil
+        
+        XCTAssertEqual(first.inheritedBlueprintEnvironment?[TestingKey1.self], nil)
+        XCTAssertEqual(second.inheritedBlueprintEnvironment?[TestingKey1.self], nil)
+        XCTAssertEqual(third.inheritedBlueprintEnvironment?[TestingKey1.self], nil)
```
n8chur, line 206: Should we also test setting a non-nil value on a non-root view and ensure that superviews are not resolving to a value?
```suggestion
        XCTAssertEqual(third.inheritedBlueprintEnvironment?[TestingKey1.self], nil)
        
        second.nativeViewNodeBlueprintEnvironment = 2

        XCTAssertEqual(first.inheritedBlueprintEnvironment?[TestingKey1.self], nil)
        XCTAssertEqual(second.inheritedBlueprintEnvironment?[TestingKey1.self], 2)
        XCTAssertEqual(third.inheritedBlueprintEnvironment?[TestingKey1.self], 2)
        
```

n8chur (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -268,7 +300,19 @@ public final class BlueprintView: UIView {
     }
     
     private func makeEnvironment() -> Environment {
-        var environment = self.environment
+                
+        let inherited : Environment = {
+            if
+                let inherited = self.inheritedBlueprintEnvironment,
+                self.automaticallyInheritsEnvironmentFromContainingBlueprintViews
```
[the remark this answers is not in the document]
kyleve: Good call

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -323,20 +366,28 @@ extension BlueprintView {
             self.viewDescription = node.viewDescription
             self.layoutAttributes = node.layoutAttributes
             self.children = []
+            
             self.view = node.viewDescription.build()
+            self.view.nativeViewNodeBlueprintEnvironment = node.environment
+        }
+        
+        deinit {
+            self.view.nativeViewNodeBlueprintEnvironment = nil
```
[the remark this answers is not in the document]
kyleve: This is a belt-and-suspenders thing – in _theory_ you should always be returning a new view from your view description; but you don't _have_ to – it could be a static / instance owned elsewhere. So I am clearing the value out when we stop controlling the view just in case.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -37,13 +37,32 @@ public final class BlueprintView: UIView {
     private let rootController: NativeViewController
 
     /// A base environment to render elements with.
-    /// Some keys may be overriden based on the traits of the view itself.
+    /// Some keys may be overridden based on the traits of the view itself.
```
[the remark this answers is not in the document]
kyleve: I updated the docs here:

```
    /// A base environment used when laying out and rendering the element tree.
    ///
    /// Some keys will be overridden with the traits from the view itself. Eg, `windowSize`, `safeAreaInsets`, etc.
    ///
    /// If this blueprint view is within another blueprint view, the environment of the parent view
    /// will be inherited by this view if `automaticallyInheritsEnvironmentFromContainingBlueprintViews`
    /// is enabled. In the case of matching keys in both the inherited environment and the provided
    /// environment, the values from this environment will take priority over the inherited environment.
```

BlueprintUI/Tests/EnvironmentTests.swift
kyleve: Sure

BlueprintUI/Tests/EnvironmentTests.swift
kyleve: I'm not quite sure I follow what you mean here – can you expand a bit?

BlueprintUI/Tests/EnvironmentTests.swift
kyleve: Done

BlueprintUI/Tests/EnvironmentTests.swift
kyleve: (If I'm reading the comment correctly; the concern I think is a superview finding a value from a subview? As implemented; that can't happen, since we only traverse up.)

n8chur: Yeah, I don't see how the test I'm suggesting would fail—no big deal if you want to omit it. It's really just asserting that you're not using a reference to the same environment in each layer of the tree.

PR #235, opened 2021-06-29
Collapse nested AdaptedEnvironments

kyleve:
Easy one – on init, check if our passed value is another `AdaptedEnvironment`, and if it is; append to its adaptors instead of wrapping it. For deeper nesting where you're doing:

```
element
.adaptedEnvironment(...)
.adaptedEnvironment(...)
.adaptedEnvironment(...)
.adaptedEnvironment(...)
```

Should remove the slight measurement / traversal penalty of deepening the hierarchy.

### DIFF
--- a/BlueprintUI/Tests/AdaptedEnvironmentTests.swift
+++ b/BlueprintUI/Tests/AdaptedEnvironmentTests.swift
@@ -1,0 +1,74 @@
+//
+//  AdaptedEnvironmentTests.swift
+//  BlueprintUI-Unit-Tests
+//
+//  Created by Kyle Van Essen on 6/28/21.
+//
+
+@testable import BlueprintUI
+import XCTest
+
+
+class AdaptedEnvironmentTests : XCTestCase {
+    
+    func test_adapting() {
+        let view = BlueprintView()
+        
+        var environment : Environment? = nil
+        
+        view.element = TestElement { environment = $0 }
+        .adaptedEnvironment(key: TestingKey1.self, value: "adapted1")
+        
+        view.layoutIfNeeded()
+        
+        XCTAssertEqual(environment?[TestingKey1], "adapted1")
+    }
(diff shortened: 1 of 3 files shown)


BlueprintUI/Tests/AdaptedEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    
+    func test_wrapping_multiple() {
+        let view = BlueprintView()
+        
+        var environment : Environment? = nil
+        
+        let element = TestElement { environment = $0 }
+        .adaptedEnvironment(key: TestingKey.self, value: "layer 1")
+        .adaptedEnvironment(key: TestingKey.self, value: "layer 2")
+        .adaptedEnvironment(key: TestingKey.self, value: "layer 3")
```
kylebshr, line 35: Can we add one more test ensuring this outermost adaptation flows all the way down (e.g., if it has a different key)

kylebshr: (Might be excessive ¯\\\_(ツ)_/¯)

kylebshr (approved):

BlueprintUI/Tests/AdaptedEnvironmentTests.swift
kyleve: Sure, done

SampleApp/Podfile.lock
```diff
@@ -2,7 +2,7 @@ PODS:
   - BlueprintUI (0.27.0)
   - BlueprintUI/Tests (0.27.0)
   - BlueprintUICommonControls (0.27.0):
-    - BlueprintUI
+    - BlueprintUI (= 0.27.0)
```
kyleve, line 5: This was out of date on `main`.

PR #231, opened 2021-06-22
Pass context with environment to backingViewDescription

kylebshr:
Adds an `ElementContext` object for passing information to `backingViewDescription`, and adds the environment to the context (along with the existing `bounds` and `subtreeExtent`). This allows view-backed elements to access the environment they're rendered in.

PR #237, opened 2021-07-01
ElementContent & Layout API Updates

kyleve:
This PR contains a few updates to the `ElementContent` and `Layout` APIs, in order to:

- Pass `Environment` (via `LayoutContext`) through to the various measurement and layout functions to ensure that you always have access to the current `Environment` as a consumer. In general I'd like us to move towards "Environment Everywhere"; this is to discourage people from passing `.empty` through to measurement methods, because, especially in Market, an `.empty` `Environment` is likely almost _always_ wrong in subtle ways (In a follow up, we may want to rename / remove `.empty` or change the `Environment` API to make getting an empty environment harder / more obvious that it's not what you want in most cases).
- Introduce the `LayoutContext` as a container for `Environment`, etc, as passed through to these methods, to reduce the need for future breaking changes when we want to add additional optional parameters to layout and measurement. This is similar to the `ViewDescriptionContext` introduced in https://github.com/square/Blueprint/pull/231.
- Take advantage of the newly introduced `LayoutContext` to fix the lifetime of static measurement / prototype views. These are now tied to the lifetime of their containing blueprint view / standalone measurement pass, via the `LayoutContext.measure(...)` methods. This fixes an issue that John D raised; where because the static shared view instances lived forever, they would outlive their context (eg being logged in), and appear as a memory leak / retain cycle.
- Update `Layout` and `SingleChildLayout` to take in  the`LayoutContext` and `LayoutItems`; `LayoutItems` is a class-based wrapper around the previous top-level array of layout items we'd pass around – turns out that this actually has a pretty significant performance benefit when looking at the bookkeeping vs passing around the array of tuples previously – something like a 20-30% perf benefit in testing a deep hierarchy where actual measurement and layout was a no-op – 0.12s before vs 0.07-0.08s afterwards.
- Add `EnvironmentElement` to allow easier creation of Environment-backed elements.

### `ElementContent`Updates
`ElementContent` and its associated initializers have been updated to take in a `LayoutContext` object.

### `Layout & SingleChildLayout`Updates

These two protocols have been updated to take in the new `LayoutContext`, as well as use the new `LayoutItems` type. I also re-ordered the parameters to make them read a little better and unify the ordering across layout and single child layout.

```
public protocol SingleChildLayout {

    func measure(
        child: Measurable,
        in constraint : SizeConstraint,
        with context: LayoutContext
    ) -> CGSize

    func layout(
        child: Measurable,
        in size : CGSize,
        with context : LayoutContext
    ) -> LayoutAttributes
}

public protocol Layout {

    func measure(
        items: LayoutItems<Self.Traits>,
        in constraint : SizeConstraint,
        with context: LayoutContext
    ) -> CGSize

    func layout(
        items: LayoutItems<Self.Traits>,
        in size : CGSize,
        with context : LayoutContext
    ) -> [LayoutAttributes]    
}
```

### Introduce `LayoutContext`, Fix Measurement View Lifetimes
Similar to `ViewDescriptionContext`, introduces a context object that is passed through the layout process; to give global access to the `Environment`, and allows us to add further optional parameters to layout and measurement methods without further breaking API changes.

LayoutContext also provides a way to perform easier measurement using prototype views that respects the containing Blueprint View's lifecycle. `UIViewElement` has been moved to use this caching method.

```
public struct LayoutContext {
    
    /// The current `Environment` for the `Element`.
    public var environment : Environment

    public func measure<ElementType:Element, ViewType:UIView>(
        _ element : ElementType,
        using view: @autoclosure () -> ViewType,
        measure : (ViewType) -> CGSize
    ) -> CGSize { ... }
}
```

### Introduce `LayoutItems<Traits>`
As mentioned above; wrapping up this previously array of tuples into a reference type provides pretty significant performance benefits when it comes to just running the mechanics of a layout cycle – I was seeing improvements of about 20-30%; or 0.12s to ~0.07s when running the measure + layout pass where measurement and layout were no-ops; allowing performance benchmarking of the mechanics of the cycle itself. Digging into this a bit; its because copying around the value type and tuples actually has a high enough cost in very hot codepaths to matter. I did this testing in release mode, and the same cost / benefit was still there.

### Introduce `EnvironmentElement`
Because we now have access to the `Environment` everywhere, we can introduce a shorthand API for elements which depend on the `Environment`, removing the need to wrap everything in a nested `EnvironmentReader` or `GeometryReader` (which IMO, hurts readability because of the extra 1-2 levels of nesting):

```
public protocol EnvironmentElement : Element {    
    func content(in size : SizeConstraint, with environment : Environment) -> Element
}
```

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -30,23 +30,32 @@ public struct ElementContent {
     }
     
     // MARK: Measurement & Children
-
-    /// Measures the required size of this element's content.
-    /// - Parameters:
-    ///   - constraint: The size constraint.
-    ///   - environment: The environment to measure in.
-    /// - returns: The layout size needed by this content.
-    public func measure(in constraint: SizeConstraint, environment: Environment) -> CGSize {
-        return measure(
+    
+    /// Measures the size needed to display the element within the provided size constraint.
+    ///
+    /// ### Usage
+    /// You usually call this method from within the `measure` method of a `Layout`, or within the
+    /// measurement function you provide to an `ElementContent` instance. In either of these cases,
+    /// you should pass through the `LayoutContext` provided to you to ensure the measured elements
+    /// downstream have full access to their measurement caches, environment, etc.
+    /// ```
+    /// public var content: ElementContent {
+    ///     ElementContent { constraint, context -> CGSize in
+    ///         self.wrapped.content.measure(in: constraint, with: context)
+    ///     }
+    /// }
+    /// ```
+    public func measure(in constraint : SizeConstraint, with context: LayoutContext) -> CGSize {
+        measure(
             in: constraint,
-            environment: environment,
+            with: context,
             cache: CacheFactory.makeCache(name: "ElementContent")
         )
     }
 
-    func measure(in constraint: SizeConstraint, environment: Environment, cache: CacheTree) -> CGSize {
-        environment.measurementCache.measurement(with: self.measurementCachingKey, in: constraint) {
-            self.storage.measure(in: constraint, environment: environment, cache: cache)
+    func measure(in constraint : SizeConstraint, with context: LayoutContext, cache: CacheTree) -> CGSize {
+        context.measurementCache.measurement(with: self.measurementCachingKey, in: constraint) {
+            self.storage.measure(in: constraint, with: context, cache: cache)
         }
     }
 
@@ -55,13 +64,14 @@ public struct ElementContent {
     }
 
     func performLayout(
-        attributes: LayoutAttributes,
-        environment: Environment,
+        in size: CGSize,
+        with context : LayoutContext,
         cache: CacheTree
-    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
-        return storage.performLayout(
-            attributes: attributes,
-            environment: environment,
+    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] // TODO: Turn this into a reference type too
+    {
+        storage.performLayout(
+            in: size,
+            with: context,
             cache: cache
         )
     }
@@ -75,10 +85,10 @@ extension ElementContent {
     /// - parameter measurementCachingKey: An optional key to use to cache measurement. See the `MeasurementCachingKey`
     ///     documentation for more.
     /// - parameter builder: A closure that provides the content `Element` based on the provided `SizeConstraint`
-    ///     and `Environment`.
+    ///     and `LayoutContext`.
     public init(
         measurementCachingKey : MeasurementCachingKey? = nil,
-        build builder: @escaping (SizeConstraint, Environment) -> Element
+        build builder: @escaping (SizeConstraint, LayoutContext) -> Element
     ) {
         self.measurementCachingKey = measurementCachingKey
         self.storage = LazyStorage(builder: builder)
@@ -99,7 +109,10 @@ extension ElementContent {
         layout: SingleChildLayout,
         measurementCachingKey : MeasurementCachingKey? = nil
     ) {
-        self = ElementContent(layout: SingleChildLayoutHost(wrapping: layout), measurementCachingKey: measurementCachingKey) {
+        self = ElementContent(
+            layout: SingleChildLayoutHost(wrapping: layout),
+            measurementCachingKey: measurementCachingKey
+        ) {
             $0.add(key: key, element: child)
         }
     }
@@ -114,7 +127,11 @@ extension ElementContent {
         child: Element,
         measurementCachingKey : MeasurementCachingKey? = nil
     ) {
-        self = ElementContent(child: child, layout: PassthroughLayout(), measurementCachingKey: measurementCachingKey)
+        self = ElementContent(
+            child: child,
+            layout: PassthroughLayout(),
+            measurementCachingKey: measurementCachingKey
+        )
     }
 
     /// Initializes a new `ElementContent` with no children that delegates to the provided `Measurable`.
@@ -135,17 +152,20 @@ extension ElementContent {
     /// Initializes a new `ElementContent` with no children that delegates to the provided measure function.
     ///
     /// - parameter measurementCachingKey: An optional key to use to cache measurement. See the `MeasurementCachingKey` documentation for more.
-    /// - parameter measureFunction: How to measure the `ElementContent` in the given `SizeConstraint`.
+    /// - parameter measureFunction: How to measure the `ElementContent` in the given `SizeConstraint` and `LayoutContext`.
     public init(
         measurementCachingKey : MeasurementCachingKey? = nil,
-        measureFunction: @escaping (SizeConstraint) -> CGSize
+        measureFunction: @escaping (SizeConstraint, LayoutContext) -> CGSize
     ) {
-        self = ElementContent(measurable: Measurer(_measure: measureFunction), measurementCachingKey: measurementCachingKey)
+        self = ElementContent(
+            measurable: Measurer(provider: measureFunction),
+            measurementCachingKey: measurementCachingKey
+        )
     }
 
     /// Initializes a new `ElementContent` with no children that uses the provided intrinsic size for measuring.
     public init(intrinsicSize: CGSize) {
-        self = ElementContent(measureFunction: { _ in intrinsicSize })
+        self = ElementContent { _, _ in intrinsicSize }
     }
 }
 
@@ -160,12 +180,12 @@ extension ElementContent {
     public init(
         child: Element,
         measurementCachingKey : MeasurementCachingKey? = nil,
-        environment environmentAdapter: @escaping (inout Environment) -> Void
+        environment adapter: @escaping (inout Environment) -> Void
     ) {
         self.measurementCachingKey = measurementCachingKey
         
         self.storage = EnvironmentAdaptingStorage(
-            adapter: environmentAdapter,
+            adapter: adapter,
             child: child
         )
     }
@@ -192,17 +212,18 @@ extension ElementContent {
 
 
 fileprivate protocol ContentStorage {
+    
     var childCount: Int { get }
 
     func measure(
-        in constraint: SizeConstraint,
-        environment: Environment,
+        in constraint : SizeConstraint,
+        with context : LayoutContext,
         cache: CacheTree
     ) -> CGSize
 
     func performLayout(
-        attributes: LayoutAttributes,
-        environment: Environment,
+        in size: CGSize,
+        with context : LayoutContext,
         cache: CacheTree
     ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)]
 }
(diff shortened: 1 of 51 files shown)


BlueprintUI/Sources/Element/MeasurementCache.swift
```diff
@@ -28,7 +28,7 @@ final class MeasurementCache
         
         let innerKey = Key(
             key: key,
-            sizeConstraint: sizeConstraint
+            sizeConstraint: context.constraint // TODO: Should probably include env in some way. Or we just rip out this caching? Or we include depth?
```
kyleve, line 31: Figure out an answer to this Q

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -143,9 +143,9 @@ private final class UIViewElementMeasurer {
     static let shared = UIViewElementMeasurer()
```
kyleve, line 143: Pass the `environment` through in this object.

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -64,24 +64,28 @@ extension EqualStack {
         var direction: Direction
         var spacing: CGFloat
 
-        func measure(in constraint: SizeConstraint, items: [(traits: Void, content: Measurable)]) -> CGSize {
+        func measure(
```
kyleve, line 67: Change it to measure(with:`?

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -123,7 +123,7 @@ public extension UIViewElement {
             }
```
kyleve, line 123: Now that we can easily flow stuff through to `measurement`, it would be good to remove the static view cache and put it on each blueprint view instead.

BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -123,7 +123,7 @@ public extension UIViewElement {
...
             
             config.apply { view in
-                self.updateUIView(view, with: .init(isMeasuring: false))
+                self.updateUIView(view, with: .init(isMeasuring: false, environment: context.environment))
```
kyleve, line 126: Didn't push it yet, but another clever way to make useView have a shorter call site is to use #file and #line to guarantee callsite uniqueness

BlueprintUI/Sources/Internal/Logger.swift
```diff
@@ -7,6 +7,9 @@ enum Logger { }
 /// BlueprintView signposts
 extension Logger {
```
kyleve, line 8: Despite even turning off logging with `BlueprintLogging.enabled`, calling these funcs still has a cost. So, added these short circuits.

BlueprintUI/Sources/Layout/Decorate.swift
```diff
@@ -65,19 +65,19 @@ public struct Decorate : ProxyElement {
...
-            LayoutWriter { context, layout in
+            LayoutWriter { constraint, context, layout in
                 
                 let contentSize = self.wrapped.content.measure(
-                    in: context.size,
-                    environment: environment
-                )
+                    in: constraint,
+                    with: context
+                ) // TODO: This doesn't pass caching through...
```
kyleve, line 73: No worse than before; probably fine?

BlueprintUI/Tests/PerformancePlayground.swift
```diff
@@ -95,7 +95,8 @@ class PerformancePlayground : XCTestCase
                 return Column { column in
```
kyleve, line 95: Revert this file before merging / break the changes into their own test func

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -15,7 +15,11 @@ public protocol Layout {
     ///   object and a `Measurable` value.
```
kyleve, line 15: Need to fix API comments in this file

SampleApp/SampleApp.xcodeproj/xcshareddata/xcschemes/SampleApp.xcscheme
```diff
@@ -79,7 +79,7 @@
       </BuildActionEntries>
```
kyleve, line 79: Revert before merge

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -55,13 +64,14 @@ public struct ElementContent {
...
-        attributes: LayoutAttributes,
-        environment: Environment,
+        in size: CGSize,
+        with context : LayoutContext,
         cache: CacheTree
-    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
-        return storage.performLayout(
-            attributes: attributes,
-            environment: environment,
+    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] // TODO: Turn this into a reference type too
```
kyleve, line 70: Remove this TODO

BlueprintUI/Sources/Element/LayoutContext.swift
```diff
@@ -0,0 +1,125 @@
...
+//
+//  Created by Kyle Van Essen on 7/2/21.
+//
+
+import Foundation
+
+
+///
+///
+public struct LayoutContext {
```
kyleve, line 13: Finish docs

BlueprintUI/Sources/Layout/Decorate.swift
kyleve: Remove this comment, need to figure out in next PR

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -26,17 +30,46 @@ public protocol Layout {
...
+
 extension Layout where Traits == () {
     
     public static var defaultTraits: () {
         return ()
     }
+}
+
+
+public final class LayoutItems<Traits> {
```
kyleve, line 53: Docs

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -26,17 +30,46 @@ public protocol Layout {
...
+    public let all : [Item]
+    
+    public let count : Int
+    
+    init(with all : [Item]) {
+        self.all = all
+        self.count = self.all.count
+    }
     
+    public struct Item {
```
kyleve, line 64: Docs

BlueprintUI/Sources/Layout/LayoutWriter.swift
```diff
@@ -47,9 +47,9 @@ public struct LayoutWriter : Element {
     //
     
     public var content: ElementContent {
-        ElementContent { size, env in
+        ElementContent { constraint, context -> Element in
```
kyleve, line 50: Do I need `-> Element` here?

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -440,13 +440,25 @@ private struct TestContainer: Element {
...
-    private class TestLayout: Layout {
-        func measure(in constraint: SizeConstraint, items: [(traits: (), content: Measurable)]) -> CGSize {
-            return .zero
+    private struct TestLayout: Layout {
+        
+        func measure(
+            items: LayoutItems<Void>,
+            in constraint : SizeConstraint,
+            with context: LayoutContext
+
```
kyleve, line 449: Bonus newline

BlueprintUI/Sources/Element/LayoutContext.swift
```diff
@@ -0,0 +1,138 @@
...
+    
+    let measurementCache : MeasurementCache
+    let measurementViews : MeasurementViews
+}
+
+
+extension LayoutContext {
+    
+    /// Internal cache used to retain prototype views for measurement operations.
+    final class MeasurementViews {
```
kyleve, line 109: Tests for this

BlueprintUI/Sources/Element/MeasurementCachingKey.swift
```diff
@@ -19,9 +19,8 @@ import Foundation
 ///
```
kyleve, line 19: These docs were wrong before.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -135,17 +152,20 @@ extension ElementContent {
     /// Initializes a new `ElementContent` with no children that delegates to the provided measure function.
     ///
     /// - parameter measurementCachingKey: An optional key to use to cache measurement. See the `MeasurementCachingKey` documentation for more.
-    /// - parameter measureFunction: How to measure the `ElementContent` in the given `SizeConstraint`.
+    /// - parameter measureFunction: How to measure the `ElementContent` in the given `SizeConstraint` and `LayoutContext`.
     public init(
```
kyleve, line 156: While in here; it might be nice to swap these out with static funcs that make it a bit easier to get to the init you want. Eg, I always end up searching through autocomplete for the one init I want, whereas it'd be a lot easier to find the one you want if we switched to:

```
var content : ElementContent {
    .byMeasuring { constraint, context in
        ....
    }
}

var content : ElementContent {
    .withChild { constraint, context in
        ....
    }
}
```

watt:
I don't have time for a full review right now, but as I mentioned before, I am not in favor of this approach for a number of reasons. The most objective one is that "environment everywhere" encourages out-of-band measurements that can't be cached.

kyleve:
It doesn't, though – this is a valid, existing use case throughout both Market, and SPOS – people need to measure content that lives in a nested view, usually a Blueprint view. Right now, they're almost always universally using the _wrong_ environment: An empty one. This is going to break things in weird ways, because the failure is silent / fails open – measurements will end up being wrong in subtle ways as the default env value and correct env value diverge. That's going to be very painful to debug and run down. By shaping the APIs in a way that fits together nicely, it's much more likely people will do the right thing because the two ends fit together logically.

In terms of not caching measurements, that's an improvement we can address – and I've started to think about that [here](https://github.com/square/Blueprint/pull/239) – but it is _already_ the case, today, with the second punch of the aforementioned incorrect environment being used.

watt:
> this is a valid, existing use case throughout both Market, and SPOS – people need to measure content that lives in a nested view, usually a Blueprint view.

This is not a good solution for that use case. There is a much smaller, simpler solution for that that doesn't expose environment everywhere.

> This is going to break things in weird ways

I agree, but I think the solution is _less_ API, not more. We can solve that case without creating more ways to do the wrong thing.

kyleve:
That feels a bit like a "and" thing; not an "or" thing – generally, just having the Environment everywhere is very desirable to me; since it is the transport layer for agnostic data around the blueprint infrastructure – eg imagine if my measurement or layout was dependent on an environment value that is injected higher up the "stack" so it doesn't need to be manually passed down several layers. Now that you can get to the env from your view description, it seems weird to _not_ be able to also easily get to it in the measurement and layout passes as well.

And again, we already have a lot of code that measures a nested element (as well as GeometryProxy, LayoutWriter) – getting that code an Environment without having to do major surgery/refactoring to correct those measurements seems desirable if we can get it with a mechanical refactor instead, which is what this unlocks (aka basically changing method signatures) – I took a stab at it last night and I (think?) it was basically done in about an hour.

kyleve:
Added something similar to what you described; so we can also migrate things over to that if it's easily as mechanical as well! `ElementContent(byMeasuring:someChild)`

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -191,18 +211,68 @@ extension ElementContent {
 }
 
 
+extension ElementContent {
```
kyleve, line 214: Tagging for a link

n8chur (commented):
This PR is fairly difficult to review as it's very large and encompasses multiple changes that could likely become independent changes (context object, logging improvements, iterator performance optimizations, view caching changes, etc.).

Could we break this PR up into a few smaller PRs?

PR #240, opened 2021-07-21
Add element content that measures with environment

kylebshr:


kyleve:
I think we have this now!

PR #241, opened 2021-07-29
fixed an issue were labels weren't accessible by default

RoyalPineapple:
Labels should almost always be accessibility elements. 
In a situation where one would want to explicitly set `isAccessabilityElement` to `false`  the `AccessibilityBlocker` would suffice.

### DIFF
--- a/BlueprintUICommonControls/Sources/Label.swift
+++ b/BlueprintUICommonControls/Sources/Label.swift
@@ -14,7 +14,6 @@ public struct Label: ProxyElement {
     public var numberOfLines: Int = 0
     public var lineBreakMode: NSLineBreakMode = .byWordWrapping
     public var lineHeight: LineHeight = .font
-    public var isAccessibilityElement = false
 
     public init(text: String, configure: (inout Label) -> Void = { _ in }) {
         self.text = text
@@ -57,7 +56,6 @@ public struct Label: ProxyElement {
     public var elementRepresentation: Element {
         AttributedLabel(attributedText: attributedText) { label in
             label.numberOfLines = numberOfLines
-            label.isAccessibilityElement = isAccessibilityElement
 
             switch lineHeight {
             case .custom(let lineHeight, .top):
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -5,7 +5,6 @@ public struct AttributedLabel: Element, Hashable {
 
     public var attributedText: NSAttributedString
     public var numberOfLines: Int = 0
-    public var isAccessibilityElement = false
 
     /// An offset that will be applied to the rect used by `drawText(in:)`.
     ///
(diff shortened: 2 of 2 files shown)


BlueprintUICommonControls/Sources/Label.swift
```diff
@@ -14,7 +14,6 @@ public struct Label: ProxyElement {
     public var numberOfLines: Int = 0
     public var lineBreakMode: NSLineBreakMode = .byWordWrapping
     public var lineHeight: LineHeight = .font
-    public var isAccessibilityElement = false
```
kyleve, line 17: IMO it might be nice to still expose this over forcing the use of accessibility blocker; but curious how others feel.

kyleve (approved):

BlueprintUICommonControls/Sources/Label.swift
kylebshr: I agree, but if @RoyalPineapple feels strongly that this shouldn't really be used ever then I'm fine with it

kylebshr (approved):

BlueprintUICommonControls/Sources/Label.swift
RoyalPineapple: I kinda do feel strongly. Any situation where you'd want a label not to be accessible should be pretty few and far between. 

My thoughts being that If you've got something you're trying to show non-accessibility users, its likely equally relevant to accessibility users, no? In that case I think a little friction is appropriate. 

It might be nice to promote accessibility properties to all blueprint elements directly and remove the use of the wrapper, but in that case i'd prefer to model the Element/Container distinction with an enum rather than this silly bool.

kyleve: Makes sense; fair enough! Can you do a PR to update the changelog in the root of the repo as well? (Sorry, I forgot to call this out!)

RoyalPineapple: Oops sorry. It wasn't in the checklist so I presumed there wasn't one. 
Should I bump the version as well?

kyleve: We'll do the version on release!

kyleve: Hmm, quite a few things use this – IMO it's a pretty cheap API to keep around; so I think it makes sense to bring this back. It's a lot more ergonomic to read than having to wrap in `blockAccessibility()`, IMO.

kyleve: (In particular, Market uses this frequently to disable accessibility on labels which are part of larger components like buttons, etc)

RoyalPineapple: I'm fine with it if this is already widely used, I don't want to force a major refactor. But you shouldn't need to turn accessibility of if the containing view is itself an accessibility element.

I'll look into that more in particular when I return.

PR #244, opened 2021-08-05
Lifecycle callbacks

### LINKED ISSUE #48: Appeared/disappeared callbacks
`ViewDescription` should provide a way to run a closure when the view appears or disappears. Could possibly be a property on the `VisibilityTransition`.

SwiftUI provides `onAppear` (which runs before appearance animations) and `onDisappear` (which runs after). We may want to provide separate hooks for the start & end of animations.

watt:
This PR introduces 4 lifecycle callbacks. Here's a state diagram:

```
          ┌───onMount───┐   ┌──onAppear───┐
          │             │   │             │
  ┌───────┴───┐     ┌───▼───┴───┐     ┌───▼───────┐
  │ Unmounted │     │  Mounted  │     │  Visible  │
  └───────▲───┘     └───┬───▲───┘     └───┬───────┘
          │             │   │             │
          └──onUnmount──┘   └─onDisappear─┘
```

"Mounting" refers to creating a backing view and adding it as a subview to the hosting `BlueprintView` as part of a render pass. This is distinct from appearing: an element may or may not become visible immediately after mounting, and it may disappear & reappear without unmounting.

These are also distinct from the appearing/disappearing _transitions_, which currently run during mounting & unmounting regardless of whether the hosting view is visible (and in the case of appearance, are suppressed when nested in another appearance transition). We could probably update transitions to match this lifecycle, but I didn't attempt it in this PR.

There's more documentation on the `LifecycleCallback` type.

Fixes #48

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -141,13 +141,6 @@ public final class BlueprintView: UIView {
         fatalError("init(coder:) has not been implemented")
     }
 
-    deinit {
-        // ensure final unmount is called if there's still a mounted element tree
-        rootController.traverse { node in
-            node.onUnmount?()
-        }
-    }
-
     ///
     /// Measures the size needed to display the view within then given constraining size,
     /// by measuring the current `element` of the `BlueprintView`.
@@ -410,14 +403,6 @@ extension BlueprintView {
             viewDescription.onDisappear
         }
 
-        var onMount: LifecycleCallback? {
-            viewDescription.onMount
-        }
-
-        var onUnmount: LifecycleCallback? {
-            viewDescription.onUnmount
-        }
-
         let view: UIView
 
         init(node: NativeViewNode) {
@@ -525,7 +510,6 @@ extension BlueprintView {
 
                         contentView.insertSubview(controller.view, at: index)
 
-                        controller.onMount?()
                         if context.viewIsVisible {
                             controller.onAppear?()
                         }
@@ -549,11 +533,10 @@ extension BlueprintView {
 
             for controller in oldChildren.values {
                 func removeChild() {
-                    controller.traverse { node in
-                        if context.viewIsVisible {
+                    if context.viewIsVisible {
+                        controller.traverse { node in
                             node.onDisappear?()
                         }
-                        node.onUnmount?()
                     }
                     controller.view.removeFromSuperview()
                 }
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -91,6 +91,18 @@ public final class BlueprintView: UIView {
     /// Provides performance metrics about the duration of layouts, updates, etc.
     public weak var metricsDelegate: BlueprintViewMetricsDelegate? = nil
 
+    private var isVisible: Bool = false {
+        didSet {
+            switch (oldValue, isVisible) {
+            case (false, true):
+                handleAppeared()
+            case (true, false):
+                handleDisappeared()
+            default: break
+            }
+        }
+    }
+
     /// Instantiates a view with the given element
     ///
     /// - parameter element: The root element that will be displayed in the view.
@@ -127,6 +139,13 @@ public final class BlueprintView: UIView {
     @available(*, unavailable)
     public required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
+    }
+
+    deinit {
+        // ensure final unmount is called if there's still a mounted element tree
+        rootController.traverse { node in
+            node.onUnmount?()
+        }
     }
 
     ///
@@ -220,6 +239,7 @@ public final class BlueprintView: UIView {
 
     public override func didMoveToWindow() {
         super.didMoveToWindow()
+        isVisible = (window != nil)
         setNeedsViewHierarchyUpdate()
     }
 
@@ -277,7 +297,8 @@ public final class BlueprintView: UIView {
         rootController.update(
             node: rootNode,
             context: .init(
-                appearanceTransitionsEnabled: hasUpdatedViewHierarchy
+                appearanceTransitionsEnabled: hasUpdatedViewHierarchy,
+                viewIsVisible: isVisible
             )
         )
 
@@ -338,6 +359,18 @@ public final class BlueprintView: UIView {
 
         return environment
     }
+
+    private func handleAppeared() {
+        rootController.traverse { node in
+            node.onAppear?()
+        }
+    }
+
+    private func handleDisappeared() {
+        rootController.traverse { node in
+            node.onDisappear?()
+        }
+    }
 }
 
 
@@ -368,6 +401,22 @@ extension BlueprintView {
         private var layoutAttributes: LayoutAttributes
 
         private(set) var children: [(ElementPath, NativeViewController)]
+
+        var onAppear: LifecycleCallback? {
+            viewDescription.onAppear
+        }
+
+        var onDisappear: LifecycleCallback? {
+            viewDescription.onDisappear
+        }
+
+        var onMount: LifecycleCallback? {
+            viewDescription.onMount
+        }
+
+        var onUnmount: LifecycleCallback? {
+            viewDescription.onUnmount
+        }
 
         let view: UIView
 
@@ -476,6 +525,11 @@ extension BlueprintView {
 
                         contentView.insertSubview(controller.view, at: index)
 
+                        controller.onMount?()
+                        if context.viewIsVisible {
+                            controller.onAppear?()
+                        }
+
                         controller.update(node: child, context: context.modified {
                             $0.appearanceTransitionsEnabled = false
                         })
@@ -494,16 +548,36 @@ extension BlueprintView {
             }
 
             for controller in oldChildren.values {
-                if let transition = controller.viewDescription.disappearingTransition {
-                    transition.performDisappearing(view: controller.view, layoutAttributes: controller.layoutAttributes, completion: {
-                        controller.view.removeFromSuperview()
-                    })
-                } else {
+                func removeChild() {
+                    controller.traverse { node in
+                        if context.viewIsVisible {
+                            node.onDisappear?()
+                        }
+                        node.onUnmount?()
+                    }
                     controller.view.removeFromSuperview()
+                }
+
+                if let transition = controller.viewDescription.disappearingTransition {
+                    transition.performDisappearing(
+                        view: controller.view,
+                        layoutAttributes: controller.layoutAttributes,
+                        completion: removeChild
+                    )
+                } else {
+                    removeChild()
                 }
             }
 
             children = newChildren
+        }
+
+        /// Perform a depth-first traversal of the view-backing tree from this node.
+        func traverse(visitor: (NativeViewController) -> Void) {
+            visitor(self)
+            for (_, child) in children {
+                child.traverse(visitor: visitor)
+            }
         }
     }
 }
@@ -516,6 +590,9 @@ extension BlueprintView.NativeViewController {
 
         /// If appearance transitions are enabled for insertions and removals.
         var appearanceTransitionsEnabled: Bool
+
+        /// True if the hosting view is in the view hierarchy
+        var viewIsVisible: Bool
 
         /// Returns a copy of the update context, modified by the changes provided.
         func modified(_ modify: (inout Self) -> Void) -> Self {
--- a/BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
+++ b/BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
@@ -1,0 +1,44 @@
+import Foundation
+
+/// This is the type used by all lifecycle callback hooks.
+///
+/// - Tag: ElementLifecycle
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [The `Environment` is now automatically propagated through to nested `BlueprintViews` within a displayed `Element` hierarchy](https://github.com/square/Blueprint/pull/234). This means that if your view-backed `Elements` themselves contain a `BlueprintView` (eg to manage their own state), that nested view will now automatically receive the correct `Environment` across `BlueprintView` boundaries. If you were previously manually propagating `Environment` values you may remove this code. If you would like to opt-out of this behavior; you can set `view.automaticallyInheritsEnvironmentFromContainingBlueprintViews = false` on your `BlueprintView`.
 
+- [Lifecycle hooks][#244]. You can hook into 4 lifecycle events:
+  - `onMount` occurs when a backing view has been created and attached to its hosting view's view hierarchy
+  - `onAppear` occurs when a mounted element becomes visible
+  - `onDisappear` occurs when a visible element is no longer visible
+  - `onUnmount` occurs when a mounted element is removed from the hosting view's view hierarchy
+
+  Each hook is can be added with a modifier of the same name:
```
kylebshr, line 24:
```suggestion
  Each hook can be added with a modifier of the same name:
```

kylebshr (approved):
Very elegant, nice work!

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -220,6 +232,7 @@ public final class BlueprintView: UIView {
 
     public override func didMoveToWindow() {
         super.didMoveToWindow()
+        isVisible = (window != nil)
```
kyleve, line 235: I think we'll need a couple more checks here; also including the hidden-ness of the view and window, perhaps

kyleve: (Also, if any views upwards in the hierarchy are hidden this will be "wrong", I think? Eg trying to think of cases where this would fall through. I wonder a more robust approach would be to keep a record of if we were previously visible, and then if that changes during a layout subviews (eg compute the effective hidden-ness of the view again); perform the appropriate onAppear vs onDisappear in addition to when we're _removed_ from a window)

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
```diff
@@ -0,0 +1,46 @@
...
+///
+/// # The Element Lifecycle
+///
+/// View-backed elements have a lifecycle with 3 states:
+///
+/// 1. **Unmounted**
+///
+///    This is the initial state of an element, before a render pass.
+///
+/// 2. **Mounted**
```
kyleve, line 15: Maybe just me; but I don't love the mounted/unmounted terminology; maybe just `new/loaded/visible` or something?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -91,6 +91,18 @@ public final class BlueprintView: UIView {
     /// Provides performance metrics about the duration of layouts, updates, etc.
     public weak var metricsDelegate : BlueprintViewMetricsDelegate? = nil
 
+    private var isVisible: Bool = false {
```
kyleve, line 94: I wonder if we want to add a `delegate` to allow more deeply integrating this into view controller hierarchies and their view transitions as well; eg to also allow a view controller to control visibility based on its state transitions; or to nest appearance transitions inside existing animation blocks, etc..

BlueprintUICommonControls/Sources/LifecycleObserver.swift
```diff
@@ -0,0 +1,173 @@
...
+            onAppear: onAppear,
+            onDisappear: onDisappear,
+            onMount: onMount,
+            onUnmount: onUnmount + callback,
+            wrapping: wrapped
+        )
+    }
+}
+
+extension Element {
```
kyleve, line 118: Might be also nice to add this version?

```
func observe(
        onAppear: LifecycleCallback? = nil,
        onDisappear: LifecycleCallback? = nil,
        onMount: LifecycleCallback? = nil,
        onUnmount: LifecycleCallback? = nil,
) -> LifecycleObserver
```

n8chur: or
```swift
func on(
    mount: LifecycleCallback? = nil,
    appear: LifecycleCallback? = nil,
    disappear: LifecycleCallback? = nil,
    unmount: LifecycleCallback? = nil
) -> LifecycleObserver
```

@kyleve are you suggesting this from a performance perspective (less nested elements)?

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
n8chur, line 15: When would hooking into the mount/unmount events be useful in practice?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
watt: > I think we'll need a couple more checks here; also including the hidden-ness of the view and window, perhaps

> (Also, if any views upwards in the hierarchy are hidden this will be "wrong", I think? Eg trying to think of cases where this would fall through. 

The idea of "visible" can get complex. What if a view is obscured? What if it has a very low alpha? I think it's OK for us to pick a definition, and not worry about chasing down every interpretation.

So, I could see checking this view's `isHidden`, but walking the ancestor chain seems like too big a lift, because we can't observe it. FWIW, SwiftUI's idea of visibility is not affected by the `hidden()` modifier.

> I wonder a more robust approach would be to keep a record of if we were previously visible, and then if that changes during a layout subviews (eg compute the effective hidden-ness of the view again); perform the appropriate onAppear vs onDisappear in addition to when we're _removed_ from a window)

I think this does what you're describing already? It tracks visibility and perform the callbacks whenever `isVisible` flips.

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
watt: > Maybe just me; but I don't love the mounted/unmounted terminology; maybe just `new/loaded/visible` or something?

Borrowed from React. We're going to have to define exactly what it means either way, I think.

watt: > When would hooking into the mount/unmount events be useful in practice?

I don't have any uses in mind! Mount/unmount were a lot easier to write than appear/disappear, and I left them in for completeness.

BlueprintUICommonControls/Sources/LifecycleObserver.swift
watt: Because of the extensions on `LifecycleObserver`, calling

```swift
element
  .onAppear { }
  .onDisappear { }
  .onMount { }
  .onUnmount { }
```

Will coalesce to a single `LifecycleObserver`.

kyleve: Mostly just since it makes the element tree cleaner (especially if debugging); but nesting would have a non-zero cost as well.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
watt: > I wonder if we want to add a `delegate` to allow more deeply integrating this into view controller hierarchies and their view transitions as well; eg to also allow a view controller to control visibility based on its state transitions; or to nest appearance transitions inside existing animation blocks, etc..

I'm not 100% clear on what you have in mind, but I don't think this should be publicly writable. Every additional knob like that we add makes the internal state harder to manage.

n8chur (approved):
I was going to suggest we remove mount/unmount if we don't have use-cases in mind since it wasn't immediately obvious to me at first, but hearing that it's borrowed from React makes me feel better about it!

BlueprintUICommonControls/Sources/LifecycleObserver.swift
kyleve: (Oops, just saw Andrew and I replied at the same time)

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
kyleve: I guess in general I don't like things like "mount" or "render" since they're not common terms in iOS land at all – maybe some naming we could borrow from SwiftUI or something?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Hmm ok, sounds good. I would maybe add the `hidden` check. I think maybe we'll see if adding additional knobs for sharp / unexpected edges is worth it based on how people use it!

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: I was mostly thinking about: My blueprint view is hosted as the view of a view controller; it might be nice to tie "visibility" to the visibility lifecycle of that VC's viewWillAppear, etc.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kylebshr: I personally don't like adding a `hidden` check, that property doesn't affect anything in the view hierarchy, and there's no way to be notified about it. I feel like it would be unexpected behavior if that affected these events. Maybe that's just me?

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
kyleve: (Maybe `create` and `teardown` or similar?)

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
```diff
@@ -0,0 +1,46 @@
...
+///           ┌───onMount───┐   ┌──onAppear───┐
+///           │             │   │             │
+///   ┌───────┴───┐     ┌───▼───┴───┐     ┌───▼───────┐
+///   │ Unmounted │     │  Mounted  │     │  Visible  │
+///   └───────▲───┘     └───┬───▲───┘     └───┬───────┘
+///           │             │   │             │
+///           └──onUnmount──┘   └─onDisappear─┘
+/// ```
+///
+/// - Note: If a hosting view is deallocated while it contains mounted elements, those elements will
```
n8chur, line 43: Could/should we call `onUnmount` in the `BlueprintView`'s `deinit` (if it isn't already in the unmounted state)?

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
watt: > I guess in general I don't like things like "mount" or "render" since they're not common terms in iOS land at all – maybe some naming we could borrow from SwiftUI or something?

SwiftUI doesn't have an analogous concept to mounting. Not sure if that's because of differences in the layout system, or if it's just something that's not exposed.

I'm happy to use existing terminology when it exists, but I don't think it's helpful to repurpose familiar terms for unfamiliar concepts.

If other folks find the mount/unmount idea confusing, perhaps we should not expose these events, or the mounted state, at all? Personally, I think that since the appear/disappear events do not directly correspond with views being created/added/removed/released, that middle state of "mounted" is useful for representing the difference, but maybe it's just overcomplicating the API?

BlueprintUI/Sources/ViewDescription/LifecycleCallback.swift
watt: > Could/should we call `onUnmount` in the `BlueprintView`'s `deinit` (if it isn't already in the unmounted state)?

Hmm, yeah, I think that's doable, if we want to guarantee  that final unmount.

kylebshr: I agree, it seems nice to guarantee it

watt: OK, done.

kyleve: image

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: 👍

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: 👍

bencochran (approved):

CHANGELOG.md
kylebshr, line 24:
```suggestion
  Each hook can be added with a modifier of the same name:
```

this never got fixed

n8chur (approved):

kyleve (approved):

CHANGELOG.md
watt: Oh geez, I think I accepted the suggested change on the web and then blew it away with a force push. 🤦

PR #245, opened 2021-08-06
Don't draw the view of PassthroughView

kylebshr:
This is a performance optimization trick that prevents unnecessary drawing of the `PassthroughView`, which is used as the backing view for elements that shouldn't actually affect the view hierarchy.


kyleve (approved):
Seems fine; but is this still actually a performance benefit? I recall eg UIStackView recently moved away from this on newer OSes, so not sure we still need this; or if UIView is smart enough to know its basically just a container.

watt (approved):

kyleve:
👋 Do we still want this? It doesn't seem to hurt, but also does it help?

kylebshr:
I feel like why not? Could still be a perf win on older versions

kyleve (approved):

PR #247, opened 2021-08-07
Specific rounded corners for Box.CornerStyle.rounded

gabreho:
#### Specific rounded corners for Box

Usage
``` swift
Box(
    backgroundColor: .black, 
    cornerStyle: .rounded(radius: 17, corners: .right)
)
...
element
    .box(background: .systemRed, corners:  .rounded(radius: 8, corners: .topLeft)
```

Screen Shot 2021-08-06 at 9 30 48 PM


kylebshr (approved):
This looks great! Would you mind add a couple snapshot tests?

kyleve:
> This looks great! Would you mind add a couple snapshot tests?

So fun fact about this! It won't work with our current snapshot setup. I ran into the same thing over here: https://github.com/square/Blueprint/pull/166, and then promptly forgot about it. Basically, the rendering method we use in our snapshot tests don't support all the modern APIs; we need to swap them out and add our tests to an app host to resolve:

> The OS X v10.5 implementation of this method does not support the entire Core Animation composition model. QCCompositionLayer, CAOpenGLLayer, and QTMovieLayer layers are not rendered. Additionally, layers that use 3D transforms are not rendered, nor are layers that specify backgroundFilters, filters, compositingFilter, or a mask values. Future versions of macOS may add support for rendering these layers and properties.

kyleve (approved):

PR #248, opened 2021-08-08
Introduce onDidLayoutContents on BlueprintView

kyleve:
Add an `onDidLayoutContents` callback on `BlueprintView` which is invoked when a layout occurs. We'll use this in `MarketViewController/MarketScreen/etc` to update the `preferredContentSize` to avoid repeated redundant calculations from workflow-based updates, or other updates which are sent more than once per layout pass.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -90,7 +90,10 @@ public final class BlueprintView: UIView {
     
     /// Provides performance metrics about the duration of layouts, updates, etc.
     public weak var metricsDelegate : BlueprintViewMetricsDelegate? = nil
-
+    
+    /// Called after the view lays out its content, as a result of a new element, a bounds change, etc.
+    public var onDidLayoutContents : (() -> ())? = nil
+    
     /// Instantiates a view with the given element
     ///
     /// - parameter element: The root element that will be displayed in the view.
@@ -277,6 +280,8 @@ public final class BlueprintView: UIView {
                 appearanceTransitionsEnabled: hasUpdatedViewHierarchy
             )
         )
+        
+        self.onDidLayoutContents?()
 
         Logger.logViewUpdateEnd(view: self)
         let viewUpdateEndDate = Date()
--- a/BlueprintUI/Tests/AdaptedEnvironmentTests.swift
+++ b/BlueprintUI/Tests/AdaptedEnvironmentTests.swift
@@ -21,7 +21,7 @@ class AdaptedEnvironmentTests : XCTestCase {
         
         view.layoutIfNeeded()
         
-        XCTAssertEqual(environment?[TestingKey1], "adapted1")
+        XCTAssertEqual(environment?[TestingKey1.self], "adapted1")
     }
     
     func test_wrapping_multiple() {
@@ -41,8 +41,8 @@ class AdaptedEnvironmentTests : XCTestCase {
         view.layoutIfNeeded()
         
         // The inner-most change; the one closest to the element; should be the value we get.
-        XCTAssertEqual(environment?[TestingKey1], "adapted1.1")
-        XCTAssertEqual(environment?[TestingKey2], "adapted2.1")
+        XCTAssertEqual(environment?[TestingKey1.self], "adapted1.1")
+        XCTAssertEqual(environment?[TestingKey2.self], "adapted2.1")
         
         // Ensure we collapsed the AdaptedEnvironments down to one level of wrapping.
         XCTAssertTrue((element as? AdaptedEnvironment)?.wrapped is TestElement)
--- a/BlueprintUI/Tests/EnvironmentTests.swift
+++ b/BlueprintUI/Tests/EnvironmentTests.swift
@@ -148,19 +148,19 @@ class EnvironmentTests: XCTestCase {
     
     func test_merged() {
         var lhs = Environment.empty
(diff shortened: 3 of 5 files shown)


BlueprintUI/Tests/AdaptedEnvironmentTests.swift
```diff
@@ -21,7 +21,7 @@ class AdaptedEnvironmentTests : XCTestCase {
         
         view.layoutIfNeeded()
         
-        XCTAssertEqual(environment?[TestingKey1], "adapted1")
+        XCTAssertEqual(environment?[TestingKey1.self], "adapted1")
```
kyleve, line 24: Xcode 12 warns about this; so fixed it.

BlueprintUI/Tests/EnvironmentTests.swift
```diff
@@ -148,19 +148,19 @@ class EnvironmentTests: XCTestCase {
     
     func test_merged() {
         var lhs = Environment.empty
-        lhs[TestKey] = .wrong
-        lhs[LHSTestKey] = 2
+        lhs[TestKey.self] = .wrong
```
kyleve, line 151: Xcode 12 warns about this; so fixed it.

n8chur (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -277,6 +280,8 @@ public final class BlueprintView: UIView {
                 appearanceTransitionsEnabled: hasUpdatedViewHierarchy
             )
         )
+        
+        self.onDidLayoutContents?()
```
watt, line 284: Can you move this all the way to the bottom, after the flags are reset and after signpost logging? Or use `defer`?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -90,7 +90,10 @@ public final class BlueprintView: UIView {
     
     /// Provides performance metrics about the duration of layouts, updates, etc.
     public weak var metricsDelegate : BlueprintViewMetricsDelegate? = nil
-
+    
+    /// Called after the view lays out its content, as a result of a new element, a bounds change, etc.
+    public var onDidLayoutContents : (() -> ())? = nil
```
watt, line 95: Based on when this gets called, `onDidRender` might be more precise.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Ahh good call

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: I didn't like "render" since its not really a common iOS term. `onDidUpdateContents` or something?

BlueprintUI/Tests/AdaptedEnvironmentTests.swift
bencochran: lol I wonder how that ever compiled

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
bencochran: `onDidLayoutContents` sounds a bit like it'll only happen on layout (from like a bounds change, etc). `onDidUpdateContents` sounds a bit like it'll only happen on a new element.

So, yeah. I don't have a strong opinion between those two, but lean toward `onDidLayoutContents` I think. To add a new bit of bikeshed: I could also go with `onDidLayoutElement` since it's the term of art here.

bencochran (approved):

BlueprintUI/Tests/AdaptedEnvironmentTests.swift
kyleve: It seems like it was supported; and now they're pushing away from it? It doesn't warn at all in Xcode 12 AFAIK... it also only (weirdly) works in subscript syntax; it doesn't work in func syntax.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
watt: "Render" is a term of art, but that's because have a concept here (the render pass) that doesn't exist in UIKit. I don't think it's necessarily better to bend an existing term if it's not accurate.

kylebshr: Yeah I personally I don't mind using `render` here because it's a blueprint specific meaning. There's already several uses of `render` in this repo to describe how elements are, well, rendered.

kyleve: I think maybe just `onDidLayout` works best here?

I still don't like "render" since it's not really an existing term here; but the layout is tied to the layout pass...

kylebshr: That's why I'm in favor of `render` - `layout` _is_ an existing term, and while the update happens to use the layout pass to coalesce renders, we're trying to communicate something distance from `didLayoutSubviews`. I'm in favor of moving forward and not wasting time, so I'm fine with `onDidLayout` but I do prefer `onDidRender` and I think it's clear what it means if you use Blueprint

kyleve:
Size caching should help with this as well, going to go that route rather than add this API

PR #249, opened 2021-08-14
Add `Equatable` to `ConstrainedSize.Constraint`

cbowns:


BlueprintUI/Sources/Layout/ConstrainedSize.swift
```diff
@@ -82,6 +82,23 @@ extension ConstrainedSize {
                 return absoluteValue
             }
         }
+
+        public static func == (lhs: Self, rhs: Self) -> Bool {
```
kyleve, line 86: Do you need this? Shouldn't the compiler synthesize it automatically since all the cases contain equatable values?

cbowns: Oh hmm. I got endless complaints when adding it via an extension in SPOS but I didn't try here to see if it differed. (I didn't think Equatable provided automatic conformance when you had associated values)

cbowns: Yup works great. Weird.

kyleve (approved):

PR #250, opened 2021-08-15
Deprecate iOS 11

kyleve:


BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -47,7 +47,7 @@ public struct Box: Element {
                 if self.backgroundColor != view.backgroundColor {
```
kyleve, line 47: Looks like there was a merge race with us landing the formatting on the changes in this file vs the corner rounding changes – so ran the formatter; otherwise CI fails.

kylebshr (approved):

PR #253, opened 2021-08-24
Round sizes to nearest pixel during measurement

kylebshr:
This fixes some cases of floating point precision loss by rounding every elements size to the nearest pixel. 

For example, a row with two labels rendered on a 3x device and then inset has the following issue.

During measurement, this is what happens:

- The labels measure to x.33333333333334 and x.66666666666669
- The row adds those two elements widths while measuring, which comes to x.00000000000003
- The inset takes the width and adds 24 twice. This addition causes the precision to get lost due to how the floating point numbers are represented, and the inset returns a width of, e.g., 212.0

Now, the size of 212.0 is ultimately used by Blueprint during the _layout_ pass. This means that, going down the tree:

- The inset is provided a constraint of 212.0
- It removes the inset when measuring the child, so measures it with a constraint of 164.0
- 💥 suddenly, the row was measured with a width of `x.00000000003` but laid out with a width of `x.0`

This disparate behavior can cause issues in, e.g., a GeometryReader. The reader might be returning two different elements if it decides which element to return based on its available width. If the element it returns during measurement is an element like this row, which is then wrapped in an inset, the geometry proxy will be laid out with a smaller width than the measured element, and end up returning the _other_ element; which results in some very weird behavior.

Register is green with these changes, other than needed to update a few (surprisingly few, to me) snapshot tests: https://github.com/squareup/ios-register/pull/59882


BlueprintUI/Sources/Internal/Logger.swift
```diff
@@ -7,79 +7,93 @@ enum Logger {}
 /// BlueprintView signposts
 extension Logger {
```
kyleve, line 8: This can be reverted

BlueprintUICommonControls.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = '5.1'
 
-  s.ios.deployment_target = '12.0'
+  s.ios.deployment_target = '11.0'
```
kyleve, line 14: Agh how did I miss this one

kyleve: Oh wait this is just out of date

kylebshr: LOL I reverted some stuff so I can test in register

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -1,16 +1,44 @@
 import UIKit
 
+/// This property wrapper checks the value of `atMost` cases, and turns it into an
+/// `unconstrained` axis if the value equals `greatestFiniteMagnitude` or `isInfinite`.
+@propertyWrapper public struct UnconstrainedInfiteAxis: Equatable, Hashable {
```
n8chur, line 5: Nit: We may want to either move this to another file if we intend for it to be useful outside of `SizeConstraint`, or nest it within `extension SizeConstraint {` (which would also allow us to remove the `SizeConstraint.`s from `SizeConstraint.Axis` in this property wrapper).

n8chur (approved):

kyleve (approved):

watt:
After some more experimentation I was disappointed to find that this doesn't solve the original problem in all cases. Here's a quick example where an Inset outside a GeometryReader causes the condition to flip between measurement and layout, even with measurement rounding:

```swift
GeometryReader { geometry in
    let wideElement = Spacer(width: CGFloat(400) / 3, height: 10)

    if geometry.constraint.width.maximum >= geometry.measure(element: wideElement).width {
        return wideElement
    } else {
        return Empty()
    }
}
.inset(uniform: 10_000) // a big value to cause precision loss
.inset(uniform: -10_000)
.centered()
```

Since this isn't a comprehensive solution, I'm not sure it's worth the tradeoff of the behavior changes it brings along. At this point, I think it makes more sense to go with a smaller scoped fix (e.g. rounding values locally within the `GeometryReader`, and maybe providing a convenience on `GeometryProxy` to do that).

kyleve:
> Since this isn't a comprehensive solution, I'm not sure it's worth the tradeoff of the behavior changes it brings along.

Does this _only_ fail with weirdly large views (eg 10k wide); or is it a general failure? Eg, if this change in practice makes only those fail, its probably fine, or maybe its a "we land this, plus other fixes" thing? 

I'm not sure how I feel about this as I type it; but wanted to raise.

kyleve:
(Another way to put the above comment: Does this fix make this purely better? If so, IMO its worth landing. We eventually always hit the issues w/ float position no matter the view size; but does this push it out far enough that's its practically no longer a problem?)

kylebshr:
Chatted with Andrew and I do think there's some "weirdness" introduced with this that makes it worth reconsidering - e.g., the equal stack now measuring larger than its constrained size in the tests.

Going to land the size constraint improvements, and make a todo for exploring improvements to geometry reader to get around the rounding issue there (which seems like the only issue this was fixing, since the label issue wasn't actually fixed by this)

PR #256, opened 2021-09-03
Swift version sync

watt:
The podspecs are failing to lint, which is blocking publishing of 0.28.0. The declared Swift version doesn't match the version in `.swift-version`, which was added in #243. Got some advice from @dnkoutso and decided to sync these up on the correct value.

So, this PR includes:
- updating `.swift-version` to 5.3
- syncing the podspecs to that as well, by copying the file's contents
- some misc ruby fixes to the podspecs while I was in there
- rerunning SwiftFormat to pick up 5.3 rules
- gitignoring the .vscode directory, as a treat for me


dnkoutso:
i will be updating cocoapods in 1.12 (master branch currently) to add a flag to allow to disconnect the two. Its a valid option for a dev to use `.swift-version` for means outside of CocoaPods.

kylebshr (approved):

version.rb
```diff
@@ -1,3 +1,5 @@
 # frozen_string_literal: true
 
 BLUEPRINT_VERSION ||= '0.28.0'
+
+SWIFT_VERSION ||= File.read(File.join(__dir__, '.swift-version'))
```
n8chur, line 5: nice

n8chur (approved):

kyleve (approved):

PR #257, opened 2021-09-03
Opt-in pixel rounding behavior

watt:
This is a fix for the issue where labels are sometimes incorrectly truncated or wrapped due to their frames being rounded 1 pixel off of the size they want to be.

It's an opt-in preference you can set on the `ViewDescription` that changes the pixel snapping behavior for a view-backed element.

Scoping the change to only labels avoids the side effects that happen if we change the behavior for all elements, such as gaps or overlaps in edges that should be flush. Gaps or overlaps can still occur with labels, but they shouldn't be noticeable since labels have a transparent background.


BlueprintUI/Sources/Internal/Extensions/CGRect.swift
```diff
@@ -24,6 +24,20 @@ extension CGRect {
         )
     }
 
+    /// Creates a new rectangle by rounding the origin and size of this rectangle.
+    /// - Parameters:
+    ///   - rule: The rounding rule.
+    ///   - scale: The rounding scale.
+    /// - Returns: A rectangle with the rounded values.
+    func roundedForSize(_ rule: FloatingPointRoundingRule, by scale: CGFloat) -> CGRect {
```
kyleve, line 32: Minor since this is an internal method, but the `roundedForFrame` vs `roundedForSize` are kind of confusing to me – eg reading the names I don't exactly know how they differ. But I also don't have a better suggestion!

kyleve: (Maybe we should fold these into one method that takes in the `ViewDescription.RoundingPriority` and switches on that?)

BlueprintUI/Sources/ViewDescription/ViewDescription.swift
```diff
@@ -129,6 +133,14 @@ public struct ViewDescription {
 
 extension ViewDescription {
 
+    /// The available prioritization methods for rounding frames to pixel boundaries.
+    public enum RoundingPriority: Equatable {
```
kyleve, line 137: I would maybe call this `RoundingBehavior` since its behaviour / actions for this element; priority to me implies a relationship/ordering between other elements.

kyleve (approved):

BlueprintUI/Sources/Internal/Extensions/CGRect.swift
kylebshr: Ha that's exactly what I went for in my prototype, with "size" and "border" cases. Agree the current names are a bit confusing

kylebshr: Double edit: looked at my code and I had:

```swift

public enum PixelRoundingPriority {
    case edges
    case size
}
```

I think edges is much clearer than frame.

BlueprintUI/Sources/ViewDescription/ViewDescription.swift
kylebshr: I actually like priority, though I don't feel super strongly. But the cases are what you're prioritizing when rounding - either prioritizing keeping the size or the edges closer to the original.

kyleve: Yes sure – but IMO priority is a "other things outside me are also being considered"; where this purely applies to this thing; eg a behavior.

bencochran: what about
```swift
enum RoundingBehavior {
    case prioritizeSize
    case prioritizeEdges
}
```

kyleve: truly why not both

BlueprintUI/Sources/Internal/Extensions/CGRect.swift
watt: Yeah edges is better, that's what I wrote in the docs too. 👍

BlueprintUI/Sources/ViewDescription/ViewDescription.swift
watt: > what about
> 
> ```swift
> enum RoundingBehavior {
>     case prioritizeSize
>     case prioritizeEdges
> }
> ```

I actually started with this, and the fact that both cases are "prioritize x" is what prompted me to instead describe it more simply in terms of x, the thing being prioritized. This value doesn't really describe the entire rounding behavior, it only describes which of these two things should be prioritized.

watt: updated to
```swift
    /// The available prioritization options for rounding frames to pixel boundaries.
    public enum FrameRoundingBehavior: Equatable {
        /// Prioritize preserving frame edge positions
        case prioritizeEdges
        /// Prioritize preserving frame sizes
        case prioritizeSize
    }
```

kylebshr (approved):

bencochran (approved):

PR #258, opened 2021-09-08
Deprecate .accessibility(...) in favor of .accessibilityElement(...) with required parameters

n8chur:
The existing `.accessibility(...)` function easily mistaken as only overriding the parameters that you specify, where in reality, it overrides all parameters even if the value provided is nil (or omitted).

By requiring the most important bits of accessibility, and renaming the function, the hope is that it we be more clear what this function is actually doing, making it a bit harder to mess up on accident.

In the future we will likely want a way to specify that we only want to override certain parameters while leaving the rest intact, but this should serve as a band-aid in the meantime.


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -96,15 +96,48 @@ public struct AccessibilityElement: Element {
...
 extension Element {
 
-    /// Wraps the element to provide the passed accessibility
-    /// options to the accessibility system.
+
+    /// Wraps the receiver in an accessibility element with the provided values.
+    ///
+    /// - Important: ⚠️ This overrides the accessibility of the contained element and all of its children ⚠️
+    ///
+    /// - SeeAlso: ``accessibilityElement``
```
n8chur, line 104: Installing Xcode 13 to ensure this doc link works as expected, but happy to use a different reference syntax if that's preferred.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -96,15 +96,48 @@ public struct AccessibilityElement: Element {
...
+        )
+    }
+
+    /// Wraps the receiver in an accessibility element with the provided values.
+    ///
+    /// Providing a `nil` value for any of these parameters will result in no resolved value for that accessibility
+    /// parameter—it does not inherit parameters from the wrapped element's accessibility configuration.
+    ///
+    /// - Important: ⚠️ This overrides the accessibility of the contained element and all of its children ⚠️
+    public func accessibilityElement(
```
kyleve, line 134: `setAccessibility`?

kyleve: `settingAccessibility`?

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: +1 for writing new docs in docc format; we'll be supporting it soon!

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: Also kind of a fan of just being very explicit, eg `overrideAccessibility`

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -96,15 +96,48 @@ public struct AccessibilityElement: Element {
 
 extension Element {
```
kyleve, line 97: The file this is in has the `AccessibilityElement` bAse type; we should rename that too and add a typealias that's deprecated.

If we want to, I move deprecated types in Listable into their own file w/ date markers; so they're later easier to know about and remove: https://github.com/kyleve/Listable/blob/main/ListableUI/Sources/Deprecations.swift

BlueprintUICommonControls/Sources/AccessibilityElement.swift
watt: I think `accessibilityElement` is pretty accurate, since this creates a new accessibility element node. "Set" and "override" both have the same problem in that they kind of imply you're putting pushing updated values onto an existing node.

watt (approved):
don't forget the changelog

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: IMO reading just `accessibilityElement` doesn't imply that for me

n8chur: I think `content.accessibility(label: "New Label")` is easy to confuse as just setting accessibility on the content, whereas `content.accessibilityElement(label: "New Label", value: nil, traits: [.staticText])` makes it much more clear that it's wrapping for me.

I'm not sure if it's clear _enough_, and probably not _more_ clear than using the word `override`, but I do think it's just as clear as `setAccessibility` and `settingAccessibility`, personally.

I'll cast a vote for what I have here, but happy to change if anyone feels really strongly (or the majority sways towards another name).

n8chur:
- [ ] Update changelog

BlueprintUICommonControls/Sources/AccessibilityElement.swift
bencochran: Casting a small vote for `overrideAccessibility` or `accessibilityOverride` or similar. The note about `element` making it clear that it's a "wrapping" thing I think is clear for those of us who speak natively in Blueprint element trees, but is less immediately obvious to casual users.

kylebshr: IMO it's the required arguments that make any difference; and I think `content.accessibility(label: "New Label", value: nil, traits: [.staticText])` accomplishes the goal without a name change.

If we really want a name change, I prefer `overrideAccessibility` but I don't think any name really solves the issue, it still feels like you're only overriding the properties you're passing in. 

So.. my vote is none of the above? Just make the arguments required (with a deprecated one without that?)

kylebshr: Hmm after discussing I think I prefer `accessibilityElement` and if I'm the tie breaker that's my vote. But my preference would be keep it `accessibility`, add permutations of all the default args & deprecate those.

n8chur: Will go with `accessibilityElement(...)` for now, then! I'd really like to get this into a release. We can update the name later—if anyone feels strongly please submit another PR on top of this one so we can discuss further.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
n8chur: Since we'll be going with `accessibilityElement(...)` (at least for now) I'm going to stick with the existing structure.

Since I believe it's unlikely that consumers are currently instantiating `AccessibilityElement` directly I'll go ahead and just update those parameters to be required (to match `accessibilityElement(...)`).

- [ ] make parameters required on `AccessibilityElement`

n8chur: Actually I can deprecate the existing initializer and create a new one with parameters that match the order defined in `accessibilityElement`. I'll do that.

kylebshr (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: What about `settingAccessibilityTo`?

kyleve: (If we make all params required; this makes it pretty clear IMO)

n8chur: Gonna go with majority rule (3/5) for this PR and stick with `accessibilityElement(...)`.

watt (approved):

PR #259, opened 2021-09-08
FocusState

watt:
This PR introduces some abstractions for managing focus.

The `Trigger` type from `TextField` is generalized into a top-level type, and wrapped in a `FocusBinding` that includes `onFocus` and `onBlur` callbacks.

You can use this the same way you'd use a trigger today, but it also facilitates a new `@FocusState` property wrapper type that is intended to be the new recommended way to manage focus. `@FocusState` has an API based on [the new `@FocusState` in SwiftUI](https://developer.apple.com/documentation/swiftui/focusstate).

## Example usage

```swift
struct LoginForm: ProxyElement {
    enum Field: Hashable {
        case username
        case password
    }

    var username: String
    var password: String
    var handleLogin: () -> Void

    @FocusState private var focusedField: Field?

        Column { column in
            column.add(
                child: TextField(text: username)
                    .focused(when: $focusedField, equals: .username)
            )
            column.add(
                child: TextField(text: password)
                    .focused(when: $focusedField, equals: .password)
            )
            column.add(
                child: Button(
                    onTap: {
                        if username.isEmpty {
                            focusedField = .username
                        } else if password.isEmpty {
                            focusedField = .password
                        } else {
                            handleLogin()
                        }
                    },
                    wrapping: Label(text: "Sign In")
                )
            )
        }
    }
}
```

This is in draft status while I add docs and clean up, but the general shape is ready for review.

### DIFF
--- a/BlueprintUI/Sources/Focus/FocusBinding.swift
+++ b/BlueprintUI/Sources/Focus/FocusBinding.swift
@@ -1,0 +1,84 @@
+import Foundation
+
+/// A two-way binding between a focusable element's backing view and a `FocusState`-wrapped
+/// property.
+///
+/// Generally you should not need to interact with this type directly. However, you can use focus
+/// bindings to add focus support to custom elements.
+///
+/// ## Adding Focus Bindings
+///
+/// On a `FocusState`, call the `binding(for:)` method to get a binding bound to an optional value,
+/// or `binding()` to get a binding bound to a boolean.
+///
+/// To set up the 2-way binding, there are 2 steps:
+///
+/// 1. Assign actions to the nested `FocusTrigger`, so that your backing view is updated when the
+///    `FocusState`'s value changes.
+///
+/// 2. Call the `onFocus` and `onBlur` callbacks when your backing view gains or loses focus, so
+///    that the value of the bound `FocusState` is updated.
+///
+/// ## Example
+///
+///     final class FocusableView: UIView {
+///         var focusBinding: FocusBinding? {
+///             didSet {
+///                 oldValue?.trigger.focusAction = nil
+///                 oldValue?.trigger.blurAction = nil
+///
+///                 guard let focusBinding = focusBinding else { return }
+///
+///                 focusBinding.trigger.focusAction = { [weak self] in
+///                     self?.becomeFirstResponder()
+///                 }
+///                 focusBinding.trigger.blurAction = { [weak self] in
+///                     self?.resignFirstResponder()
+///                 }
+///
+///                 if isFirstResponder {
+///                     focusBinding.onFocus()
+///                 } else {
+///                     focusBinding.onBlur()
+///                 }
+///             }
+///         }
+///
+///         @discardableResult
+///         override func becomeFirstResponder() -> Bool {
+///             let focused = super.becomeFirstResponder()
+///             if focused {
+///                 focusBinding?.onFocus()
+///             }
+///             return focused
+///         }
+///
+///         @discardableResult
+///         override func resignFirstResponder() -> Bool {
+///             let blurred = super.resignFirstResponder()
+///             if blurred {
+///                 focusBinding?.onBlur()
+///             }
+///             return blurred
+///         }
+///     }
+///
+/// - Tag: FocusBinding
+///
+public struct FocusBinding {
+    /// A trigger, which is responsible for piping focus changes from a `FocusState` into a backing
+    /// view.
+    public let trigger = FocusTrigger()
+    /// A callback to be called by a backing view when it is focused, to pipe changes from a backing
+    /// view to a bound `FocusState`.
+    public var onFocus: () -> Void
+    /// A callback to be called by a backing view when it loses focus, to pipe changes from a
+    /// backing view to a bound `FocusState`.
+    public var onBlur: () -> Void
+
+    /// Creates a new binding
+    public init(onFocus: @escaping () -> Void, onBlur: @escaping () -> Void) {
+        self.onFocus = onFocus
+        self.onBlur = onBlur
+    }
+}
--- a/BlueprintUI/Sources/Focus/FocusState.swift
+++ b/BlueprintUI/Sources/Focus/FocusState.swift
@@ -1,0 +1,249 @@
+import Foundation
+
+/// A property wrapper type that can read and write a value that represents the placement of focus.
+///
+/// Use this property wrapper in conjunction with modifiers on elements that support focus, such as
+/// `TextField.focused(when:equals)` and `TextField.focused(when:)`, to describe when those elements
+/// should have focus. When focus enters the modified element, the wrapped value of this property
+/// updates to match a given value. Similarly, when focus leaves, the wrapped value of this property
+/// resets to `nil` or `false`. Setting this property's value programmatically has the reverse
+/// effect, causing focus to move to the element associated with the updated value.
+///
+/// In the following example of a simple login screen, when the user presses the Sign In button and
+/// one of the fields is still empty, focus moves to that field. Otherwise, the sign-in process
+/// proceeds.
+///
+///     struct LoginForm: ProxyElement {
+///         enum Field: Hashable {
+///             case username
+///             case password
+///         }
+///
+///         var username: String
+///         var password: String
+///         var handleLogin: () -> Void
+///
+///         @FocusState private var focusedField: Field?
+///
+///         var elementRepresentation: Element {
+///             Column { column in
+///                 column.add(
+///                     child: TextField(text: "")
+///                         .focused(when: $focusedField, equals: .username)
+///                 )
+///
+///                 column.add(
+///                     child: TextField(text: "")
+///                         .focused(when: $focusedField, equals: .password)
+///                 )
+///
+///                 column.add(
+///                     child: Button(
+///                         onTap: {
+///                             if username.isEmpty {
+///                                 focusedField = .username
+///                             } else if password.isEmpty {
+///                                 focusedField = .password
+///                             } else {
+///                                 handleLogin()
+///                             }
+///                         },
+///                         wrapping: Label(text: "Sign In")
+///                     )
+///                 )
+///             }
+///         }
+///     }
+///
+/// To allow for cases where focus is completely absent from a view tree, the wrapped value must be
+/// either an optional or a Boolean. Set the focus binding to `false` or `nil` as appropriate to
+/// remove focus from all bound fields. You can also use this to remove focus from a ``TextField``
+/// and thereby dismiss the keyboard.
+///
+/// ### Auto-Focus
+///
+/// To auto-focus a field when it appears, set the value in an `onAppear` hook.
+///
+///     struct Example: ProxyElement {
+///         @FocusState var isFocused: Bool
+///
+///         var elementRepresentation: Element {
+///             TextField(text: "")
+///                 .focused(when: $isFocused)
+///                 .onAppear {
+///                     isFocused = true
+///                 }
+///         }
+///     }
+///
+/// ### Avoid Ambiguous Focus Bindings
+///
+/// A `TextField` can have only one focus binding, stored in its `focusBinding` property. If you apply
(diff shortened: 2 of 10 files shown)


BlueprintUI/Sources/Focus/FocusBinding.swift
```diff
@@ -0,0 +1,20 @@
...
+/// You can use bindings to add focus support to custom elements.
+///
+/// TODO
+///
+public struct FocusBinding {
+    public typealias Callback = () -> Void
+
+    public let trigger = FocusTrigger()
+    public var onFocus: Callback?
+    public var onBlur: Callback?
```
n8chur, line 14: Is `blur` a common way to describe unfocusing from the perspective of interactivity?

I get the analogy, but it took me a minute to put this together. I'm wondering if `onUnfocus` would be more clear, but if others are comfortable with using `blur` in this context then I'm a fan!

BlueprintUI/Sources/Focus/FocusState.swift
```diff
@@ -0,0 +1,238 @@
...
+///                 column.add(
+///                     child: TextField(text: "")
+///                         .focused(when: $focusedField, equals: .name)
+///                 )
+///
+///                 column.add(
+///                     child: TextField(text: "")
+///                         .focused(when: $focusedField, equals: .name) // incorrect re-use of .name
+///                 )
+///             }
```
n8chur, line 90: Might be nice to show an example of how to give something focus `onAppear`

n8chur (commented):
This looks really great! Nice work!

I'm trying to wrap my head around what it might look like if we were to build a reusable component that manages focus (e.g. an Address Entry component), and then embed that in a screen with other focusable items. Is that something that we might support? What would that look like?

BlueprintUI/Sources/Focus/FocusBinding.swift
kyleve: Might be good to align w/ the terminology from UIKit or SwiftUI; which for UIKit is resign and become first responder? (Unsure what language SwiftUI uses)

kylebshr: I think the SwiftUI analogous way would be an onFocusChange with a bool isFocused - though SwiftUI's focus state is really quite different.

I agree onBlur was not initially obvious, but if it's well documented I think it's okay.

watt: I've matched the SwiftUI API where possible, but this part is plumbing that isn't exposed in SwiftUI.

> Is blur a common way to describe unfocusing from the perspective of interactivity?

Yes! Perhaps most notably in [the HTML DOM](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur), the world's most popular declarative UI framework. ;)

BlueprintUI/Sources/Focus/FocusState.swift
watt: > Might be nice to show an example of how to give something focused `onAppear`

Sure thing. I adapted these examples from [Apple's](https://developer.apple.com/documentation/swiftui/focusstate).

BlueprintUI/Sources/Focus/FocusBinding.swift
kyleve: Do we think this is a term iOS engineers will understand? After all, iOS engineers usually don't write HTML.

BlueprintUI/Sources/Focus/FocusState.swift
```diff
@@ -0,0 +1,238 @@
...
+
+        var wrappedValue: Value {
+            get { value }
+            set { value = newValue }
+        }
+
+        private var value: Value {
+            didSet {
+                guard oldValue != value else {
+                    print("skip set")
```
kyleve, line 218: Reminder to remove these before merging

BlueprintUI/Sources/Focus/FocusState.swift
```diff
@@ -0,0 +1,238 @@
...
+
+    /// The current state value, taking into account whatever bindings might be
+    /// in effect due to the current location of focus.
+    ///
+    /// When focus is not in any view that is bound to this state, the wrapped
+    /// value will be `nil` (for optional-typed state) or `false` (for `Bool`-
+    /// typed state).
+    public var wrappedValue: Value {
+        get { value }
+        nonmutating set { value = newValue }
```
kyleve, line 120: I guess we don't have another option; but this property wrapper does break value semantics since it holds a reference type

BlueprintUI/Sources/Focus/FocusTrigger.swift
```diff
@@ -0,0 +1,22 @@
...
+
+    public func focus() {
+        focusAction?()
+    }
+
+    public func blur() {
+        blurAction?()
+    }
+
+    public func releaseActions() {
```
kyleve, line 18: Maybe `clearActions`?

BlueprintUICommonControls/Tests/Sources/TextFieldTests.swift
```diff
@@ -68,5 +68,34 @@ class TextFieldTests: XCTestCase {
 
     }
 
+    func test_focus() {
+        // make a fake window so that focus works
+        let window = UIWindow(frame: UIScreen.main.bounds)
+        let view = BlueprintView()
+        window.addSubview(view)
+        window.makeKeyAndVisible()
```
kyleve, line 76: Do we need to explicitly tear down the window? I forget if showing the window retains it, and eg would then live beyond this test.

kyleve: (Might be worth copying/pasting this over and leveraging it: https://github.com/squareup/market/blob/main/ios/MarketUITesting/Sources/XCTestCase+AppHost.swift#L22)

watt:
> I'm trying to wrap my head around what it might look like if we were to build a reusable component that manages focus (e.g. an Address Entry component), and then embed that in a screen with other focusable items. Is that something that we might support? What would that look like?

Pretty straightforward, I think. While a `TextField` can only have one binding, any number of `@FocusState` properties can coexist.

An entirely self-contained form could look something like this:

```swift
struct AddressForm: ProxyElement {
    enum Field: Hashable { case name, street, city }

    var name: String
    var street: String
    var city: String

    @FocusState private var focusedField: Field?

        Column { column in
            column.add(
                child: TextField(text: name)
                    .focused(when: $focusedField, equals: .name)
            )

            column.add(
                child: TextField(text: street)
                    .focused(when: $focusedField, equals: .street)
            )

            column.add(
                child: TextField(text: city)
                    .focused(when: $focusedField, equals: .city)
            )
        }
    }
}
```

And if you wanted to hoist focus management up, you can inject it as well:

```swift
struct AddressForm: ProxyElement {
    init(name: String, street: String, city: String, focusedField: FocusState<Field?>) {
        self.name = name
        self.street = street
        self.city = city
        self._focusedField = focusedField
    }
    // other stuff...
}

struct ParentForm: ProxyElement {
    var nickname: String

    @FocusState var focusedAddressField: AddressForm.Field?
    @FocusState var isNicknameFocused: Bool

        Column { column in
            column.add(
                child: TextField(text: nickname)
                    .focused(when: $isNicknameFocused)
            )

            column.add(
                child: AddressForm(
                    name: "",
                    street: "",
                    city: "",
                    focusedField: $focusedAddressField
                )
            )
        }
    }
}
```

BlueprintUI/Sources/Focus/FocusBinding.swift
watt: I might be setting myself up for disappointment, but I'm pretty confident that folks will be able to figure this out!

BlueprintUI/Sources/Focus/FocusState.swift
watt: added!

BlueprintUICommonControls/Tests/Sources/TextFieldTests.swift
watt: Nope! The window isn't retained. I added a helper for scoping though.

BlueprintUI/Sources/Focus/FocusTrigger.swift
```diff
@@ -0,0 +1,31 @@
...
+///
+/// ## See Also
+/// [FocusBinding](x-source-tag://FocusBinding)
+///
+public final class FocusTrigger {
+    /// The action to be invoked on focus, which will be set by a backing view.
+    public var focusAction: (() -> Void)?
+    /// The action to be invoked on blur, which will be set by a backing view.
+    public var blurAction: (() -> Void)?
+
```
n8chur, line 21: Missing initializer?

BlueprintUICommonControls/Sources/TextField.swift
```diff
@@ -148,4 +137,158 @@ fileprivate final class CallbackTextField: UITextField, UITextFieldDelegate {
...
+        }
+        return blurred
+    }
+
+}
+
+extension TextField {
+    /// Modifies this text field by binding its focus state to the given state value.
+    ///
+    /// Use this modifier to cause the text field to receive focus whenever the the `state` equals
```
n8chur, line 163:
```suggestion
    /// Use this modifier to cause the text field to receive focus whenever the `state` equals
```

n8chur (approved):
Awesome work! 🎉

BlueprintUI/Sources/Focus/FocusState.swift
```diff
@@ -0,0 +1,249 @@
+import Foundation
+
+/// A property wrapper type that can read and write a value that represents the placement of focus.
+///
+/// Use this property wrapper in conjunction with modifiers on elements that support focus, such as
+/// `TextField.focused(when:equals)` and `TextField.focused(when:)`, to describe when those elements
```
kylebshr, line 6: Nit
```suggestion
/// `TextField.focused(when:equals:)` and `TextField.focused(when:)`, to describe when those elements
```

BlueprintUI/Sources/Focus/FocusState.swift
kylebshr: Fun trick, I'm guessing some of Apple's own wrappers do this. Seems okay for stuff like this? Is a property wrapper _really_ a struct? :P

kylebshr (approved):
Really slick, nice work

BlueprintUI/Sources/Focus/FocusState.swift
kyleve: SwiftUI does some fancy object introspection so they can maintain value semantics; but we don't really have the ability to do it here that easily.

watt: > SwiftUI does some fancy object introspection so they can maintain value semantics; but we don't really have the ability to do it here that easily.

Aren't the semantics the same though? Both have a `nonmutating set`. I don't think you can say the semantics are different because of the implementation.

kyleve: Thinking about this case:

```
var myElement = MyElement() // contains an `@FocusState` internally.
var anotherElement = myElement // makes a value copy

blueprintView1.element = myElement
blueprintView2.element = anotherElement
```

These two will (AFAIK) now share the same underlying `Storage`

watt: Yep, I see what you mean. I'm not sure you can really do that in SwiftUI either, though.

Anchors and triggers work the same way too. I could make `FocusState` a class, but since it's typically used as a property wrapper, I don't know if that really helps.

BlueprintUI/Sources/Focus/FocusTrigger.swift
watt: fixed!

BlueprintUI/Sources/Focus/FocusState.swift
kyleve: I wonder if we could/should set up some sort of validation for this – eg "you only get to be in one blueprint view at once" or something. Since it's insidious if you do mess it up; but maybe no one actually ever will...

PR #260, opened 2021-09-10
Defer lifecycle callbacks until view updates complete

watt:
This changes the time when lifecycle callbacks encountered during a view update are run: instead of running them in-line with the view update traversal, they are aggregated and run after the view update completes. It does not change the relative order of the callbacks themselves.

This ensures that all view descriptions are applied before running `onAppear` callbacks. Notably, this means that any focus triggers on children will be set up, so callbacks set on parents can use them.

Some additional docs on the appearance modifiers explain what is guaranteed (relative callback order) and what is not (superview removal).


kylebshr (approved):

n8chur (approved):
Thanks for the quick turn-around on this!

PR #262, opened 2021-09-13
Ensure infinite constraints are `.unconstrained`

kylebshr:
This init for `Axis` attempts to handle this, but there are many ways to create a constraint axis with a value of `.atMost([.greatestFiniteMagnitude|.infinity])` - e.g., using the enum case directly rather than the initializer. This PR ensures any constraint axes set on a `SizeConstraint` get set to unconstrained if it's one of those values.

### DIFF
--- a/BlueprintUI/Sources/Measuring/SizeConstraint.swift
+++ b/BlueprintUI/Sources/Measuring/SizeConstraint.swift
@@ -7,10 +7,10 @@ import UIKit
 public struct SizeConstraint: Hashable {
 
     /// The width constraint.
-    public var width: Axis
+    @UnconstrainedInfiteAxis public var width: Axis
 
     /// The height constraint.
-    public var height: Axis
+    @UnconstrainedInfiteAxis public var height: Axis
 
     public init(width: Axis, height: Axis) {
         self.width = width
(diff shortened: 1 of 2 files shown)


BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -176,3 +176,33 @@ extension SizeConstraint {
 
     }
 }
+
+extension SizeConstraint {
+    /// This property wrapper checks the value of `atMost` cases, and turns it into an
+    /// `unconstrained` axis if the value equals `greatestFiniteMagnitude` or `isInfinite`.
+    @propertyWrapper public struct UnconstrainedInfiteAxis: Equatable, Hashable {
```
watt, line 183:
```suggestion
    @propertyWrapper public struct UnconstrainedInfiniteAxis: Equatable, Hashable {
```

BlueprintUI/Sources/Measuring/SizeConstraint.swift
```diff
@@ -176,3 +176,33 @@ extension SizeConstraint {
...
+        public var wrappedValue: Axis {
+            get { correctedAxis }
+            set { correctedAxis = Self.correctedAxis(for: newValue) }
+        }
+
+        public init(wrappedValue value: Axis) {
+            correctedAxis = Self.correctedAxis(for: value)
+        }
+
+        private static func correctedAxis(for axis: Axis) -> Axis {
```
watt, line 195: Mind adding a unit test for this behavior?

watt (approved):

BlueprintUI/Sources/Measuring/SizeConstraint.swift
kylebshr: Done!

PR #263, opened 2021-09-15
Return UIView.noIntrinsicMetric for intrinsicContentSize when there is no element

n8chur:
When building out support for custom input views I noticed a log message when UIKit attempted to interact with a `BlueprintView` that didn't have an element set yet:

> 2021-09-14 15:01:12.232905-0700 Demo[23792:556733] API error: <MarketUI.MarketInputViewElementContainer: 0x7ff394d34460; baseClass = UIInputView; frame = (0 0; 390 71); layer = <CALayer: 0x600003d390a0>> returned 0 width, assuming UIViewNoIntrinsicMetric

According to [the docs for `UIView.intrinsicContentSize`](https://developer.apple.com/documentation/uikit/uiview/1622600-intrinsiccontentsize):

> If a custom view has no intrinsic size for a given dimension, it can use noIntrinsicMetric for that dimension.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -188,7 +188,12 @@ public final class BlueprintView: UIView {
     /// UILabel’s `intrinsicContentSize` behavior)
     public override var intrinsicContentSize: CGSize {
 
-        guard let element = element else { return .zero }
+        guard let element = element else {
+            return CGSize(
+                width: UIView.noIntrinsicMetric,
+                height: UIView.noIntrinsicMetric
+            )
+        }
 
         let constraint: SizeConstraint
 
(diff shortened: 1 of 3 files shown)


bencochran (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -188,7 +188,12 @@ public final class BlueprintView: UIView {
     /// UILabel’s `intrinsicContentSize` behavior)
     public override var intrinsicContentSize: CGSize {
```
kyleve, line 189: Shouldn't we also be assigning the width and height that comes out of `return element.content.measure(`, eg if it returns `0` for a dimension? (Eg, imagine the element is an `Empty()`)

tyten: I think in those cases `0` would be acceptable? The error we're trying to prevent only presents itself if there's not an element yet, right?

watt (approved):
A quick google indicates this error is correlated with input views, which I just thought was interesting.

tyten (approved):

n8chur:
> A quick google indicates this error is correlated with input views, which I just thought was interesting.

Yup! I got this error when the BlueprintView was a subview of an `inputView`

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Fair enough!

PR #264, opened 2021-09-18
Add keys to Overlay

watt:
Adding disambiguation keys to Overlay, like we currently have in other stack-likes.

### DIFF
--- a/BlueprintUI/Sources/Layout/Overlay.swift
+++ b/BlueprintUI/Sources/Layout/Overlay.swift
@@ -10,28 +10,39 @@ import UIKit
 public struct Overlay: Element {
 
     /// All elements displayed in the overlay.
-    public var elements: [Element]
+    public var children: [Child]
 
     /// Creates a new overlay with the provided elements.
     public init(
         elements: [Element] = [],
         configure: (inout Overlay) -> Void = { _ in }
     ) {
-        self.elements = elements
+        children = elements.map { Child(element: $0) }
         configure(&self)
     }
 
     /// Adds the provided element to the overlay.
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/Overlay.swift
```diff
@@ -10,28 +10,39 @@ import UIKit
...
         elements: [Element] = [],
         configure: (inout Overlay) -> Void = { _ in }
     ) {
-        self.elements = elements
+        children = elements.map { Child(element: $0) }
         configure(&self)
     }
 
     /// Adds the provided element to the overlay.
+    @available(*, deprecated, renamed: "add(child:)")
```
kyleve, line 25: IMO keeping this and not deprecating it is probably fine; since not everyone needs/wants keys?

kyleve: (Eg, so the shorter method is probably fine to avoid a ton of deprecation warnings)

kyleve (approved):

BlueprintUI/Sources/Layout/Overlay.swift
kylebshr: meh, stacks use `add(child:)` and I kinda like the consistency? no strong opinion though

kylebshr (approved):

n8chur (approved):

BlueprintUI/Sources/Layout/Overlay.swift
watt: I thought it was confusing to have both `add(_:)` and `add(child:)`, and the latter is consistent with other stacks.

PR #265, opened 2021-09-21
Add ShrinkBehavior for stack children

kylebshr:
`ShrinkBehavior` allows stack children to specify whether they should fill the available space when shrinking (the behavior today), or be re-measured in the new smaller space and "fitted" to that new size.

This allows you to improve behavior when an elements "intrinsic" size can change based on the available space - for example, a label that wraps.

Concretely, in `MarketRow` we use a `Row` to build the content:

```
|-Row-------------------------------|
| Leading | Text Content | Trailing |
|-----------------------------------|
```

In the case where Text Content consists of another `Row` with a multiline-label flexible followed by a fixed label, there can be undesired spacing introduced. This is because the label is measured within the constraint of the `Row`, then shrunk to fit, causing words to wrap that may not have been wrapped, which reduces the width of the text. However, the `Row` doesn't know this and can't currently account for it. 

Desired behavior:

```
|-Row-------------------------------------------|
|         | Multiline                |          |
| Leading | Text       Label         | Trailing |
|         | Content                  |          |
|-----------------------------------------------|
```

Actual behavior:

```
|-Row-------------------------------------------|
|         | Multiline                |          |
| Leading | Text            Label    | Trailing |
|         | Content                  |          |
|-----------------------------------------------|
```

In this case, specifying `.fit` for the shrink behavior tells the row to re-measure the label when it shrinks it, resulting in the desired behavior.

### Open Questions

Should this be automatic? We could measure shrunk children and use the new measurement if it's smaller than the original. I _think_ it's usually desired, but doing it automatically means more measurements vs. opting in. On the other hand, the bug the current behavior introduces is pretty subtle and it's not obvious how to fix it, so it might be worth it - possibly worth profiling 

### Todo

- [ ] Update changelog
- [ ] Refine implementation, probably

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -75,6 +75,18 @@ extension StackElement {
     ///     }
     ///     ```
     ///
+    ///   - shrinkBehavior:
+    ///
+    ///     If the layout overflows and the child shrinks, this allows you to specify whether the child
+    ///     should fill the space its given, or be re-measured in the shrunk space and fitted based on
+    ///     the new measurement.
+    ///
+    ///     For example, if a label in a row wraps a word to a new line when forced to shrink, `fill` will make
+    ///     the label larger than necessary, with whitespace on the trailing edge. The `fit` option will shrink
+    ///     the label further so it encloses the new wrapped layout.
+    ///
+    ///     Note that "surrendered" space is not distributed among other elements.
+    ///
     ///   - alignmentGuide:
     ///
     ///     A closure that can be used to provide a custom alignment guide for this child.
@@ -98,6 +110,7 @@ extension StackElement {
     public mutating func add(
         growPriority: CGFloat = 1.0,
         shrinkPriority: CGFloat = 1.0,
+        shrinkBehavior: StackLayout.ShrinkBehavior = .fill,
         alignmentGuide: ((ElementDimensions) -> CGFloat)? = nil,
         key: AnyHashable? = nil,
         child: Element
@@ -107,7 +120,8 @@ extension StackElement {
             traits: StackLayout.Traits(
                 growPriority: growPriority,
                 shrinkPriority: shrinkPriority,
-                alignmentGuide: alignmentGuide.map(StackLayout.AlignmentGuide.init)
+                alignmentGuide: alignmentGuide.map(StackLayout.AlignmentGuide.init),
+                shrinkBehavior: shrinkBehavior
             ),
             key: key
         ))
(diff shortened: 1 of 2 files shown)


kyleve:
> Should this be automatic? We could measure shrunk children and use the new measurement if it's smaller than the original. I think it's usually desired, but doing it automatically means more measurements vs. opting in { ... }

I would make it opt in. There's already a pretty expensive cost to nesting multiple rows/columns due to the re-measurements in many cases; and this would have a penalty for all existing rows and columns.

(Also in general; just keeping the layout behavior "as is" is probably better so we don't accidentally break things?)

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -187,6 +203,17 @@ public struct StackLayout: Layout {
         public var computeValue: (ElementDimensions) -> CGFloat
     }
 
+    /// Determines how a stack child will behave when it shrinks.
+    public enum ShrinkBehavior {
```
kyleve, line 207: I wonder if these might be clearer as `scale` and `remeasure`?

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -537,9 +591,18 @@ extension StackLayout {
 
         var axisOrigin: CGFloat = 0.0
 
-        let axisSegments = zip(basisSizes, shrinkPriorities).map { basis, shrinkPriority -> Segment in
+        let axisSegments = basisSizes.enumerated().map { index, basis -> Segment in
```
kyleve, line 594: Fwiw, might be worth adding and using an `indexedMap` here given its a pretty hot codepath. When I was doing performance testing; I saw (even in release mode) about a 25% perf bump from using this:

```
extension Array {

    /// A `map` implementation that also passes the `index` of each element in the original array.
    ///
    /// This method is more performant than calling `array.enumerated().map(...)` by up
    /// to 25% for large collections, so prefer it when needing an indexed `map` in areas where performance is critical.
    @inlinable func indexedMap<Mapped>(_ map : (Int, Element) -> Mapped) -> [Mapped] {

        let count = self.count

        var mapped = [Mapped]()
        mapped.reserveCapacity(count)

        for index in 0..<count {
            mapped.append(map(index, self[index]))
        }

        return mapped
    }
```

I believe because there's no enumeration structure being created.

kyleve (approved):

kyleve:
Based on doing some preliminary testing in Market, I think this PR https://github.com/square/Blueprint/pull/348 ends up resolving this issue somewhat, because the measurement now takes place in a much more realistic size.

kyleve:
Closing, given that https://github.com/square/Blueprint/pull/348 addresses this for the most part

PR #267, opened 2021-09-23
Bump CI Xcode version to 12.5.1 and macOS to 11

n8chur:


.github/workflows/tests.yaml
```diff
@@ -28,8 +30,8 @@ jobs:
     steps:
```
n8chur, line 30: I can't comment on the line, but it's above on line 27/28:

Should we bump this to 14.5?

.github/workflows/tests.yaml
```diff
@@ -28,8 +30,8 @@ jobs:
...
     - uses: actions/checkout@v1
 
-    - name: Switch Xcode 12
-      run: sudo xcode-select -switch /Applications/Xcode_12.2.app
+    - name: Switch to Xcode 12.5.1
+      run: sudo xcode-select -s /Applications/Xcode_12.5.1.app
 
     - name: Install iOS ${{ matrix.sdk }}
       run: xcversion simulators --install="iOS ${{ matrix.sdk }}"
```
n8chur, line 37: I can't comment on the line, but it's above on line 27/28:

Should we bump the iOS 14 version to 14.5?

kylebshr: I think SPOS is doing that so probably should match it

kylebshr: Makes everything easier (e.g., will likely have matching sims installed locally)

kylebshr (approved):

n8chur:
Re-requesting review since I bumped the `swift-version`—just double checking that that seems safe @kylebshr!

n8chur:
I didn't bump the `Package.swift` tooling version as part of these changes but we probably want to do that soon. @watt pointed out that the 5.4 tooling change how resources are resolved so it shows a bunch of warnings we'll have to fix when we bump that.

.github/workflows/tests.yaml
n8chur: - [ ] Use 14.5 simulator

kyleve (approved):

PR #268, opened 2021-09-24
UserInteractionEnabled and Hidden layout attributes

kylebshr:
Make user interaction and hidden layout attributes, move `UserInteractionEnabled` into `BlueprintUI`, and add `Hidden`.


BlueprintUI/Sources/Layout/Hidden.swift
```diff
@@ -0,0 +1,45 @@
+import CoreGraphics
+
+/// `Hidden` conditionally hides its wrapped element.
+///
+/// - Note: When an element is hidden, any elements within the wrapped element will be hidden.
+public struct Hidden: Element {
```
kyleve, line 6: Do we want to go with `IsHidden`, maybe?

BlueprintUI/Sources/Layout/Hidden.swift
```diff
@@ -0,0 +1,45 @@
...
+            return attributes
+        }
+    }
+}
+
+extension Element {
+    /// Conditionally hide the wrapped element.
+    ///
+    /// - Note: When an element is hidden, any elements within the wrapped element will be hidden.
+    public func hidden(_ hidden: Bool = true) -> Hidden {
```
kyleve, line 42: Similar Q: re `isHidden`?

BlueprintUI/Sources/Layout/Hidden.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    private struct Layout: SingleChildLayout {
+        var isHidden: Bool
+
+        func measure(in constraint: SizeConstraint, child: Measurable) -> CGSize {
+            child.measure(in: constraint)
+        }
+
+        func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
+            var attributes = LayoutAttributes(size: size)
```
kyleve, line 31: Outloud thought: It'd be nice if `LayoutAttributes` had a configure block to clean up these callsites:

```
LayoutAttributes(size: size) {
   $0.isHidden = isHidden
}
```

kyleve (approved):
Nice

BlueprintUI/Sources/Layout/Hidden.swift
n8chur: We already have `UserInteractionEnabled` (without `is`).

I'm fine with this (or updating both + any others that are similar if they exist)—I'd mainly just want to be consistent.

BlueprintUI/Tests/LayoutAttributesTests.swift
```diff
@@ -42,6 +42,20 @@ final class LayoutAttributesTests: XCTestCase {
             XCTAssertNotEqual(attributes, other)
         }
 
+        do {
+            /// Enabled
```
n8chur, line 46:
```suggestion
            /// User Interaction Enabled
```

n8chur (approved):

BlueprintUI/Sources/Layout/Hidden.swift
kylebshr: I prefer `Hidden`, I think it's a better name since it's an element and not a boolean property.

BlueprintUI/Sources/Layout/Hidden.swift
kylebshr: I also prefer hidden here; and it matches SwiftUI

n8chur: Agree with trying to match SwiftUI here (at least with the default `.hidden()` usage), but don't feel strongly.

kyleve: Hmm ok; SG. This will diverge from UIKit, but I guess that's fine!

n8chur: Does UIKit have a functional chaining syntax to set hidden on a view like this?

This isn't used for if checks where "is" reads better (although the wrapping view's properly is `isHidden`). IMO chaining `hidden()` vs `isHidden()` reads better.

PR #269, opened 2021-09-27
Remove deprecated AccessibilityElement initializer

n8chur:
The deprecated initializer caused existing call sites for `AccessibilityElement` to fail to compile due to ambiguity on which initializer it should resolve to.

I was already on the fence on whether or not we should just make this a breaking change on the first pass, so I'm leaning towards this approach of just removing the deprecated initializer, especially since those who already integrated the previous version of Blueprint were forced to make changes in a majority of these cases (e.g. in Register).

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -49,31 +49,6 @@ public struct AccessibilityElement: Element {
         wrappedElement = element
     }
 
-    @available(
-        *,
-        deprecated,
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -121,57 +96,57 @@ public struct AccessibilityElement: Element {
 
 extension Element {
```
n8chur, line 97: The changes here look complicated, but it's really just switching the order in which these functions are defined—the deprecated function is now below the new one.

kylebshr (approved):

kyleve (approved):

PR #270, opened 2021-09-29
Introduced helpers for adding spacers to stacks

kyleve:
It's a pretty common request of "how do add a flexible spacer to do X", so this seems useful.

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -180,6 +180,37 @@ extension StackElement {
             child: child
         )
     }
+
+    /// Adds a `Spacer` element to the stack, which will grow (and shrink) depending on the amount
+    /// of left over available space when measuring the elements.
+    public mutating func addFlexibleSpacer() {
+        precondition(
+            layout.underflow.growsInUnderflow,
+            """
+            Flexible spacers have no effect if the Stack's underflow does not stretch its contents. \
+            Underflow must be one of: `.growProportionally`, `.growUniformly`.
+            """
+        )
+
+        add(growPriority: 1, shrinkPriority: 1, child: Spacer(1.0))
+    }
+
+    /// Adds a fixed `Spacer` element to the stack, with the provided size – width for rows, and height for columns.
+    ///
+    /// ### Note
+    /// Do not use this method to add spacing between each element in your stack.
+    ///
+    /// Instead, utilize the `minimumHorizontalSpacing` (`Row`)
+    /// or `minimumVerticalSpacing` (`Column`) to control uniform spacing
+    /// between elements in your stack.
+    public mutating func addFixedSpacer(_ size: CGFloat) {
+        switch layout.axis {
+        case .horizontal:
+            addFixed(child: Spacer(width: size, height: 0))
+        case .vertical:
+            addFixed(child: Spacer(width: 0, height: size))
+        }
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -180,6 +180,37 @@ extension StackElement {
...
         )
     }
+
+    /// Adds a `Spacer` element to the stack, which will grow (and shrink) depending on the amount
+    /// of left over available space when measuring the elements.
+    public mutating func addFlexibleSpacer() {
+        precondition(
+            layout.underflow.growsInUnderflow,
+            """
+            Flexible have no effect if the Stack's underflow does not stretch contents. \
```
kyleve, line 190: Flexible _spacers_ have...

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -180,6 +180,37 @@ extension StackElement {
             child: child
         )
     }
+
+    /// Adds a `Spacer` element to the stack, which will grow (and shrink) depending on the amount
+    /// of left over available space when measuring the elements.
+    public mutating func addFlexibleSpacer() {
```
kyleve, line 186: Of note; this won't work with result builders. I think we may need a special element type to deal with that. Or we just make `Spacer()` have a default size of `1` so that growing and shrinking works right...

n8chur, line 186: Should we allow passing in a `minimumSize: CGFloat = 1`? Also, should the default be `0`?

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -180,6 +180,37 @@ extension StackElement {
...
+        precondition(
```
n8chur, line 187: This doesn't seem worth crashing for IMO. I'd expect a flexible spacer to just not space when in an underflow distribution mode that doesn't cause this to do more than the minimum size (currently `1`).

Curious what others think.

BlueprintUI/Sources/Layout/Stack.swift
kyleve: It needs to be `1` because the value is multiplied by the row in order to scale it – if we pass 0, the math becomes `something *0`, which becomes 0. Passing a minimum size seems good, too

BlueprintUI/Sources/Layout/Stack.swift
kyleve: I guess I did the crash over a silent / confusing failure that people usually run into, since a lot of the time people don't realize that the flexible sizing has no effect outside of these underflow options.

BlueprintUI/Sources/Layout/Stack.swift
bencochran: IMO, 1 technically works but it's usually not the right call. That is, if you have flexible space you also typically want it to have a decent looking minimum (of, say, 8pts or whatever's appropriate for your design).

n8chur: > in order to scale it

Oh, right, I forgot `Spacer` would be scaled just like any other Element in a stack... I guess I intuitively expected `Spacer` to behave more like a one-off `minimumSpacing` that can optionally flex to fill space, but not scale proportionally with other `Element`s in a stack. Hm...

kyleve:
This isn't useful for result builders; so we need an easier way to do that there. Closin

PR #272, opened 2021-10-11
Add result builders for Overlay and SegmentedControl

kylebshr:
These were the only places I found that made sense to add result builders. 

Now you can build an overlay:

```swift
let overlay = Overlay {
    TestElement()
    TestElement2().overlayChild(key: "foo")
    if state.showBadge {
        BadgeElement()
    }
}
```

or a segmented control:

```swift
let control = SegmentedControl {
    SegmentedControl.Item(title: "First", onSelect: {})
    SegmentedControl.Item(title: "Second", onSelect: {})
    if state.isThirdOptionEnabled {
        SegmentedControl.Item(title: "Third", onSelect: {})
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added a `Hidden` element and `.hidden()` modifier for hiding elements.
+- `Overlay` now supports result builders.
+- `SegmentedControl` now supports result builders.
 
 ### Removed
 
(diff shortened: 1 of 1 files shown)


ldstreet:
> > Would this be a bad idea?
> 
> 
> 
> Oh that's nice - should we ditch the `overlayChild` func if we add that?

I wouldn't think so? It's more discoverable to start with your element then layer on a key I think. But maybe it could be renamed/replaced with `keyed(String) -> Keyed` though? Or does that function already exist?

kylebshr:
`keyed` already exists (and returns a Keyed) so I think it removes the need
for overlayChild until/if it supports more than keys

kyleve:
IMO it might be nice to to keep overlayChild since we may want more than keys at some point?

PR #277, opened 2021-11-05
Change how we cache sizing across sizeThatFits and autolayout

kyleve:
@n8chur recently added support for caching intrinsic content size, which made me wonder: Why don't we do this everywhere? This will provide benefits for any existing impl which calls `sizeThatFits` or relies on the other autolayout methods, since we now only do an expensive re-measure when content has actually changed.

### Changes
- `sizeThatFits` now also caches results.
- `intrinsicContentSize` is now backed by `sizeThatFits`.
- Added `systemLayoutSizeFitting...` overrides to call through to `sizeThatFits`.
- We invalidate sizing when environment, safe area, etc, change.

### One more thing!
The old impl of `intrinsicContentSize` would use an unconstrained width if we had a new element, but had not yet laid out. This seems really wrong to me, since it was very inconsistent and hard to predict, to the point of non-determinism: Had you laid out yet? Great! Your constraint is the view width! Had you not? Uhh, you have no width constraint. This is a behaviour change, and there was value to the old behaviour, so it might make sense to add an optional/mode for this. Thoughts?

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -36,7 +36,7 @@ public final class BlueprintView: UIView {
 
     private let rootController: NativeViewController
 
-    private var cachedIntrinsicContentSize: CGSize? = nil
+    private var sizesThatFit: [SizeConstraint: CGSize] = [:]
 
     /// A base environment used when laying out and rendering the element tree.
     ///
@@ -49,7 +49,7 @@ public final class BlueprintView: UIView {
     public var environment: Environment {
         didSet {
             setNeedsViewHierarchyUpdate()
-            invalidateIntrinsicContentSize()
+            invalidateCachedSizes()
         }
     }
 
@@ -67,8 +67,12 @@ public final class BlueprintView: UIView {
     ///
     public var automaticallyInheritsEnvironmentFromContainingBlueprintViews: Bool = true {
         didSet {
+            guard oldValue != automaticallyInheritsEnvironmentFromContainingBlueprintViews else {
+                return
+            }
+
             setNeedsViewHierarchyUpdate()
-            invalidateIntrinsicContentSize()
+            invalidateCachedSizes()
         }
     }
 
@@ -83,7 +87,7 @@ public final class BlueprintView: UIView {
             Logger.logElementAssigned(view: self)
 
             setNeedsViewHierarchyUpdate()
-            invalidateIntrinsicContentSize()
+            invalidateCachedSizes()
         }
     }
 
@@ -140,7 +144,7 @@ public final class BlueprintView: UIView {
 
     @available(*, unavailable)
     public required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+        fatalError("init(coder:) is not available for BlueprintView.")
     }
 
     ///
@@ -161,7 +165,7 @@ public final class BlueprintView: UIView {
     /// blueprintView.sizeThatFits(.zero)
     /// ```
     ///
-    public override func sizeThatFits(_ size: CGSize) -> CGSize {
+    public override func sizeThatFits(_ fittingSize: CGSize) -> CGSize {
         guard let element = element else {
             return .zero
         }
@@ -179,45 +183,109 @@ public final class BlueprintView: UIView {
             )
         }
 
-        return element.content.measure(
-            in: measurementConstraint(with: size),
+        let constraint = measurementConstraint(with: fittingSize)
+
+        if let cachedSize = sizesThatFit[constraint] {
+            return cachedSize
+        }
+
+        let measurement = element.content.measure(
+            in: constraint,
             environment: makeEnvironment(),
             cache: CacheFactory.makeCache(name: "sizeThatFits:\(type(of: element))")
         )
+
+        sizesThatFit[constraint] = measurement
+
+        return measurement
+    }
+
+    ///
+    /// Measures the size needed to display the view within then given constraining size,
+    /// by measuring the current `element` of the `BlueprintView`.
+    ///
+    /// If you would like to not constrain the measurement in a given axis,
+    /// pass `0.0` or `.greatestFiniteMagnitude` for that axis.
+    ///
+    public override func systemLayoutSizeFitting(
+        _ targetSize: CGSize
+    ) -> CGSize {
+        /// For us, this is the same as `sizeThatFits`, since blueprint does not
+        /// translate the concept of constraints.
+        sizeThatFits(targetSize)
+    }
+
+    ///
+    /// Measures the size needed to display the view within then given constraining size,
+    /// by measuring the current `element` of the `BlueprintView`.
+    ///
+    /// If you would like to not constrain the measurement in a given axis,
+    /// pass `0.0` or `.greatestFiniteMagnitude` for that axis.
+    ///
+    public override func systemLayoutSizeFitting(
+        _ targetSize: CGSize,
+        withHorizontalFittingPriority horizontalFittingPriority: UILayoutPriority,
+        verticalFittingPriority: UILayoutPriority
+    ) -> CGSize {
+        /// For us, this is the same as `sizeThatFits`, since blueprint does not
+        /// translate the concept of horizontal or vertical constraints.
+        sizeThatFits(targetSize)
     }
 
     public override var intrinsicContentSize: CGSize {
-        if let intrinsicContentSize = cachedIntrinsicContentSize {
-            return intrinsicContentSize
+        guard element != nil else {
+            return CGSize(
+                width: UIView.noIntrinsicMetric,
+                height: UIView.noIntrinsicMetric
+            )
         }
 
-        let intrinsicContentSize = resolveIntrinsicContentSize()
-        cachedIntrinsicContentSize = intrinsicContentSize
-        return intrinsicContentSize
-    }
+        let constraint = { () -> SizeConstraint in
+            if bounds.width == 0 {
+                return .unconstrained
+            } else {
+                return .init(width: bounds.width)
+            }
+        }()
 
-    public override func invalidateIntrinsicContentSize() {
-        cachedIntrinsicContentSize = nil
-        super.invalidateIntrinsicContentSize()
+        return sizeThatFits(constraint.maximum)
     }
 
     public override var semanticContentAttribute: UISemanticContentAttribute {
         didSet {
             if semanticContentAttribute != oldValue {
                 setNeedsViewHierarchyUpdate()
+                invalidateCachedSizes()
             }
         }
     }
 
     public override func safeAreaInsetsDidChange() {
         super.safeAreaInsetsDidChange()
+
         setNeedsViewHierarchyUpdate()
+        invalidateCachedSizes()
+    }
+
+    public override var frame: CGRect {
+        didSet {
+            guard oldValue != frame else { return }
+
+            invalidateIntrinsicContentSize()
+        }
+    }
+
+    public override var bounds: CGRect {
+        didSet {
+            guard oldValue != bounds else { return }
+
+            invalidateIntrinsicContentSize()
+        }
     }
 
     public override func layoutSubviews() {
         super.layoutSubviews()
-        invalidateIntrinsicContentSize()
-        performUpdate()
+        updateViewHierarchyIfNeeded()
     }
 
     public override func didMoveToWindow() {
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -49,7 +49,7 @@ public final class BlueprintView: UIView {
     public var environment: Environment {
         didSet {
             setNeedsViewHierarchyUpdate()
-            invalidateIntrinsicContentSize()
```
kyleve, line 52: Oops should've kept this one

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -226,18 +294,20 @@ public final class BlueprintView: UIView {
         setNeedsViewHierarchyUpdate()
     }
 
-    private func performUpdate() {
```
kyleve, line 229: This was called in exactly one place and only did this, so I removed it.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,45 +183,109 @@ public final class BlueprintView: UIView {
...
-    public override func invalidateIntrinsicContentSize() {
-        cachedIntrinsicContentSize = nil
-        super.invalidateIntrinsicContentSize()
+        return sizeThatFits(constraint.maximum)
     }
 
     public override var semanticContentAttribute: UISemanticContentAttribute {
         didSet {
             if semanticContentAttribute != oldValue {
                 setNeedsViewHierarchyUpdate()
```
kylebshr, line 257: Maybe calls to `setNeedsViewHierarchyUpdate` should just call `invalidateCachedSizes`?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,45 +183,109 @@ public final class BlueprintView: UIView {
...
+
         setNeedsViewHierarchyUpdate()
+        invalidateCachedSizes()
+    }
+
+    public override var frame: CGRect {
+        didSet {
+            guard oldValue != frame else { return }
+
+            invalidateIntrinsicContentSize()
```
kylebshr, line 274: I would have thought that UIKit does this automatically; maybe I'm wrong

kylebshr (approved):
Does this automatically solve our perf issues with preferredContentSize?

kyleve:
> Does this automatically solve our perf issues with preferredContentSize?

Ish. Kinda. But not enough. In my testing, it seems to help with about 50% of the excess perf, since we now cache some calls to `sizeThatFits`, but because in many workflows, elements are re-rendered pretty often, the cache is still evicted too frequently. Even in the best case, this still only will get us down to 1 extra measurement, which can be 200-500ms+ for complex screens.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Frame and bounds are actually derived from the auto layout engine when using it, so AFAIK this doesn't happen automatically.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Yeah, was thinking about that. It does seem like we never call one without the other.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,45 +183,109 @@ public final class BlueprintView: UIView {
...
-        let intrinsicContentSize = resolveIntrinsicContentSize()
-        cachedIntrinsicContentSize = intrinsicContentSize
-        return intrinsicContentSize
-    }
+        let constraint = { () -> SizeConstraint in
+            if bounds.width == 0 {
+                return .unconstrained
+            } else {
+                return .init(width: bounds.width)
+            }
```
n8chur, line 248: Should we also be doing this for height?

n8chur:
> *One more thing!*
> 
> The old impl of intrinsicContentSize would use an unconstrained width if we had a new element, but had not yet laid out. This seems really wrong to me, since it was very inconsistent and hard to predict, to the point of non-determinism: Had you laid out yet? Great! Your constraint is the view width! Had you not? Uhh, you have no width constraint. This is a behaviour change, and there was value to the old behaviour, so it might make sense to add an optional/mode for this. Thoughts?

I'm not sure I follow—what is the functionality you'd be missing out on by omitting this mode? Couldn't you still get this same behavior by attempting a layout with a 0-width frame with the changes you have in this PR? I'm not sure I understand the value of the current state over what's here.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: We'd have to switch on the content hugging priority I think; since without a passed in constraint or layout priority, we'd have to base it on those.

kyleve:
> I'm not sure I follow—what is the functionality you'd be missing out on by omitting this mode? 

I think (unintentionally or not), this previously allowed Swift previews to always display the natural size of an embedded blueprint view. @bencochran added the existing behaviour and could speak more to it

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Done

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,45 +183,109 @@ public final class BlueprintView: UIView {
...
+
+        return measurement
+    }
+
+    ///
+    /// Measures the size needed to display the view within then given constraining size,
+    /// by measuring the current `element` of the `BlueprintView`.
+    ///
+    /// If you would like to not constrain the measurement in a given axis,
+    /// pass `0.0` or `.greatestFiniteMagnitude` for that axis.
```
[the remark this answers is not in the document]
kyleve: I honestly don't know how these work at all. I think as a medium, I'll make sure the `withHorizontalFittingPriority:,verticalFittingPriority:` vary based on priority.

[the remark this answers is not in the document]
kyleve: (Also, since these are both sizes, not just dimensions, and we don't really have "content hugging" or resistance in blueprint as a concept, I don't think it's super useful...)

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: I am going to remove this. The docs say: 

> This intrinsic size **must be independent of the content frame**, because there’s no way to dynamically communicate a changed width to the layout system based on a changed height, for example.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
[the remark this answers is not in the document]
kyleve: Actually, yeah, I'm not sure theres a lot for us to do here. Only the size is really useful here for us – the priorities mean nothing without a constraint system.

kyleve:
This should be good for a re-review.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,26 +183,72 @@ public final class BlueprintView: UIView {
...
-        cachedIntrinsicContentSize = intrinsicContentSize
-        return intrinsicContentSize
-    }
+        let constraint = { () -> SizeConstraint in
+            if bounds.width == 0 {
+                return .unconstrained
+            } else {
+                return .init(width: bounds.width)
+            }
+        }()
```
watt, line 249: optional nit: IMO immediately invoked closures are hard to read, and this one can be avoided

```suggestion
        let constraint: SizeConstraint = bounds.width.isZero
            ? .unconstrained
            : SizeConstraint(width: bounds.width)
```

watt (commented):
I'm a little concerned that the interactions between various functions, caches, and cache invalidations are intertwined and hard to reason about. Perhaps it was just as bad before and I just got used to it.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -226,15 +276,17 @@ public final class BlueprintView: UIView {
...
+
+        if needsViewHierarchyUpdate { return }
+
         needsViewHierarchyUpdate = true
 
-        /// We currently rely on CA's layout pass to actually perform a hierarchy update.
+        /// We use `UIView`'s layout pass to actually perform a hierarchy update.
+        /// If a manual update is required, call `layoutIfNeeded()`.
         setNeedsLayout()
     }
```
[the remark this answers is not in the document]
kyleve: I had them separated before, but Kyle rightly pointed out that we called the two methods in all the same places – so _not_ unifying to one method was more error-prone for future callsites. Maybe worth renaming this method instead? Or something...

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -36,7 +36,7 @@ public final class BlueprintView: UIView {
 
     private let rootController: NativeViewController
 
-    private var cachedIntrinsicContentSize: CGSize? = nil
+    private var sizesThatFit: [SizeConstraint: CGSize] = [:]
```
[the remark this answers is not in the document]
kyleve: We _could_; but those are currently all short-lived because they depend on the exact element state at call time (eg, the subtrees are stored by element index); so it seemed to make more sense to introduce a top-level cache that holds "inert" measurement values. Eg, we'd still need to invalidate the `CacheTree` in all the same places we do here, and keeping one around seems to potentially invite mis-use when all we really care about here is the top-level size; not digging into child sizes.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -70,6 +70,11 @@ public struct Environment {
         }
     }
 
+    /// If the `Environment` contains any values.
+    public var isEmpty: Bool {
```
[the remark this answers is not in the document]
kyleve: Hmm, true; though in that case there's no values stored; so in the cases where this is used the default values would always be the same anyway. I guess this probably isn't useful outside of Blueprint though so I can make it internal.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,26 +183,72 @@ public final class BlueprintView: UIView {
...
+                return .unconstrained
+            } else {
+                return .init(width: bounds.width)
+            }
+        }()
 
-    public override func invalidateIntrinsicContentSize() {
-        cachedIntrinsicContentSize = nil
-        super.invalidateIntrinsicContentSize()
+        return sizeThatFits(constraint.maximum)
```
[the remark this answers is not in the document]
kyleve: Sure;  will do; though note that `sizeThatFits` already handles this case because people call `sizeThatFits` with tons of weird "unconstrained" values.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Will go with an inline `func constraint() -> SizeConstraint {` since Xcode's autocomplete doesn't seem to indent this correctly.

BlueprintUI/Sources/Environment/Environment.swift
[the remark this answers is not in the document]
kyleve: Done

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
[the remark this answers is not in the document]
watt: > we'd still need to invalidate the CacheTree in all the same places we do here

Yep, that's exactly what I had in mind. The `CacheTree` has an advantage in that a top-level miss can still have nested cache hits.

This is fine as-is though; I'll explore converting to `CacheTree` later.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -179,26 +180,80 @@ public final class BlueprintView: UIView {
...
         }
 
-        return element.content.measure(
-            in: measurementConstraint(with: size),
+        return sizeThatFits(measurementConstraint(with: fittingSize))
+    }
+
+    /// Measures the size needed to display the view within the given `SizeConstraint`.
+    /// by measuring the current `element` of the `BlueprintView`.
+    public func sizeThatFits(_ constraint: SizeConstraint) -> CGSize {
```
watt, line 188: internal?

watt (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: It felt reasonable to expose publicly; since the `SizeConstraint` through its types (eg, `Axis.unconstrained`) is clearer about how to use the provided `SizeConstraint` vs the regular `CGSize` magic numbers for unconstrained measurements.

bencochran: Yeah, I like this for the same reason as the original suggestion to make it explicit. That is, for consumers who know they have a blueprint view, they can do unconstrained measurement without jumping through greatest finite magnitude weirdness.

PR #278, opened 2021-11-05
Text links

kylebshr:
Add support for the `link` attribute in `AttributedLabel`. Also supports detecting addresses, phone numbers, urls, and dates (similar to UITextView).

| Touches | Voice over |
|-|-|
| Kapture 2021-12-13 at 18 37 29 | https://user-images.githubusercontent.com/3526783/145923409-299d50b6-347a-4ea6-b0f8-4593f178af63.mp4 |

- [x] Code clean up
- [x] Documentation
- [x] Voice Over
- [x] URL handling in through environment
- [x] Tests


AttributedLabel+Link.swift
```diff
@@ -0,0 +1,49 @@
...
+            perform = onTap
+            self.range = range
+        }
+
+        public func hash(into hasher: inout Hasher) {
+            hasher.combine(hashValue)
+        }
+
+        public static func == (lhs: AttributedLabel.Link, rhs: AttributedLabel.Link) -> Bool {
+            rhs.hashValue == lhs.hashValue
```
kylebshr, line 46: should include the range

AttributedLabel+Link.swift
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public init(range: NSRange, onTap: @escaping () -> Void) {
+            hashValue = UUID().hashValue
+            perform = onTap
+            self.range = range
+        }
+
+        public func hash(into hasher: inout Hasher) {
+            hasher.combine(hashValue)
```
kylebshr, line 42: Should include the range

n8chur:
Approach overall seems fine to me for extending this component to support links.

What is the ideal end-goal for the developer experience of building a string with a link (or even other attributes, like underlined, bold, italic, etc.)? Would we add layer on top of this or a convenience for setting up links using Markdown for example?

AttributedLabel+Link.swift
tyten: I know it's still in draft state at this point, but can we be sure to add why a `Link` has a range? Reading further through this I get why, but I think in the future it'll help us out when we read this.

tyten:
I like this approach. I do see that we are setting a default color of `systemBlue` on the links. Are we going to make this rely on tokens/style in the future?

AttributedLabel+Link.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+
+extension AttributedLabel {
+    public enum LinkDetectionType: Equatable, Hashable {
```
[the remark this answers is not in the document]
kyleve: Yeah, this is nice! (though we'll definitely want to make it opt in, the perf isn't free here)

AttributedLabel+Link.swift
```diff
@@ -0,0 +1,49 @@
...
+            case .link: return .link
+            case .phoneNumber: return .phoneNumber
+            }
+        }
+    }
+
+    public struct Link: Equatable, Hashable {
+        internal var perform: () -> Void
+
+        public private(set) var hashValue: Int
```
kyleve, line 23: What's the point of this vs storing and hashing the URL?

AttributedLabel+Link.swift
kyleve, line 46: This impl is wrong – the hash value being equal does not guarantee equality. Should include the URL as well.

AttributedLabel+Link.swift
```diff
@@ -0,0 +1,49 @@
...
+        public var range: NSRange
+
+        public init(url: URL, range: NSRange) {
+            hashValue = url.hashValue
+            self.range = range
+
+            perform = {
+                UIApplication.shared.open(url, options: [:], completionHandler: nil)
```
kyleve, line 31: Thought: It might be nice to read a `URLHandler` from the `Environment`, which can default to `UIApplication`. That allows for easy stubbing of the handler (eg, maybe using it to show an in-app browser, etc).

kylebshr: Good idea

AttributedLabel+Link.swift
kylebshr: Good catch; there's a version that takes in a closure and uses a UUID for equatable/hashable, I suppose I need to store either the URL or UUID, not just the hash value

AttributedLabel+Link.swift
kylebshr: There's an init that takes in a closure instead of a URL and uses a UUID - I forgot how hashing works though, need to change this

kylebshr:
> I do see that we are setting a default color of systemBlue on the links. Are we going to make this rely on tokens/style in the future?

Absolutely, this is in Blueprint so I picked a sane default; `MarketLabel` will have properties for customizing that are `MarketColor` (perhaps we default to the `textColor` in `MarketLabel` so you only need to set it if you're using links)

AttributedLabel+Link.swift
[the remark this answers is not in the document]
kylebshr: It is opt-in!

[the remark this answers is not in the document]
kylebshr: And should only be re-detecting when needed

AttributedLabel+Link.swift
kyleve: thats why they pay me the medium bucks

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +56,560 @@ public struct AttributedLabel: Element, Hashable {
...
+                    links.append(.init(url: url, range: range))
+                } else if let urlString = url as? String,
+                          let url = URL(string: urlString)
+                {
+                    links.append(.init(url: url, range: range))
+                }
+            }
+
+            string.enumerateAttribute(
+                .labelLink,
```
kylebshr, line 318: Remove this

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +56,560 @@ public struct AttributedLabel: Element, Hashable {
...
+        /// The range of text over which to render the link.
+        public var range: NSRange
+
+        /// Create a link with a URL.
+        ///
+        /// - Parameters:
+        ///   - url: The url to open.
+        ///   - range: The range of text over which to render the link.
+        public init(url: URL, range: NSRange) {
+            identifier = url.hashValue
```
n8chur, line 116: (fine to keep as is): Any reason not to just use `UUID()` here as well? It seems a little weird that two links in different locations (with matching URLs) would share an `identifier`.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +56,560 @@ public struct AttributedLabel: Element, Hashable {
...
+
+        override func touchesCancelled(_ touches: Set<UITouch>, with event: UIEvent?) {
+            trackingLinks = nil
+            applyLinkColors()
+        }
+    }
+
+    private final class LinkAccessibilityElement: UIAccessibilityElement {
+        private let link: AttributedLabel.Link
+        private weak var container: LabelView?
```
n8chur, line 478: Nit/suggestion (feel free to ignore):
```swift
private var container: LabelView? { accessibilityContainer as? LabelView }
```

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +56,560 @@ public struct AttributedLabel: Element, Hashable {
...
+
+        init(
+            container: LabelView,
+            label: String,
+            link: AttributedLabel.Link
+        ) {
+            self.link = link
+            self.container = container
+            super.init(accessibilityContainer: container)
+            accessibilityLabel = label
```
n8chur, line 488: Do we also need to set traits on this? (I don't think we do, but wanted to double check)

n8chur (approved):
Looks great overall!

(as we discussed) I'm not sure whether or not we need the closure-based Link, its seems like we could just always support passing a URL or string directly to the link handler. Either seems fine, but if the purely identifier-based approach works it might be nice to simplify this a bit!

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: This is all deleted

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: No, setting them didn't change anything

SampleApp/Sources/TextLinkViewController.swift
```diff
@@ -0,0 +1,45 @@
...
+            $0.addFixed(child: label)
+
+            $0.addFixed(child: Label(text: "Custom link handling:"))
+
+            $0.addFixed(
+                child: label.openLink {
+                    self.presentAlert(message: $0)
+                }
+            )
+        }
```
n8chur, line 35: Nit: Could we use the result builder / in-line syntax here?

n8chur (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+                        self.accessibilityLinkIndex = max(0, self.accessibilityLinkIndex)
+
+                        let link = self.accessibilityLinks[self.accessibilityLinkIndex]
+                        return UIAccessibilityCustomRotorItemResult(targetElement: link, targetRange: nil)
+                    },
+                ]
+            }
+        }
+
+        var linkHandler: LinkHandler?
```
kyleve, line 153: Should this be `weak`? Eg, if someone makes a view controller be their link handler; and the vc retains the blueprint view, and thus this view, we'll have a retain cycle.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+                }
+                accessibilityTraits = union
+            }
+        }
+
         override func drawText(in rect: CGRect) {
             super.drawText(in: rect.offsetBy(dx: textRectOffset.horizontal, dy: textRectOffset.vertical))
         }
+
+        func makeTextStorage() -> NSTextStorage? {
```
kyleve, line 192: Do we only make these if there's links to render?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+            }
+
+            for link in activeLinks {
+                mutableString.addAttributes(activeLinkAttributes, range: link.range)
+            }
+
+            attributedText = mutableString
+        }
+
+        override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
```
kyleve, line 369: I see  a`UIGestureRecognizerDelegate` conformance above – is that still needed given we're using these methods directly?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+struct DefaultLinkHandler: LinkHandler {
+    func open(link: String) {
+        if let url = URL(string: link) {
+            UIApplication.shared.open(url, options: [:], completionHandler: nil)
+        }
+    }
+}
+
+public struct LinkHandlerEnvironmentKey: EnvironmentKey {
+    public static var defaultValue: LinkHandler {
```
kyleve, line 466: Minor nit: Just `public static let defaultValue = LinkHandler()`?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+
+    func open(link: String) {
+        openLink(link)
+    }
+}
+
+extension AttributedLabel {
+    /// Handle links opened in the receiver using the provided closure.
+    ///
+    public func openLink(_ closure: @escaping (String) -> Void) -> Element {
```
kyleve, line 490: Maybe `onLinkTapped` or something?

kyleve, line 490: Should this be a `String` or a `URL`?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+        AdaptedEnvironment(
```
kyleve, line 491: Minor nit: Would be a bit more readable here to use

`self.adaptedEnvironment(\.linkHandler, ClosureLinkHandler(openLink: closure))`?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -1,4 +1,5 @@
 import BlueprintUI
```
kyleve, line 1: General question / thought: if you haven't already, it would be worth running these changes through instruments, plus ensuring that when there aren't links, we're not going through newer expensive code paths for handling links? Given this is the base of every label in the app, ensuring we're not adding additional work for the 99% case of no links will be super important

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,508 @@ public struct AttributedLabel: Element, Hashable {
...
+
+// MARK: Environment
+
+/// Conform to this protocol to handle links tapped in an `AttributedLabel`.
+///
+/// Use the `LinkHandlerEnvironmentKey` or `Environment.linkHandler` property to override
+/// the link handler in the environment.
+///
+public protocol LinkHandler {
+    func open(link: String)
```
kyleve, line 454: Similar to below, any reason for this to be a string vs URL?

SampleApp/Sources/TextLinkViewController.swift
kylebshr: - [ ] Use result builders in demo

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: It's a protocol, unless we force it to be a class it can't be weak. Wouldn't this problem exist for any reference type putting itself into its Blueprint's environment though? The cycle exists there without this I think (I could be wrong?)

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: Yes

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - Remove `UIGestureRecognizerDelegate`

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] static let

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Rename openLink

kylebshr: Both `NSAttributedString` and cmark don't force links to be actual URLs, so I didn't want to place that restriction here and have nothing happen if you put a non-url string in there

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Improve wrapper

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Profile

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Hmm, that's true; good point... Fair enough then! This seems fine.

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Hmm interesting, OK – it _seems_ ok to require this, but I guess people could in theory put UUIDs etc in here too.

kylebshr: Yeah, or if they just want tap handling they could put identifiers/enums for different links in it then switch over those

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+            trackingLinks = nil
+            applyLinkColors()
+        }
+    }
+
+}
+
+extension AttributedLabel {
+    struct Link: Equatable, Hashable {
+        var text: String
```
watt, line 405: Hmmmm, I think my preference would be to only support URLs here, rather than having a mushier API.

If we find that folks need more flexibility, rather than supporting stringly-typed IDs, that indicates to me that we need to add a closure-based API.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+        }
+
+        override func accessibilityActivate() -> Bool {
+            container?.linkHandler?.onTap(link: link.text)
+            return true
+        }
+    }
+}
+
+// MARK: Environment
```
watt, line 446: nit: Can we move some of this stuff to separate files?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+        var linkHandler: LinkHandler?
+
+        func update(model: AttributedLabel, linkHandler: LinkHandler?, isMeasuring: Bool) {
+            let previousString = attributedText?.string
+
+            linkAttributes = model.linkAttributes
+            activeLinkAttributes = model.activeLinkAttributes
+            linkDetectionTypes = model.linkDetectionTypes ?? []
+            attributedText = model.attributedText
+                .applyingDefaultFont()
```
watt, line 162: Curious about this — are we relying on the presence of font attributes?

watt (approved):
I'd be interested in talking more about URLs vs strings, but otherwise LGTM!

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+        /// a different link.
+        private var trackingLinks: [Link]?
+
+        private var linkDetectionTypes: Set<LinkDetectionType> = []
+        private var linkAttributes: [NSAttributedString.Key: Any] = [:]
+        private var activeLinkAttributes: [NSAttributedString.Key: Any] = [:]
+
+        private var links: [Link] = [] {
+            didSet {
+                isUserInteractionEnabled = !links.isEmpty
```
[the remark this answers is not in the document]
kylebshr: Ah yeah, I don't think it's that important - was more of an issue when I was using a gesture recognizer I think

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
         }
 
+        /// The view needs to keep track of the current index for the accessibility rotor.
+        private var accessibilityLinkIndex = -1
+
+        /// These elements need to be retained by the view, and cannot be created inside the
+        /// `accessibilityCustomRotors` getter.
+        private var accessibilityLinks: [LinkAccessibilityElement] = []
+
+        override var accessibilityCustomRotors: [UIAccessibilityCustomRotor]? {
```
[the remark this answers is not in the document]
kylebshr: camera-take-a-picture

BlueprintUICommonControls/Sources/AttributedLabel.swift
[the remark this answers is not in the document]
kylebshr: - [x] Remove interaction

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+            linkDetectionTypes = model.linkDetectionTypes ?? []
+            attributedText = model.attributedText
+                .applyingDefaultFont()
+                .replacingLinkAttributes()
+            numberOfLines = model.numberOfLines
+            textRectOffset = model.textRectOffset
+            isAccessibilityElement = model.isAccessibilityElement
+            self.linkHandler = linkHandler
+
+            if !isMeasuring, previousString != attributedText?.string {
```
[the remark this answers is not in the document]
kylebshr: `previousString` stores the string before we update `attributedText`, then we update it from the `model` before this comparison; so this is just checking to see if the string actually changed. Realizing this is the incorrect thing to do though, since we're checking for links in the attributes. Need to update.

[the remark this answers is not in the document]
kylebshr: - [ ] Fix this check

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+
+            let labelSize = bounds.size
+            let textBoundingBox = layoutManager.usedRect(for: textContainer).offsetBy(
+                dx: textRectOffset.horizontal,
+                dy: textRectOffset.vertical
+            )
+            let textContainerOffset = CGPoint(
+                x: (labelSize.width - textBoundingBox.size.width) * 0.5 - textBoundingBox.origin.x,
+                y: (labelSize.height - textBoundingBox.size.height) * 0.5 - textBoundingBox.origin.y
+            )
```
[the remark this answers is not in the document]
kylebshr: You are totally correct, except if the text is centered or right aligned and is smaller than the labels bounds. Meant to come back and fix this - will push up a commit that accounts for those.

- [ ] Fix text rect offset

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: ok, I'll update it to only use URLs

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: The attributed string needs a font attribute specified across the whole thing in order for the layout manager to match the label - I initially was just setting the attribute of the attributed string when I measure it, and setting it to the `font` property, but the rect was off with that approach (not really sure why, but I think the default `font` is special in some way). Explicitly using a default font for the label like this fixes that issue (and matches the default font of UILabel appearance-wise).

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Move to separate file

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Only use urls

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
             super.drawText(in: rect.offsetBy(dx: textRectOffset.horizontal, dy: textRectOffset.vertical))
         }
+
+        func makeTextStorage() -> NSTextStorage? {
+            guard let attributedText = attributedText, attributedText.length > 0 else {
+                return nil
+            }
+
+            let textStorage = NSTextStorage()
+            let layoutManager = NSLayoutManager()
```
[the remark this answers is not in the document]
kylebshr: - [ ] usesFontLeading

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -43,53 +57,502 @@ public struct AttributedLabel: Element, Hashable {
...
+            let textContainer = NSTextContainer()
+
+            textContainer.lineFragmentPadding = 0
+            textContainer.lineBreakMode = lineBreakMode
```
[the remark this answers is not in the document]
kylebshr: - [ ] Number of lines

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Oh, actually, I can tell you why. NSAttributedString has a [defined default font of 12-point Helvetica](https://developer.apple.com/documentation/foundation/nsattributedstring/key/1528839-font).

> If you do not specify this attribute, the string uses a 12-point Helvetica(Neue) font by default.

And UILabel has its own default, probably because it predates attributed strings.

kylebshr: Right, I didn't expect the defaults to match; but adding `self.font` as the `font` attribute when before feeding the string to layout manager wasn't working as expected, which is the part I didn't really get. It's possible I was doing something else wrong (this was early in my explorations) but this seems like a good "safe" way to make sure they match

PR #279, opened 2021-11-09
Improve result builder error messages

kyleve:
Because we have `configure` and result builder variants; the compiler gets confused if you try to return an optional value from a result builder with trailing closure syntax, and gives a non-obvious error. Adding these overloads provides a clearer error.

Before:
image

After:
image

### DIFF
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -23,6 +23,15 @@ public struct Builder<Child> {
         [child]
     }
 
+    @available(
+        *,
+        unavailable,
+        message: "Optionals must be unwrapped with `if let value = ...` to be returned from a result builder."
+    )
+    public static func buildExpression(_ child: Child?) -> Children {
+        fatalError()
+    }
+
     /// Allow for an array of `Child` to be flattened into the overall result.
     public static func buildExpression(_ children: [Child]) -> Children {
         children
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -29,6 +29,16 @@ public struct Builder<Child> {
         message: "Optionals must be unwrapped with `if let value = ...` to be returned from a result builder."
     )
     public static func buildExpression(_ child: Child?) -> Children {
+
+        /// This method ensures better compile-time error messages are shown when
+        /// returning an `Optional` from a result builder. Without this method,
+        /// when an `Optional` is returned from a result builder, the compiler
+        /// will instead resolve the `configure`-based version of our component
+        /// initializers, resulting in a confusing error message.
+        ///
+        /// Adding this explicit override ensures that the compiler continues to attempt to resolve
+        /// the trailing closure as a result builder-based closure, and returns the above error message.
+
         fatalError()
     }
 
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/ElementBuilder.swift
```diff
@@ -7,6 +7,15 @@ extension ElementBuilder {
         [Child(element)]
     }
 
+    @available(
+        *,
+        unavailable,
+        message: "Optionals must be unwrapped with `if let value = ...` to be returned from a result builder."
```
ldstreet, line 13: Is it necessary to have both of these? Shouldn't the first one suffice?

ldstreet (approved):
Nice!

BlueprintUI/Sources/Layout/ElementBuilder.swift
kyleve: One checks `Element` explicitly, and the other checks `Child`, which for most of these might be `OverlayChild`, etc. Without both, you could end up without the old error message.

ldstreet: Oh right...forgot that Elements were implicitly converted.

BlueprintUI/Sources/Layout/Builder.swift
```diff
@@ -23,6 +23,15 @@ public struct Builder<Child> {
         [child]
     }
 
+    @available(
+        *,
+        unavailable,
+        message: "Optionals must be unwrapped with `if let value = ...` to be returned from a result builder."
+    )
+    public static func buildExpression(_ child: Child?) -> Children {
+        fatalError()
```
n8chur, line 32: Does this work?
```suggestion
        guard let child = child else {
            return []
        }
        return child
```

kyleve: This code can never be reached though – so implementing this provides no value.

kylebshr: Would it work if we removed `configure`? Or if `configure` is used, are you sure it doesn't disambiguate and then go down this path?

kyleve: If we remove `configure`, we'd have to go through and migrate everyone to the result builder syntaxes, which seems like a lot of work I'd rather not sign up for. If the configuration variant of the methods isn't there, we get a more useful error message. The problem here is b/c the result builder infra cant resolve a method to handle the optional, the type checker looks for another method that would work, which then finds the `configure` variation.

kylebshr (approved):

BlueprintUI/Sources/Layout/Builder.swift
n8chur: Ohhh I see the issue... Hm...

We should probably add a comment here why this is useful for us since it's not obvious.

Out of curiosity, does providing the parameter name for the builder resolve the compiler error as well?

ldstreet: I want to say I've seen this ambiguous error when I get syntax wrong inside the closure such as mislabeled/missing parameter. Removing `configure` seems like a good idea to me long term, but a huge undertaking for now. What about deprecating it to discourage future use and encourage refactoring existing uses?

kyleve: There's thousands upon thousands of usages of the configure syntax now – deprecating it is going to make a lot of fairly useless noise.

kyleve: > Out of curiosity, does providing the parameter name for the builder resolve the compiler error as well?

Yessir

PR #280, opened 2021-11-09
Prep 0.31.0

kyleve:


kylebshr (commented):
Not to be pedantic, but this isn't a minor bug fix so it should be `0.31.0`

kyleve:
True, I guess I considered the caching and better error messages "fixes", not really new un-breaking features, but 0.31.0 is easy enough...

kyleve:
Switched to 0.31.0, if anyone feels strongly I can switch it back to 0.30.1 too

n8chur (approved):
Make sure you update the PR description if you decide to keep it `0.31.0`

PR #282, opened 2021-11-10
Improve assertion messages for unconstrained values

kyleve:
Improve the unconstrained assertion messages – these seem to always confuse people.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -241,10 +241,11 @@ public final class BlueprintView: UIView {
     private func updateViewHierarchyIfNeeded() {
         guard needsViewHierarchyUpdate || bounds != lastViewHierarchyUpdateBounds else { return }
 
-        assert(
+        precondition(
             !isInsideUpdate,
             "Reentrant updates are not supported in BlueprintView. Ensure that view events from within the hierarchy are not synchronously triggering additional updates."
         )
+
         isInsideUpdate = true
 
         needsViewHierarchyUpdate = false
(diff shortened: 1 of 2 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -241,10 +241,11 @@ public final class BlueprintView: UIView {
     private func updateViewHierarchyIfNeeded() {
         guard needsViewHierarchyUpdate || bounds != lastViewHierarchyUpdateBounds else { return }
 
-        assert(
+        precondition(
```
kyleve, line 244: This seems like it should always fire, even in prod.

kylebshr: err, do you think so? does this cause undefined behavior or will things still work? if things still work, not crashing seems better

kyleve: Generally we want our assertions to align between prod and debug (this is why POS includes `NSAssert` in prod); and yes it could cause undefined behavior because we do not actually ever test this case.

n8chur (approved):

kyleve:
Merging to start prepping the release, can change the messages if anyone has strong opinions!

BlueprintUI/Sources/Layout/LayoutAttributes.swift
```diff
@@ -152,19 +152,54 @@ public struct LayoutAttributes {
...
+            center.isFinite,
+            "LayoutAttributes.center must only contain finite values."
+        )
     }
 
     private func validateTransform() {
-        assert(transform.isFinite, "LayoutAttributes.transform only not contain finite values.")
+        assert(
+            transform.isFinite,
+            "LayoutAttributes.transform only not contain finite values."
```
watt, line 194: this one could probably benefit from a grammar pass, lol

kyleve: Omg how did I not see this

PR #285, opened 2021-11-15
Fix FocusState retain cycle

watt:
Fixing a retain cycle in the storage of FocusState. Also fixing a retain cycle in the PostsViewController demo that I ran into while testing this out.

### DIFF
--- a/BlueprintUI/Sources/Focus/FocusState.swift
+++ b/BlueprintUI/Sources/Focus/FocusState.swift
@@ -117,16 +117,12 @@ public struct FocusState<Value> where Value: Hashable {
 
     /// Creates a focus state that binds to a Boolean.
     public init() where Value == Bool {
-        _value = Storage(wrappedValue: false)
-        // In Xcode 12.5 use this instead:
-        // value = false
+        value = false
     }
 
     /// Creates a focus state that binds to an optional type.
     public init<T>() where Value == T?, T: Hashable {
-        _value = Storage(wrappedValue: nil)
-        // In Xcode 12.5 use this instead
-        // value = nil
+        value = nil
     }
 
     /// The current state value, taking into account whatever bindings might be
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/Focus/FocusState.swift
```diff
@@ -157,24 +153,7 @@ public struct FocusState<Value> where Value: Hashable {
...
-                if self.value == value {
-                    self.value = storage.defaultValue
-                }
-            }
-        )
-
-        storage.bindings[value] = binding
-
-        return binding
+        storage.binding(for: value)
```
watt, line 156: moving this to `Storage` made the weakifying dance a bit less awkward

BlueprintUI/Tests/FocusStateTests.swift
```diff
@@ -72,4 +72,16 @@ final class FocusStateTests: XCTestCase {
         binding.onBlur()
         XCTAssertEqual(state.isFocused, false)
     }
+
+    func test_memoryRelease() {
```
kyleve, line 76: Not sure if it's worth it, but I always copy around this function, which waits for the release/dealloc to happen over a couple of runloop cycles; given the autorelease pool timing seem to change between iOS releases (or based on our code, depending on when values get set), which makes the tests more stable: https://github.com/squareup/market/blob/e7f264ca2ab8a7d1bcade4a7b5fe179e2f41302a/ios/MarketUITesting/Sources/XCTestCase%2BTestHelpers.swift#L68

kyleve (approved):

n8chur (approved):

BlueprintUI/Tests/FocusStateTests.swift
watt: There should be no need for any autorelease pool here! AFAIK this test is deterministic and guaranteed by ARC semantics.

PR #286, opened 2021-11-15
Add support for for loop and available checks in result builders

kyleve:
- This fixes some build errors when you try to use `for...in` loops or `available` checks in result builders. I copied the tests over from Listable to ensure everything works on the base type.

### DIFF
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -43,7 +43,7 @@ public struct Builder<Child> {
     }
 
     public static func buildArray(_ components: [Children]) -> Children {
-        components.reduce(into: []) { $0 += $1 }
+        components.flatMap { $0 }
     }
 
     public static func buildArray(_ components: Children) -> Children {
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -42,6 +42,18 @@ public struct Builder<Child> {
         fatalError()
     }
 
+    public static func buildArray(_ components: [Children]) -> Children {
+        components.reduce(into: []) { $0 += $1 }
+    }
+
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/Layout/Builder.swift
```diff
@@ -42,6 +42,18 @@ public struct Builder<Child> {
         fatalError()
     }
 
+    public static func buildArray(_ components: [Children]) -> Children {
+        components.reduce(into: []) { $0 += $1 }
```
ldstreet, line 46: I think this might be a little more efficient:
```suggestion
        components.flatMap { $0 }
```

BlueprintUI/Sources/Layout/Builder.swift
```diff
@@ -42,6 +42,18 @@ public struct Builder<Child> {
...
+    }
+
+    public static func buildArray(_ components: Children) -> Children {
+        components
+    }
+
+    public static func buildLimitedAvailability(_ component: Children) -> Children {
```
ldstreet, line 53: Cool 👍

ldstreet (approved):

BlueprintUI/Sources/Layout/Builder.swift
kyleve: I tested this and didn't find any particularly large difference

ldstreet: Yeah I'm sure its not too much different, especially on arrays with such few elements. I do think `flatMap` is more semantically appropriate in this scenario since we are flattening a two dimensional array into a single level collection which jives with Apple's [documentation](https://developer.apple.com/documentation/swift/sequence/2905332-flatmap). But its not so important to me that I'd debate it any more than writing this post 😆

kyleve: True, I can switch it up

PR #288, opened 2021-11-18
Allow specifying an explicit constraint when measuring in a GeometryProxy

kyleve:
In Market, we have this method:

```
extension GeometryProxy {
    func element(
        ifFitsWidth: @autoclosure () -> Element,
        else fallback: @autoclosure () -> Element
    ) -> Element {
        let ifFitsWidthElement = ifFitsWidth()
        
        if measure(element: ifFitsWidthElement).width <= constraint.maximum.width {
            return ifFitsWidthElement
        } else {
            return fallback()
        }
    }
}
```

Which utilizes `measure(element:)` on `GeometryProxy`. However, this is not actually a correct implementation of `ifFitsWidth`. The problem is we're not actually passing an `.unconstrained` width to the element during this measurement, we're passing the constraint of the `GeometryProxy`:

```
public struct GeometryProxy {
    var environment: Environment

    /// The size constraint of the element being laid out.
    public var constraint: SizeConstraint

    /// Measure the given element, using this proxy's constraint and the current environment.
    public func measure(element: Element) -> CGSize {
        element.content.measure(in: constraint, environment: environment)
    }
}
```

So thus, if the `ifFitsWidth` contains any sort of element (eg, a text label) which will "re-flow" when it hits a width constraint, the measurement will result in a width `<= constraint.maximum.width`, and we will return "yes, I do fit the width!", when in reality, it does not. 

Thus, introduce a way to specify an explicit constraint when calling `measure`; while still measuring within the `GeometryProxy`'s `Environment`.

### DIFF
--- a/BlueprintUI/Sources/Layout/GeometryReader.swift
+++ b/BlueprintUI/Sources/Layout/GeometryReader.swift
@@ -48,8 +48,8 @@ public struct GeometryProxy {
     /// The size constraint of the element being laid out.
     public var constraint: SizeConstraint
 
-    /// Measure the given element, using this proxy's constraint and the current environment.
-    public func measure(element: Element) -> CGSize {
-        element.content.measure(in: constraint, environment: environment)
+    /// Measure the given element, using the provided constraint (or the proxy's constraint if not provided), and the current environment.
+    public func measure(in explicit: SizeConstraint? = nil, element: Element) -> CGSize {
+        element.content.measure(in: explicit ?? constraint, environment: environment)
     }
 }
--- a/BlueprintUI/Sources/Layout/GeometryReader.swift
+++ b/BlueprintUI/Sources/Layout/GeometryReader.swift
@@ -49,7 +49,7 @@ public struct GeometryProxy {
     public var constraint: SizeConstraint
 
     /// Measure the given element, using the provided constraint (or the proxy's constraint if not provided), and the current environment.
-    public func measure(in explicit: SizeConstraint? = nil, element: Element) -> CGSize {
+    public func measure(element: Element, in explicit: SizeConstraint? = nil) -> CGSize {
         element.content.measure(in: explicit ?? constraint, environment: environment)
     }
 }
--- a/BlueprintUI/Sources/Layout/GeometryReader.swift
+++ b/BlueprintUI/Sources/Layout/GeometryReader.swift
@@ -48,7 +48,7 @@ public struct GeometryProxy {
     /// The size constraint of the element being laid out.
     public var constraint: SizeConstraint
 
-    /// Measure the given element, using the provided constraint (or the proxy's constraint if not provided), and the current environment.
+    /// Measure the given element, constrained to the same size as the `GeometryProxy` itself (unless a constraint is explicitly provided).
     public func measure(element: Element, in explicit: SizeConstraint? = nil) -> CGSize {
         element.content.measure(in: explicit ?? constraint, environment: environment)
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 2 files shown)


BlueprintUI/Sources/Layout/GeometryReader.swift
```diff
@@ -48,8 +48,8 @@ public struct GeometryProxy {
     /// The size constraint of the element being laid out.
     public var constraint: SizeConstraint
 
-    /// Measure the given element, using this proxy's constraint and the current environment.
-    public func measure(element: Element) -> CGSize {
-        element.content.measure(in: constraint, environment: environment)
+    /// Measure the given element, using the provided constraint (or the proxy's constraint if not provided), and the current environment.
+    public func measure(in explicit: SizeConstraint? = nil, element: Element) -> CGSize {
```
kyleve, line 52: Another option is exposing the `environment`, so people can just perform their own measurements.

n8chur: Could/should we just update the `constraint` before measuring (in the consuming code) instead of making these changes?

kyleve: We could; but that's really weird (IMO) – the geometry reader's proxy is meant to represent the size of the reader itself. I'd rather not encourage folks to mutate it.

bencochran: Yeah, agreed about that feeling odd. In fact, though I’m generally a fan of struct members being `var`s across the board, I’d be in favor of `constraint` being a `let` (or `internal(set)`) here to make that distinction clearer.

bencochran: On the original topic, I prefer the approach in the PR currently over exposing `environment` here. Keeps the focus clear and is a little easier to discover IMO.

BlueprintUI/Sources/Layout/GeometryReader.swift
```diff
@@ -48,8 +48,8 @@ public struct GeometryProxy {
     /// The size constraint of the element being laid out.
     public var constraint: SizeConstraint
 
-    /// Measure the given element, using this proxy's constraint and the current environment.
-    public func measure(element: Element) -> CGSize {
-        element.content.measure(in: constraint, environment: environment)
+    /// Measure the given element, using the provided constraint (or the proxy's constraint if not provided), and the current environment.
```
bencochran, line 51: I’d maybe flip the precedence of the comments here to guide people toward the more-common approach. Maybe something like:
```
/// Measure the given element, constrained to the same size as the `GeometryProxy` itself (unless a constraint is explicitly provided).
```
(I also dropped the environment part because I think it's more of an expected implementation detail than something consumers need be concerned with)

bencochran (approved):
Makes sense to me. A more compositional approach would be to add something like `ConstrainedSize` but which _doesn't_ constrain measurement based on its incoming constraint. But a) that imaginary wrapper is maybe a bit cumbersome (slash "when do I use that vs when do I use `ConstrainedSize`?") and b) that encourages measuring a different tree than you provide during actual layout and that feels funky.

watt (approved):

n8chur (approved):

BlueprintUI/Sources/Layout/GeometryReader.swift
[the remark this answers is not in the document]
kyleve: Ah true; I had avoided putting the element first since we usually avoid that on `init` methods for container elements, but its unlikely you'll be constructing a new element here so it wont leave the orphan "in:foo" likely.

BlueprintUI/Sources/Layout/GeometryReader.swift
kyleve: Done

BlueprintUI/Sources/Layout/GeometryReader.swift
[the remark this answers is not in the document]
kyleve: Done

PR #291, opened 2021-11-20
Add an ElementContent.init(measureFunction:) variant which takes in an environment

kyleve:
Adds a new variant to `ElementContent`, where the `measureFunction` takes in an `Environment`. This allows us to stop doing things like this:

```swift
    ElementContent { constraint -> CGSize in
        makeElement(for: .normal)
            .content
            .measure(in: constraint, environment: .empty)
    }
}
```

And instead pass the correct environment:

```swift
    ElementContent { constraint, env -> CGSize in
        makeElement(for: .normal)
            .content
            .measure(in: constraint, environment: env)
    }
}
```


kylebshr (approved):

watt (approved):
LGTM.

Linking #140 for future reference. `ViewDescriptionContext` solves the other half of this.

PR #292, opened 2021-11-22
Fix a regression when using Blueprint with Autolayout

kylebshr:
Release 0.33.1 with fix

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -91,6 +91,15 @@ public final class BlueprintView: UIView {
         }
     }
 
+    /// We need to invalidateIntrinsicContentSize when the bounds change for Auto Layout to work correctly.
+    public override var bounds: CGRect {
+        didSet {
+            guard oldValue != bounds else { return }
+
+            invalidateIntrinsicContentSize()
+        }
+    }
+
     /// An optional name to help identify this view
     public var name: String?
 
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -91,10 +91,10 @@ public final class BlueprintView: UIView {
         }
     }
 
-    /// We need to invalidateIntrinsicContentSize when the bounds change for Auto Layout to work correctly.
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -91,6 +91,15 @@ public final class BlueprintView: UIView {
         }
     }
 
+    /// We need to invalidateIntrinsicContentSize when the bounds change for Auto Layout to work correctly.
+    public override var bounds: CGRect {
+        didSet {
+            guard oldValue != bounds else { return }
```
kyleve, line 97: Minorest of nits, even though I wrote this code: This could actually check `bounds.size`, not that it matters much.

kyleve (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kylebshr:
```suggestion
            guard oldValue.size != bounds.size else { return }
```

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -91,6 +91,15 @@ public final class BlueprintView: UIView {
         }
     }
 
+    /// We need to invalidateIntrinsicContentSize when the bounds change for Auto Layout to work correctly.
```
kylebshr, line 94:
```suggestion
    /// We need to invalidateIntrinsicContentSize when `bound.size` changes for Auto Layout to work correctly.
```

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -91,6 +91,15 @@ public final class BlueprintView: UIView {
         }
     }
 
+    /// We need to invalidateIntrinsicContentSize when `bound.size` changes for Auto Layout to work correctly.
+    public override var bounds: CGRect {
```
kyleve, line 95: Double thinking through this: We used to `invalidateIntrinsicContentSize` in `layoutSubviews`, is there any benefit or drawback to doing it here instead? This seems better since we now only invalidate when our bounds as applied by autolayout actually change, but the timing is slightly different from `layoutSubviews`.

kylebshr: Yeah I was wondering about that myself, but if this works it seems better to me since it'll be hit less?

PR #293, opened 2021-11-23
Prep 0.34.0

kyleve:


kylebshr (approved):

kylebshr (changes requested):
I stupidly didn't check the commit sha, and tagged a634ac9 as 0.33.1, which was your merge of the new element content init. I guess we should just update the changelog to say as much? Or can we just change the tag? 🤔

kyleve:
I would say let's just update the changelog - unpublishing / changing the tag is going to be a pain since we already published, and we also already whitelisted in our internal specs repo.

PR #295, opened 2021-11-23
SPM-friendly resource structure

watt:
As of Swift 5.3, you can [declare resources in Package.swift](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html).

This PR does that, and:
- updates the `swift-tools-version` to 5.4, matching our source compatibility
- moves test resources into a subdirectory of Sources, which feels weird to me, but [is what it wants](https://developer.apple.com/documentation/swift_packages/bundling_resources_with_a_swift_package)

Snapshot image URLs are still resolved the old way rather than the new `Bundle.module` accessor, because we need a writable URL in the checked out repo, not a URL in the compiled bundle's build directory.


kyleve (approved):

kylebshr (approved):

n8chur (approved):
Did we still want to land this? Probably needs a rebase before we merge.

watt:
Whoops! I'll rebase and merge.

PR #297, opened 2021-11-30
Adjust scroll view inset correctly and release bug fix

kylebshr:
In my basic test cases I forgot that `adjustedContentInset` _includes_ the `contentInset` - so, if the keyboard height changed, or the insets changed, and the adjustment happened again, it was incorrect. We need to use `adjustedContentInset - contentInset` to determine the applied safe area insets.

Resolves https://square.slack.com/archives/CTJ4UNLHF/p1638230791044100

### DIFF
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -357,6 +357,7 @@ fileprivate final class ScrollerWrapperView: UIView {
     private func applyContentInset(with scrollView: ScrollView) {
         let contentInset = ScrollView.calculateContentInset(
             scrollViewInsets: scrollView.contentInset,
+            contentInsets: scrollView.contentInset,
             adjustedContentInsets: self.scrollView.adjustedContentInset,
             keyboardBottomInset: bottomContentInsetAdjustmentForKeyboard,
             refreshControlState: scrollView.pullToRefreshBehavior,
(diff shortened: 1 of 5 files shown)


n8chur (approved):
Pretty sure this looks right to me!

Were you able to test this change integrated into Register in any of the reported flows where it was causing issues?

BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -407,21 +408,27 @@ extension ScrollView {
...
         keyboardBottomInset: CGFloat,
         refreshControlState: PullToRefreshBehavior,
         refreshControlBounds: CGRect?
     ) -> UIEdgeInsets {
         var finalContentInset = scrollViewInsets
 
+        // The safe area may not actually be applied to the scroll view depending on
+        // the content height and inset adjustment mode. We need to determine the actual applied
+        // safe area insets by subtracting the content inset from the adjusted content inset.
+        let appliedSafeAreaInset = adjustedContentInsets - contentInsets
```
kyleve, line 422: Do we also need to check the `contentInsetAdjustmentBehavior`; or is this right in all cases?

kylebshr: This should (finally) be right in all cases; since the `adjustedContentInset` will be adjusted based on the `contentInsetAdjustmentBehavior`

kylebshr:
@n8chur yep, I tested the change in register!

BlueprintUICommonControls/Sources/ScrollView.swift
kyleve: 👍 great!

PR #298, opened 2021-12-01
Implicitly animate the Box's shadowPath

kylebshr:
This approach is different from my fixes in Market, because I did a little more research and this seems more "correct" - rather than having a side effect of starting an animation in `layoutSubviews`, we use the same method that UIKit uses for implicit animations: returning a CAAction that runs the animation. This also ensures the system runs the animation exactly alongside the other implicit animations.

| Before | After |
|-|-|
| Kapture 2021-12-01 at 10 12 39 | Kapture 2021-12-01 at 10 14 22 |

### DIFF
--- a/BlueprintUICommonControls/Sources/Box.swift
+++ b/BlueprintUICommonControls/Sources/Box.swift
@@ -334,6 +334,8 @@ fileprivate final class BoxView: UIView {
         }
     }
 
+    /// This method is overridden to provide a `ShadowPathAction` for the `shadowPath` event;
+    /// this ensure the shadow implicitly animates alongside changes in the views size.
     override func action(for layer: CALayer, forKey event: String) -> CAAction? {
         let keyPath = #keyPath(CALayer.shadowPath)
 
--- a/BlueprintUICommonControls/Sources/Box.swift
+++ b/BlueprintUICommonControls/Sources/Box.swift
@@ -334,4 +334,37 @@ fileprivate final class BoxView: UIView {
         }
     }
 
+    override func action(for layer: CALayer, forKey event: String) -> CAAction? {
+        let keyPath = #keyPath(CALayer.shadowPath)
+
+        guard event == keyPath,
+              let currentPath = layer.shadowPath,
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -334,4 +334,39 @@ fileprivate final class BoxView: UIView {
         }
     }
 
+    /// This method is overridden to provide a `ShadowPathAction` for the `shadowPath` event;
+    /// this ensure the shadow implicitly animates alongside changes in the views size.
+    override func action(for layer: CALayer, forKey event: String) -> CAAction? {
```
kyleve, line 339: TIL!

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -334,4 +334,39 @@ fileprivate final class BoxView: UIView {
...
+              let sizeAnimation = layer.animation(forKey: "bounds.size") as? CABasicAnimation
+        else {
+            return super.action(for: layer, forKey: event)
+        }
+
+        let animation = sizeAnimation.copy() as! CABasicAnimation
+        animation.keyPath = keyPath
+        animation.fromValue = currentPath
+
+        return ShadowPathAction(pendingAnimation: animation)
```
bencochran, line 353: Can you not return `animation` itself here? I’ve done that in the past with success

kylebshr: How do you set the `toValue`? Do you not need to since the actual value is set?

kylebshr: Huh, it certainly seems to work!

bencochran: Yeah, tbh this is one of those mysterious/under-documented parts of CoreAnimation. But `toValue` seems to happen magically in there somewhere.

bencochran (approved):

n8chur (approved):

watt (approved):

PR #299, opened 2021-12-07
Introduce AttributedText

kylebshr:
`AttributedText` allows you to apply strongly-typed attributes to strings (much like the `AttributedString` type introduced in iOS 15). You can then access the `attributedString` property to get an attributed string with those attributes applied.

For example:

```swift
var text = AttributedText(string: "Hello, world")
// Apply a font to the entire range
text.font = .systemFont(ofSize: 20)

// Apply a color to part of the string
let range = text.string.range(of: "world")!
text[range].color = .blue

// Render the attributed text
let label = AttributedLabel(attributedText: text.attributedString)
```

### DIFF
--- a/BlueprintUI/Sources/AttributedText/AttributedText.swift
+++ b/BlueprintUI/Sources/AttributedText/AttributedText.swift
@@ -1,0 +1,149 @@
+import UIKit
+
+/// `AttributedText` allows you to apply strongly-typed attributes to strings (much like the `AttributedString` type
+/// introduced in iOS 15). You can then access the `attributedString` property to get an attributed string with those
+/// attributes applied.
+///
+/// For example:
+///
+/// ```swift
+/// var text = AttributedText(string: "Hello, world")
+/// // Apply a font to the entire range
+/// text.font = .systemFont(ofSize: 20)
+///
+/// // Apply a color to part of the string
+/// let range = text.string.range(of: "world")!
+/// text[range].color = .blue
+///
+/// // Render the attributed text
+/// let label = AttributedLabel(attributedText: text.attributedString)
+/// ```
+///
+@dynamicMemberLookup public struct AttributedText {
+
+    /// The wrapped string, with no attributes.
+    public let string: String
+
+    /// An `NSAttributedString` representation of the attributed text.
+    public var attributedString: NSAttributedString {
+        mutableAttributedString
+    }
+
+    private let mutableAttributedString: NSMutableAttributedString
+
+    /// Create some `AttributedText` from a plain string.
+    public init(string: String) {
+        self.string = string
+        self.mutableAttributedString = NSMutableAttributedString(string: string)
+    }
+
+    /// Create some `AttributedText` from an attributed string. The attributes are preserved.
+    public init(attributedString: NSAttributedString) {
+        self.string = attributedString.string
+        self.mutableAttributedString = NSMutableAttributedString(attributedString: attributedString)
+    }
+
+    public func range<StringType: StringProtocol>(of aString: StringType) -> Range<String.Index>? {
+        string.range(of: aString)
+    }
+
+    /// Dynamic member getter or setter for any attributes defined on `TextAttributeContainer`.
+    /// Applies the attribute to the entire range of text, for example:
+    ///
+    /// ```swift
+    /// var text = AttributedText(string: "Hello, world")
+    /// text.font = .systemFont(ofSize: 20)
+    /// ```
+    /// Note that only attributes applying to the entire range will be returned. For example, if the text
+    /// has two different `font` attributes, then `text.font` will be `nil`.
+    ///
+    public subscript<Value>(dynamicMember keyPath: WritableKeyPath<TextAttributeContainer, Value>) -> Value {
+        get {
+            self[entireRange][keyPath: keyPath]
+        }
+        set {
+            self[entireRange][keyPath: keyPath] = newValue
+        }
+    }
+
+    /// Get or set a `TextAttributeContainer` for the provided range of text. This allows you to set attributes
+    /// for specific ranges using strong types:
+    ///
+    /// ```swift
+    /// var text = AttributedText(string: "Hello, world")
+    /// let range = text.string.range(of: "Hello")!
+    /// text[range].font = .systemFont(ofSize: 20)
+    /// ```
+    ///
+    /// Note that the returned `TextAttributeContainer` will only contain attributes that apply to the entire subscript
+    /// range. (Setting an attribute will set it across the subscript range regardless of any existing contents).
+    ///
+    public subscript(range: Range<String.Index>) -> TextAttributeContainer {
+        get {
+            let range = NSRange(range, in: string)
+            return makeAttributeStore(range: range)
+        }
+        set {
+            let range = NSRange(range, in: string)
+            mutableAttributedString.addAttributes(newValue.storage, range: range)
+        }
+    }
+
+    /// Concatenate two pieces of `AttributedText` together.
+    ///
+    public static func + (lhs: AttributedText, rhs: AttributedText) -> AttributedText {
+        let newString = NSMutableAttributedString(attributedString: lhs.mutableAttributedString)
+        newString.append(rhs.mutableAttributedString)
+        return AttributedText(attributedString: newString)
+    }
+
+    private var entireRange: Range<String.Index> {
+        string.startIndex..<string.endIndex
+    }
+
+    /// The implementation of this function may not be intuitive, but is necessary due to how enumerating attributes
+    /// works. For example, consider the string "Block" with the font attribute set for the whole string, the color
+    /// red set for the range of "Blo" and the color blue set for the range of "ck".
+    ///
+    /// If we get the attributes for the entire range of the string, the desired output is a
+    /// `TextAttributeContainer` with only the `font` specified, since it's the only attribute that applies to that
+    /// whole range. If we enumerate the attributes of the string however, the system will give us two results:
+    ///
+    /// - The range of "Blo" with the font and red color in the dictionary
+    /// - The range of "ck" with the font and the blue color in the dictionary
+    ///
+    /// Now we need to "merge" attributes that match across the whole range, but in practice we cannot do that: the
+    /// dictionary contains values of type "any" so we cannot compare them. We know that a font attribute is present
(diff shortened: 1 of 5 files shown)


BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,149 @@
...
+/// ```
+///
+@dynamicMemberLookup public struct AttributedText {
+
+    /// The wrapped string, with no attributes.
+    public let string: String
+
+    /// An `NSAttributedString` representation of the attributed text.
+    public var attributedString: NSAttributedString {
+        mutableAttributedString
```
kyleve, line 29: Nit: Return a `.copy()` of this – otherwise you're returning the actual underlying mutable value, which can continue to mutate.

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,149 @@
...
+    /// range. (Setting an attribute will set it across the subscript range regardless of any existing contents).
+    ///
+    public subscript(range: Range<String.Index>) -> TextAttributeContainer {
+        get {
+            let range = NSRange(range, in: string)
+            return makeAttributeStore(range: range)
+        }
+        set {
+            let range = NSRange(range, in: string)
+            mutableAttributedString.addAttributes(newValue.storage, range: range)
```
kyleve, line 88: `NSMutableAttributedString` is a mutable reference (class) type. This struct is a class type; you need to implement copy on write semantics here otherwise each copy of the struct is gonna reference the same underlying mutable string

BlueprintUI/Sources/AttributedText/AttributedTextKey.swift
```diff
@@ -0,0 +1,97 @@
...
+        get { self[ColorKey.self] }
+        set { self[ColorKey.self] = newValue }
+    }
+}
+
+// MARK: Kern
+
+public enum KernKey: AttributedTextKey {
+    public typealias Value = CGFloat
+    public static var name: NSAttributedString.Key { .kern }
```
watt, line 49: For the built-in keys, I think we might want to expose tracking instead of kerning, via `kCTTrackingAttributeName as NSAttributedString.Key`.

Check out [this comment thread where we added it to POSUI for context](https://github.com/squareup/ios-register/pull/49844#discussion_r554402886).

BlueprintUI/Sources/AttributedText/AttributedText.swift
kylebshr: Ah yep, very good catch...

kylebshr: Added a test for this and pushed up a fix

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,158 @@
...
+    private var mutableAttributedString: NSMutableAttributedString
+
+    /// Create some `AttributedText` from a plain string.
+    public init(string: String) {
+        self.string = string
+        self.mutableAttributedString = NSMutableAttributedString(string: string)
+    }
+
+    /// Create some `AttributedText` from an attributed string. The attributes are preserved.
+    public init(attributedString: NSAttributedString) {
```
kylebshr, line 41: Should this also just be `string:`? Should both inits have no name for the parameter?

```suggestion
    public init(string: NSAttributedString) {
```

BlueprintUI/Tests/AttributedTextTests.swift
```diff
@@ -0,0 +1,107 @@
...
+        let hello = text.range(of: "Hello")!
+        let world = text.range(of: "world")!
+        let loWo = text.range(of: "lo wo")!
+        let rld = text.range(of: "rld")!
+
+        text[hello].font = font
+        text[world].color = green
+        text[loWo].color = blue
+
+        XCTAssertEqual(text[hello].font, font, "The font should be returned when it applies to the whole range")
```
watt, line 40: Can you also test that a contained range like `"ell"` and a containing range like `"hello wo"` are both nil?

watt (approved):

watt:
For anybody else who got curious about how hard a custom storage implementation would be:
- [CFAttributedString](https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/String.subproj/CFAttributedString.c)
- [the new AttributedString](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/AttributedString/AttributedString.swift)

I was expecting something like an [interval tree](https://en.wikipedia.org/wiki/Interval_tree), but both of them seem to boil down to arrays of runs.

BlueprintUI/Sources/AttributedText/AttributedText.swift
n8chur: I'm pretty sure we want to omit the parameter here based on the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#type-conversion), but I might be misunderstanding them.

n8chur: Or maybe we are supposed to put the parameter for `String` because it's "narrowing" and not put the parameter for `NSAttributedString` because it's just converting?

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,149 @@
...
+/// text[range].color = .blue
+///
+/// // Render the attributed text
+/// let label = AttributedLabel(attributedText: text.attributedString)
+/// ```
+///
+@dynamicMemberLookup public struct AttributedText {
+
+    /// The wrapped string, with no attributes.
+    public let string: String
```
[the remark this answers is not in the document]
kylebshr: Hmm I guess so, initially I didn't expose `range(of:)` so consumers could use this string directly. Maybe it doesn't need to be public yet?

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,149 @@
...
+    /// for specific ranges using strong types:
+    ///
+    /// ```swift
+    /// var text = AttributedText(string: "Hello, world")
+    /// let range = text.string.range(of: "Hello")!
+    /// text[range].font = .systemFont(ofSize: 20)
+    /// ```
+    ///
+    /// Note that the returned `TextAttributeContainer` will only contain attributes that apply to the entire subscript
+    /// range. (Setting an attribute will set it across the subscript range regardless of any existing contents).
```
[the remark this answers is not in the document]
kylebshr: If the attributes extend beyond the range they'll still be returned! It's only attributes that apply to part of the range that aren't returned

BlueprintUI/Sources/AttributedText/AttributedText.swift
kyleve: Nice

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,149 @@
...
+    ///
+    public subscript(range: Range<String.Index>) -> TextAttributeContainer {
```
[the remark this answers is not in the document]
kylebshr: Yep, that works!

BlueprintUI/Tests/AttributedTextTests.swift
kylebshr: Done (although "ell" returns the font)

BlueprintUI/Sources/AttributedText/AttributedText.swift
kylebshr: I'm for that

BlueprintUI/Sources/AttributedText/AttributedTextKey.swift
kylebshr: Done

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,158 @@
...
+        for attribute in attributesInRange {
+            mutableAttributedString.enumerateAttribute(
+                attribute,
+                in: range,
+                options: []
+            ) { value, attributeRange, _ in
+                if let value = value, attributeRange == range {
+                    store.storage[attribute] = value
+                }
+            }
```
watt, line 153: So if I'm following what you're saying here:

> If the attributes extend beyond the range they'll still be returned! It's only attributes that apply to part of the range that aren't returned

The `attributedRange` passed into this block by `enumerateAttribute` will be no larger than `range`, even if that attributes's range does extend outside of `range`?

BlueprintUI/Tests/AttributedTextTests.swift
```diff
@@ -0,0 +1,129 @@
+import XCTest
+@testable import BlueprintUI
+
+class AttributedTextTests: XCTestCase {
```
kyleve, line 4: Nit: It would also be good to add some test using emojis and non-ascii characters to validate this does what we expect there as well.

kylebshr: Done

BlueprintUI/Sources/AttributedText/AttributedText.swift
kylebshr: Yes! Although it's not documented well, that's the behavior I'm seeing (and expected, tbh)

kyleve (approved):

watt (approved):

n8chur (approved):
Nice work!

BlueprintUI/Sources/AttributedText/AttributedText.swift
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    private var entireRange: Range<String.Index> {
+        string.startIndex..<string.endIndex
+    }
+
+    private mutating func addAttributes(attributes: TextAttributeContainer, to range: NSRange) {
+        if !isKnownUniquelyReferenced(&mutableAttributedString) {
+            mutableAttributedString = NSMutableAttributedString(attributedString: mutableAttributedString)
```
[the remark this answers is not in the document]
kyleve: Shouldn't matter beyond being a bit less wordy.

PR #300, opened 2021-12-07
Support CALayerCornerCurve for box corner styles

kyleve:
`CALayerCornerCurve` is the "new" (uh, iOS 7? But finally exposed in APIs in 13) style for rounded corners on icons, etc, in iOS. Add support for specifying the corner style as such.

### DIFF
--- a/BlueprintUICommonControls/Sources/Box.swift
+++ b/BlueprintUICommonControls/Sources/Box.swift
@@ -56,6 +56,12 @@ public struct Box: Element {
                     view.layer.cornerRadius = self.cornerStyle.radius(for: context.bounds)
                 }
 
+                if #available(iOS 13.0, *) {
+                    if self.cornerStyle.cornerCurve != view.layer.cornerCurve {
+                        view.layer.cornerCurve = self.cornerStyle.cornerCurve
+                    }
+                }
+
                 if self.borderStyle.color?.cgColor != view.layer.borderColor {
                     view.layer.borderColor = self.borderStyle.color?.cgColor
                 }
@@ -110,8 +116,10 @@ extension Box {
 
     public enum CornerStyle {
         case square
-        case capsule
-        case rounded(radius: CGFloat, corners: Corners = .all)
+        case capsule(curve: CornerCurve = .circular)
+        case rounded(radius: CGFloat, corners: Corners = .all, curve: CornerCurve = .circular)
+
+        public static let capsule: Self = .capsule()
 
         public struct Corners: OptionSet {
             public let rawValue: UInt8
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -110,8 +116,10 @@ extension Box {
 
     public enum CornerStyle {
         case square
-        case capsule
-        case rounded(radius: CGFloat, corners: Corners = .all)
+        case capsule(curve: CornerCurve = .circular)
+        case rounded(radius: CGFloat, corners: Corners = .all, curve: CornerCurve = .circular)
```
kyleve, line 120: The defaults here keep the old standard, however `.continuous` definitely looks more "iOS-y"; thoughts on swapping the defaults? Would affect POS snapshots.

BlueprintUICommonControls/Sources/Box.swift
```diff
@@ -173,6 +181,20 @@ extension Box {
         }
     }
 
+    public enum CornerCurve: Equatable {
+
+        case circular
+        case continuous
+
+        @available(iOS 13.0, *)
+        public var toLayerCornerCurve: CALayerCornerCurve {
```
kyleve, line 190: This can be internal

BlueprintUICommonControls/Sources/Box.swift
n8chur: It's unfortunate that this would be a breaking change.. maybe we should add a:
```swift
@deprecated(...)
static var capsule: CornerStyle {
    .capsule()
}
```
to assist with migration?

n8chur (approved):

BlueprintUICommonControls/Sources/Box.swift
n8chur: Ah, missed that.

Should we deprecate the non `()` version or something?

I got some push back on adding conveniences like this in the past. I _think_ it was @watt, so curious he has thoughts.

kyleve: To me the deprecation doesn't add a lot of value; just more noise unless we explicitly migrate everyone every time we add cases to enums; so I'd keep both.

Another option I was considering was adding `cornerCurve` as a top level property on `Box`, so it doesn't need to be in the enum. Thoughts?

watt: > I got some push back on adding conveniences like this in the past. I _think_ it was @watt, so curious he has thoughts. I think I do like having these whenever there are ideal defaults in the parameter list.

Could have been me. I dislike the ambiguity of having both `capsule` and `capsule()` options, especially undocumented. The parens let you know there's an associated value — you can't call a function `foo(bar: Int = 0)` as `foo` either.  I won't block on it though.

> The defaults here keep the old standard, however .continuous definitely looks more "iOS-y"; thoughts on swapping the defaults? Would affect POS snapshots.

I think we should keep the circular default. This is particularly important for `capsule`, which I think is primarily used to make circles.

> Another option I was considering was adding cornerCurve as a top level property on Box, so it doesn't need to be in the enum. Thoughts?

I'd be OK with this. We have a [flattened representation in Modals](https://github.com/squareup/market/blob/main/ios/Modals/Sources/ModalRoundedCorners.swift).

kyleve: I'm going to move this up to the `Box` root then, I think it's nicer to not keep making these enums longer and longer; and solves the source-breaking `()` addition

kylebshr (approved):

kyleve:
Ugh I just realized I also need to set the `cornerCurve` on `view.contentView` here too because it provides the clipping. Will get that PR up tomorrow...

PR #305, opened 2021-12-15
Add support for adjusting text sizing when it does not fit in the given layout rectangle

kyleve:
Allows adjusting the type size when the given type doesn't fit in layout rect. This is done by adding the `UILabel` APIs to `Label` and `AttributedLabel` for folks to enable.

Todo:

- [x] Snapshot Tests


kylebshr:
Does this break measuring in any way? How does it play with attributed text, since we're explicitly setting a font?

kyleve:
> Does this break measuring in any way? How does it play with attributed text, since we're explicitly setting a font?

Based on testing, it seems to "just work". I was expecting this to not work at all, since we're setting the attributed text on UILabel (and all the attributed contained within), but it does the right thing, somehow

watt:
> > Does this break measuring in any way? How does it play with attributed text, since we're explicitly setting a font?
> 
> Based on testing, it seems to "just work". I was expecting this to not work at all, since we're setting the attributed text on UILabel (and all the attributes contained within), but it does the right thing, somehow

Can you elaborate? Do labels measure as their normal size, or return an adjusted size (I hope it's the former)?

kyleve:
> Can you elaborate? Do labels measure as their normal size, or return an adjusted size (I hope it's the former)?

Former! Adding some tests so this can be verified tho.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -186,6 +258,10 @@ extension AttributedLabel {
             numberOfLines = model.numberOfLines
             textRectOffset = model.textRectOffset
 
+            if !isMeasuring {
```
kyleve, line 261: In case this behavior were to change, wrapping this in the measurement check.

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -144,50 +167,41 @@ class AttributedLabelTests: XCTestCase {
...
+                expectedSize,
+                label.content.measure(in: SizeConstraint(size)),
+                file: file,
+                line: line
+            )
 
-        // Strings are measured slightly differently on iOS 10
+            label.textFitting = .adjusts(allowsTightening: true, minimumScale: 0.5)
 
-        if #available(iOS 11, *) {
```
kyleve, line 155: Removed this bit.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -15,6 +15,9 @@ public struct AttributedLabel: Element, Hashable {
     /// Determines if the label should be included when navigating the UI via accessibility.
     public var isAccessibilityElement = true
 
+    /// Controls how to adjust the size of the text when it does not fit in the given layout.
+    public var textFitting: TextFittingAdjustment = .noAdjustment
```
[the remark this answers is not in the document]
kyleve: Yeah, good point. I had kinda liked being able to only pass through one property when going from marketlabel -> attributedlabel or label -> attributedlabel, but I guess the divergence isn't worth it.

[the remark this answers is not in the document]
kyleve: Moved to UILabel-style APIs!

kylebshr (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -38,6 +38,42 @@ public struct AttributedLabel: Element, Hashable {
...
+
+    /// The minimum scale factor for the label’s text.
+    ///
+    /// If the adjustsFontSizeToFitWidth is true, use this property to specify the
+    /// smallest multiplier for the current font size that yields an acceptable
+    /// font size for the label’s text.
+    ///
+    /// If you specify a value of 0 for this property, the label doesn't scale the text down.
+    /// The default value of this property is 0.
+    public var minimumScaleFactor: CGFloat = 0
```
n8chur, line 63: Nit: Should this be nullable instead of having a separate `adjustsFontSizeToFitWidth` property?

Seems fine to just match the UIKit APIs as you have, but I like having that connected state wrapped up in a single property since they're dependent on each other.

n8chur (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: I figured if we want the same properties, we may as well match them exactly, especially since there's no functional difference between someone assigning 0 vs nil.

PR #309, opened 2022-01-06
Add support for showing the system UIMenuController via the EditingMenu element

kyleve:
Adds the `EditingMenu` element, which allows showing the system's editing menu (`UIMenuController`).

Simulator Screen Shot - iPhone 13 Pro Max - 2022-01-06 at 14 51 05

### DIFF
--- a/BlueprintUICommonControls/Sources/EditingMenu.swift
+++ b/BlueprintUICommonControls/Sources/EditingMenu.swift
@@ -1,0 +1,669 @@
+//
+//  EditingMenu.swift
+//  BlueprintUICommonControls
+//
+//  Created by Kyle Van Essen on 1/5/22.
+//
+
+import BlueprintUI
+import Foundation
+import UIKit
+
+
+///
+/// Allows showing the system's `UIMenuController` editing menu.
+///
+/// You can show the menu upon tap or long press:
+/// ```
+/// myElement.editingMenu(show: .onLongPress) {
+///     EditingMenuItem.copying("A String")
+///
+///     EditingMenuItem(.select) {
+///         print("Selected!")
+///     }
+/// }
+/// ```
+///
+/// You can also show the menu as a result of another element's `onTap` closure,
+/// using the trigger-based API:
+/// ```
+/// EditingMenu { menu in
+///     MyRow(text: "Hello, World") {
+///         menu.show()
+///     }
+/// } items: {
+///     EditingMenuItem.copying("A String")
+/// }
+/// ```
+public struct EditingMenu: Element {
+
+    /// The wrapped element to display.
+    public var wrapped: Element
+
+    /// The editing items to show in the editing menu.
+    public var items: [EditingMenuItem]
+
+    let presentationMode: PresentationMode
+
+    /// Creates a new editing menu, wrapping the provided element, and displaying the provided items.
+    public init(
+        show gesture: Gesture,
+        wrapping: Element,
+        @Builder<EditingMenuItem> with items: () -> [EditingMenuItem]
+    ) {
+        let items = items()
+
+        precondition(items.isEmpty == false, "Must provide at least one `EditingMenuItem` to the `EditingMenu`.")
+
+        presentationMode = .gesture(gesture)
+
+        wrapped = wrapping
+        self.items = items
+    }
+
+    /// Creates a new editing menu, wrapping the provided element, and displaying the provided items.
+    public init(
+        wrapping: (MenuTrigger) -> Element,
+        @Builder<EditingMenuItem> items: () -> [EditingMenuItem]
+    ) {
+        let trigger = MenuTrigger()
+
+        let items = items()
+
+        precondition(items.isEmpty == false, "Must provide at least one `EditingMenuItem` to the `EditingMenu`.")
+
+        presentationMode = .trigger(trigger)
+
+        wrapped = wrapping(trigger)
+
+        self.items = items
+    }
+
+    // MARK: Element
+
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/EditingMenu.swift
```diff
@@ -0,0 +1,309 @@
...
+}
+
+
+extension EditingMenuItem {
+
+    /// All of the different menu item kinds supported by an editing menu.
+    ///
+    /// ### Note
+    /// Matches methods from `UIResponderStandardEditActions`.
+    public enum Kind {
```
kylebshr, line 86: Should this be a struct with static methods so you can add custom ones?

BlueprintUICommonControls/Sources/EditingMenu.swift
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        // MARK: UIResponderStandardEditActions
+
+        @objc override func cut(_ any: Any?) {
+            perform(kind: .cut)
+        }
+
+        @objc override func copy(_ any: Any?) {
+            perform(kind: .copy)
```
kylebshr, line 220: Do you think we should offer any default implementations? If not, I think the demo app should at illustrate actually copying, and "pasting" to the console

BlueprintUICommonControls/Sources/EditingMenu.swift
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        @objc private func handleLongPress() {
+
+            guard longPress.state == .began else { return }
+
+            _ = showMenu()
+        }
+
+        private func showMenu() -> Bool {
```
kylebshr, line 294: If this bool isn't used anywhere, does it make sense to keep it? (_Should_ we be using it somehow?)

kylebshr:
Neat, some small questions though

BlueprintUICommonControls/Sources/EditingMenu.swift
kyleve: Ah good idea! I can do that.

BlueprintUICommonControls/Sources/EditingMenu.swift
kyleve: True – I was using this for an accessibility integration, but removed that for the time being since I need to figure out the correct way to do it (rotors).

BlueprintUICommonControls/Sources/EditingMenu.swift
kyleve: Soooo the way that the `UIMenuController` API works is really miserable: The actions you can perform are based on the menu traversing the responder chain, and seeing which things in the responder chain handle the provided selector via `canPerformAction` (like `cut:`, `paste:`, etc); and then calls that selector on the responder. Because of this, custom actions would need to pass in a custom object with a custom selector that can get forward by the backing view here – which was a giant pain in the ass, so I decided to just not.

kyleve: Figured out a reasonable-ish way to support this, going to add.

kylebshr (approved):

BlueprintUICommonControls/Sources/EditingMenu.swift
```diff
@@ -0,0 +1,496 @@
...
+            return #selector(actions.toggleItalics)
+        case .toggleUnderline:
+            return #selector(actions.toggleUnderline)
+        case .increaseSize:
+            return #selector(actions.increaseSize)
+        case .decreaseSize:
+            return #selector(actions.decreaseSize)
+
+        case .printContent:
+            guard #available(iOS 15.0, *) else { fatalError() }
```
n8chur, line 302: Nit: Do we need these available checks if the case itself is `@available` checked?

BlueprintUICommonControls/Sources/EditingMenu.swift
```diff
@@ -0,0 +1,496 @@
...
+        }
+    }
+}
+
+
+extension EditingMenu {
+
+    fileprivate final class View: UIView {
+
+        var items: [EditingMenuItem.Resolved] = []
```
n8chur, line 322: Do we need to update the menu controller on `didSet`?:

```swift
var items: [EditingMenuItem.Resolved] = [] {
    didSet {
        UIMenuController.shared.update()
    }
}
```

(if so we probably want to pull `UIMenuController.shared` into a property)

n8chur (approved):
Nice!

One thing I was wondering though is if there might be a good use case for presenting on tap instead of hold. I feel like I've seen that before, but can't remember where. Might me a nice option to add in the future if we discover a use case.

BlueprintUICommonControls/Sources/EditingMenu.swift
kyleve: We do, weirdly enough

BlueprintUICommonControls/Sources/EditingMenu.swift
kyleve: Ah true – I figured it wasn't common since the menu is only visible for a short period of time, but I can add it in to be belt n' suspenders

kyleve:
> One thing I was wondering though is if there might be a good use case for presenting on tap instead of hold. I feel like I've seen that before, but can't remember where. Might me a nice option to add in the future if we discover a use case.

1Password does this, for example. I can take a look at that

kyleve:
I added an API to allow showing the menu via a Trigger as well; so you can call it, eg in a MarketRow onTap callback.

PR #311, opened 2022-01-08
Fix an issue with links and multi-line labels

kylebshr:
If `NSTextContainer` has a line break mode that isn't "wrapping" then it lays out the text on a single line regardless of the `maximumNumberOfLines`. If we have > 1 line of text, we don't set the line break mode, and leave the default of word-wrapping so the link detection still works. If the text is actually truncated, you can't touch the truncated text, so it doesn't matter if the text container doesn't match the label in this case.


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -192,10 +192,15 @@ extension AttributedLabel {
             let textContainer = NSTextContainer()
 
             textContainer.lineFragmentPadding = 0
-            textContainer.lineBreakMode = lineBreakMode
             textContainer.maximumNumberOfLines = numberOfLines
             textContainer.size = textRect(forBounds: bounds, limitedToNumberOfLines: numberOfLines).size
 
+            /// This can only be set if we have one line because NSTextContainer assumes the number of lines is 1 if
+            /// anything other than wrapping by word or character. This should still work correctly usually.
+            if numberOfLines == 1 {
```
kyleve, line 200: Any way to add a test for this?

kyleve (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -192,10 +192,15 @@ extension AttributedLabel {
             let textContainer = NSTextContainer()
 
             textContainer.lineFragmentPadding = 0
-            textContainer.lineBreakMode = lineBreakMode
             textContainer.maximumNumberOfLines = numberOfLines
             textContainer.size = textRect(forBounds: bounds, limitedToNumberOfLines: numberOfLines).size
 
+            /// This can only be set if we have one line because NSTextContainer assumes the number of lines is 1 if
+            /// anything other than wrapping by word or character. This should still work correctly usually.
```
watt, line 199: `Label` does this:
```swift
    private var effectiveLineBreakMode: NSLineBreakMode {
        // These line break modes don't work when numberOfLines is 1, and they break line height adjustments.
        // Normalize them to clipping mode instead.
        if numberOfLines == 1 && (lineBreakMode == .byCharWrapping || lineBreakMode == .byWordWrapping) {
            return .byClipping
        }
        return lineBreakMode
    }
```
Can we do that here? Maybe extract it to an extension on `NSLineBreakMode`?

kylebshr: - [ ] Extract `effectiveLineBreakMode`

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: - [ ] Try to test

n8chur (approved):
LGTM once comments are addressed!

PR #312, opened 2022-01-11
Add LayoutPhase API to LayoutWriter

kyleve:
Fixes an issue where the `Decorate` element wasn't stretching its base/root element to the full laid out size. To do this, I introduced access to the current `LayoutPhase` to the `LayoutWriter`, which can be checked to vary behavior if needed between measurement and layout passes.


BlueprintUI/Sources/Layout/LayoutWriter.swift
```diff
@@ -193,6 +200,51 @@ extension LayoutWriter {
...
+        /// Returns the provided value based on if a measurement or layout is occurring.
+        public func ifMeasuring<Value>(
+            _ ifMeasuring: @autoclosure () -> Value,
+            ifLayout: @autoclosure () -> Value
+        ) -> Value {
+            switch self {
+            case .measurement: return ifMeasuring()
+            case .layout: return ifLayout()
+            }
+        }
```
n8chur, line 234: I'm fine with adding these if we're finding them useful, but IMO it's easier and more clear to just use a switch where you need it.

E.g. declare an in-scope var where you need it
```swift
var size: CGSize {
    switch context.phase {
    case .measuring:
        return contentSize
    case .layout:
        return context.size.maximum
    }
}
```

use a closure that's executed immediately
```swift
size: {
    switch context.phase {
    case .measuring:
        return contentSize
    case .layout:
        return context.size.maximum
    }
}()
```

or provide a `var isMeasuring: Bool` convenience:
```swift
size: context.phase.isMeasuring ? contentSize : context.size.maximum
```

n8chur (approved):
Changes seem fine to me, but would like for someone else to weigh in!

BlueprintUI/Sources/Layout/LayoutWriter.swift
kyleve: I updated this to be a bit more useful – the layout callback now takes in the size, so consumers don't need to "implicitly" convert the constraint to a size.

PR #314, opened 2022-01-13
Fix AttributedLabel's UIAccessibilityTraits

kyleve:


BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -5,6 +5,63 @@ import XCTest
...
+        let updatesFrequently = AttributedLabel(attributedText: string) { label in
+            label.accessibilityTraits = [.updatesFrequently]
+        }
+
+        let view = BlueprintView(frame: CGRect(x: 0, y: 0, width: 100, height: 100))
+
+        nilTraits.accessBackingView(in: view) { label in
+            let label = label as! UILabel
+
+            XCTAssertEqual(label.accessibilityTraits, [])
```
kyleve, line 37: Hmm actually I'm surprised this is empty, vs being `.staticText`.

kyleve: Any reason I _shouldn't_ make this be `.staticText`?

n8chur: Alternatively make `[.staticText]` the default traits on the model instead of `nil`.

kyleve: I'm going to go with the former – include static text if the traits are nil. If the traits are `[]`, that means they were explicitly set to empty.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -168,13 +173,16 @@ extension AttributedLabel {
         }
 
         private func updateAccessibilityTraits(_ model: AttributedLabel) {
+
```
n8chur, line 176: Nit: Could we just move the `.staticText` setting up here?

I think we'd need to re-add `.staticText` for an update from `.updatesFrequently` back to an empty set for example.

Maybe something like:
```swift
var traits = UIAccessibilityTraits(with: [.staticText])

if let modelTraits = model.accessibilityTraits {
    traits = traits.union(.init(with: modelTraits))
    // UILabel has the `.staticText` trait by default. If we explicitly set `.updatesFrequently` this should be removed.
    if traits.contains(.updatesFrequently) {
        traits.subtract(.staticText)
    }
}

accessibilityTraits = traits
```

n8chur (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Will do

kyleve: This means that `[]` won't _remove_ `.staticText`, which I guess is fine?

kylebshr (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: Hmm, might be better to respect `[]`? Not strong opinion

kyleve: Sounds good, updating!

kyleve: Actually, hm – it seems like for labels, you'll always want either `.staticText`, or `.updatesFrequently` – if you want to disable the element, you can set `.isAccessibilityElement = false`. So maybe always including it does make sense?

kyleve: Looping in @RoyalPineapple

kyleve: Good summary here: https://mobilea11y.com/blog/traits/

I think we can be a lot smarter in the label itself about this and not require these traits at all. In the mean time I'll get this working how it was before

kyleve: In the mean time, West pointed out that we should just fix the original bug. And not change the behaviour – so doing that.

n8chur: To clarify I do think we should follow up with improving this with the help of @RoyalPineapple! I just want to get a release out with existing behavior if possible and would like to give the better fix enough time.

PR #316, opened 2022-01-14
Fix link detection with certain line break modes

kylebshr:
If `NSTextContainer` has a line break mode that isn't "wrapping" then it lays out the text on a single line regardless of the `maximumNumberOfLines`, so we need to normalize the line break mode to `byWordWrapping` on the text container if there's > 1 line.

This results in link detection not lining up with the rendered text for the `byTruncatingHead` and `byTruncatingMiddle` line break modes in multi-line labels, since `UILabel` will render the last line of text with that truncation, but the text container can't use those modes. (I even explored using `CoreText` - it also doesn't wrap text if rendering a truncated mode, this is `UILabel` specific behavior.)

The solution is to normalize the line break mode of the rendered text to one that makes sense for the number of lines:

- If the number of lines is not 1 and the line break mode is truncating head or middle, normalize to truncating tail

Additionally, we currently normalize the line break mode in `Label` since word and char wrapping break line height rendering for single line labels. I've decided to move this into `AttributedLabel`, to consolidate the logic and because it doesn't affect the rendering of the text (other than fixing the line height issue). So we also normalize:

- If the number of lines is 1 and the line break mode is word or char wrapping, normalize to clipping

Here are examples of the broken behaviors - the red rect is the rect of text as rendered by the text container, and the yellow rects with cyan borders are where it detects links:

| | Broken | Fixed |
|-|-|-|
| Not normalizing text container to a wrapping mode | test_textContainerRects_byTruncatingTail_2-lines_lineheight-None_width-320 0_iOS-14 | test_textContainerRects_byTruncatingTail_2-lines_lineheight-None_width-320 0_iOS-14 2 |
| Not normalizing head truncation | test_textContainerRects_byTruncatingHead_2_lines_lineheight_None_280_iOS-14 | test_textContainerRects_byTruncatingMiddle_2_lines_lineheight_None_280_iOS-14 2 |
| Not normalizing middle truncation | test_textContainerRects_byTruncatingMiddle_2_lines_lineheight_None_280_iOS-14 | test_textContainerRects_byTruncatingMiddle_2_lines_lineheight_None_280_iOS-14 2 |


n8chur (approved):
Nice work finding a reasonable solution!

Love the new snapshot tests.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue where `AttributedLabel` might not detect link taps in multi-line labels.
```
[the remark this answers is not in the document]
kylebshr: Oh definitely, I forgot to update this - thanks

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -156,6 +170,7 @@ extension AttributedLabel {
                 .attributedText
                 .applyingDefaultFont()
                 .replacingLinkAttributes()
+                .normalizingLineBreakMode(for: model.numberOfLines)
```
kyleve, line 173: This is minor, but I wonder how much of a perf hit we get here with the 4 copy, edit, returns from this builder style.

kyleve (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
kylebshr: Yeah, I think I'll combine them. Probably not the best thing to do for every label on every render.

PR #322, opened 2022-01-20
added helpers for accessibility modifications to elements

RoyalPineapple:
These four accessibility helper methods cover 85% of the use cases for accessibility configuration as seen in the wild.

> 
image

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -121,6 +121,77 @@ extension Element {
         )
     }
 
+    /// Wraps the receiver in an accessibility element with the provided values.
+    ///
+    /// - Important: ⚠️: This will remove any previously resolved accessibility hint, identifier, or frame size of the element and all of its children. ⚠️
+    public func accessibilityElement(
+        label: String,
+        value: String,
+        traits: Set<AccessibilityElement.Trait>
+    ) -> AccessibilityElement {
+        AccessibilityElement(
+            label: label,
+            value: value,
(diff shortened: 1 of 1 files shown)


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -121,6 +121,77 @@ extension Element {
         )
     }
 
+    /// Wraps the receiver in an accessibility element with the provided values.
```
kyleve, line 124: Could we provide a better name for this method? Eg, the only difference here is that the method passes through `nil`, which we could also do through extending the base method with optional parameters. Eg, could we instead give this a better name for the use cases where these parameters are always nil? Like `accessibilityButton` or something.

n8chur:
Are these just being added so that the autocomplete suggestions look better? This doesn't make any effective API changes beyond that, right?

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -121,6 +121,77 @@ extension Element {
...
+    ///
+    /// - Important: ⚠️: This will remove any previously resolved accessibility hint, identifier, or frame size of the element and all of its children. ⚠️
+    public func accessibilityElement(
+        label: String,
+        value: String,
+        traits: Set<AccessibilityElement.Trait>
+    ) -> AccessibilityElement {
```
n8chur, line 131: I'm fairly certain this (and likely a few others) will end up being ambiguous to the compiler which to chose. Could you add a test or something that references each to make sure the compiler doesn't get confused?

PR #325, opened 2022-02-01
Add .fittingConstraint to ScrollView

kyleve:
This allows, eg, contained `Column`s to include flexible elements which will expand to fill the available space when there's underflow, and have a minimum size on overflow. This also means you could now implement `centersUnderflow` purely via composition of a contained column, instead of needing a special mode on the scroll view:

```
Column(alignment: .fill) {

   row.stackChild(.fixed)
   row.stackChild(.fixed)
   row.stackChild(.fixed)

  Spacer(height: 1.0).stackChild(.grows) // I forget the exact syntax for this.

   row.stackChild(.fixed)
}
.scrollable {
   $0.size = .fittingConstraint
}
```

Etc...

### DIFF
--- a/BlueprintUICommonControls/Tests/Sources/ScrollViewTests.swift
+++ b/BlueprintUICommonControls/Tests/Sources/ScrollViewTests.swift
@@ -121,6 +121,7 @@ class ScrollViewTests: XCTestCase {
         test(contentSize: .fittingContent)
         test(contentSize: .fittingHeight)
         test(contentSize: .fittingWidth)
+        test(contentSize: .fittingConstraint)
     }
 
     private struct UnderflowElement: ProxyElement {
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Tests/Sources/ScrollViewTests.swift
```diff
@@ -121,6 +121,7 @@ class ScrollViewTests: XCTestCase {
         test(contentSize: .fittingContent)
         test(contentSize: .fittingHeight)
         test(contentSize: .fittingWidth)
+        test(contentSize: .fittingConstraint)
```
kyleve, line 124: TODO: Commit snapshots.

watt:
> This allows, eg, contained Columns to include flexible elements which will expand to fill the available space when there's underflow

Hm, I'm not sure I see how this will achieve that. It seems like this will just prevent things from growing large enough that scrolling kicks in?

kyleve:
Going to iterate on this one a bit more – I think an option that fills/stretches the scroll view content to a min size (the height of the scroll view) is a better approach.

kyleve:
Closing this one and opening another one which introduces `.fill` for ensuring the contained element at least is given the allowed space

PR #326, opened 2022-02-10
Add modify conditional

kyleve:
This lets you edit a prop inline without needing to call `map` and make a local copy


n8chur:
Could it be worth introducing a `Modifiable` protocol to make this easier to apply to other types?

```swift
public protocol Modifiable {

    func modify(_ modification: (inout Self) -> Void) -> Self
}

extension Modifiable {

    public func modify(_ modification: (inout Self) -> Void) -> Self {
        var modified = self
        modification(&modified)
        return modified
    }
}

extension Element: Modifiable {}
```

n8chur:
Also wonder if `configure` might be a more appropriate name since we've been using that naming convention in other places.

kyleve:
I was debating that (the larger protocol), but given we already have all these element conditionals, I didn't wanna get into a larger debate about it. Thoughts?

n8chur:
I think we discussed this previously and I forget the take-away, but I feel like it'd be nice to use this pattern instead of explicit `configure` closures for all of our components.

Would definitely like input from others though.

watt:
Kind of torn on this. I see the utility, but I also see that it's very clever — maybe too clever. It's not idiomatic Swift, nor a pattern used by SwiftUI. I think the bar for introducing new patterns needs to be pretty high.

n8chur:
@kyleve did you have a concrete use-case in mind we could review?

kylebshr:
Dang, I just remembered seeing this PR in my email and came to finally check it out since I wanted it today. I think this is a good addition, especially as we move to result builders. One example is this snippet - we used to be able to modify stacks in the configure block:

```swift
var content: Element {
    Row { row in
        row.add(child: ...)
        row.add(child: ...)
        if style.layoutDirection == .rightToLeft {
            row.children.reverse()
        }
    }
}
```

Now with result builders, we lose the ability to do this as easily. In fact, we lost the RTL support in banner recently when [migrating to builders](https://github.com/squareup/market/commit/d20013f33f53e2631813355f6d01b3b509d08b22#diff-77cd8640c6b44917799ab043ade8211d28b95bbd34a00a77eee14eeb8898df6fL92).

`modify` lets you use builders and still customize the element without having to make a copy.

```swift
var content: Element {
    Row {
        child1
        child2
    }
    .modify { row in
        if style.layoutDirection == .rightToLeft {
            row.children.reverse()
        }
    }
}
```

vs.
```swift
var content: Element {
    var row = Row {
        child1
        child2
    }
    
    if style.layoutDirection == .rightToLeft {
        row.children.reverse()
    }
    
    return row
}
```

kyleve:
+1, that was another case I was running into, now that you mention it

watt:
> Now with result builders, we lose the ability to do this as easily. In fact, we lost the RTL support in banner recently when [migrating to builders](https://github.com/squareup/market/commit/d20013f33f53e2631813355f6d01b3b509d08b22#diff-77cd8640c6b44917799ab043ade8211d28b95bbd34a00a77eee14eeb8898df6fL92).
> 
> `modify` lets you use builders and still customize the element without having to make a copy.

I think if you want to make the argument that we've lost functionality with result builders, that's fair, but I think we could look for idiomatic & declarative ways to fill those specific gaps, before tacking on the old `configure`-style imperative escape hatch to such a broad scope.

For something like RTL layout, I can think of a couple ideas offhand:
- the row could do this automatically
- the row could take a parameter
- the row could have a `.reversed()` modifier

And then we don't have to open Pandora's box of imperative API surfaces.

kyleve:
> For something like RTL layout, I can think of a couple ideas offhand:

This seems like one example, but there's many others; any time you want to config an element.

> And then we don't have to open Pandora's box of imperative API surfaces.

Is it really imperative? It's basically just config on the thing. I don't think it's fair to say that anything but builders are "imperative".

nononoah:
For whatever it is worth, I don't see `modify` as particularly non-Swifty, but rather as a special kind of `map` on non-container types.

I think it's OK to bridge these gaps where doing so preserves a more legible structure.

n8chur (approved):

PR #329, opened 2022-02-16
Fix voice over rotor crash

kylebshr:
Fixes an issue where the app would crash if Voice Over attempted to navigate a screen with labels without links (in some cases - couldn't always reproduce). Also doesn't return the rotor at all if there are no links.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -140,20 +140,25 @@ extension AttributedLabel {
         override var accessibilityCustomRotors: [UIAccessibilityCustomRotor]? {
             set { fatalError() }
             get {
-                [
-                    UIAccessibilityCustomRotor(systemType: .link) { [weak self] predicate in
-                        guard let self = self else {
-                            return nil
-                        }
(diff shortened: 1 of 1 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -140,20 +140,25 @@ extension AttributedLabel {
...
-
-                        self.accessibilityLinkIndex += predicate.searchDirection == .next ? 1 : -1
-                        self.accessibilityLinkIndex = min(self.accessibilityLinks.count - 1, self.accessibilityLinkIndex)
-                        self.accessibilityLinkIndex = max(0, self.accessibilityLinkIndex)
-
-                        let link = self.accessibilityLinks[self.accessibilityLinkIndex]
-                        return UIAccessibilityCustomRotorItemResult(targetElement: link, targetRange: nil)
-                    },
-                ]
+                accessibilityLinks.isEmpty
```
kyleve, line 143: Not sure if it's worth it, but might be nice to add a test for this since it deals with finicky index math

kyleve (approved):

PR #331, opened 2022-02-22
Text shadows

watt:
shadows

For `Label` and `AttributedLabel`.

Supports a single shadow applied to the whole element, using `CALayer` properties. I also tried the `shadow` attributed string key, but I couldn't find a way to prevent that from being clipped by the bounds of the view.


.github/workflows/tests.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
     - name: Package snapshot changes
       if: ${{ failure() }}
       run: |
-        git ls-files -m BlueprintUICommonControls/Tests/Sources/ReferenceImages | xargs tar -cvf snapshot_changes_${{ matrix.sdk }}.tar
+        git ls-files -mo BlueprintUICommonControls/Tests/Sources/ReferenceImages | xargs tar -cvf snapshot_changes_${{ matrix.sdk }}.tar
```
watt, line 52: The [`-o` flag](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt--o) includes untracked files. I changed this in order to grab the new snapshot images off of CI instead of running all these sims locally.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -184,9 +192,26 @@ extension AttributedLabel {
...
+                    accessibilityLinks = accessibilityLinks(for: links, in: model.attributedText)
+                }
+
+                if let shadow = model.shadow {
+                    layer.shadowRadius = shadow.radius
+                    layer.shadowOpacity = Float(shadow.opacity)
+                    layer.shadowOffset = CGSize(width: shadow.offset.horizontal, height: shadow.offset.vertical)
+                    layer.shadowColor = shadow.color.cgColor
+
+                    // enable rasterization on this layer, since text is too complex for a shadow path
```
kyleve, line 207: I take it this is faster at rendering than turning the text into a path and setting that as the shadow path? Or is that what this comment means?

kyleve (approved):

.github/workflows/tests.yaml
kyleve: TIL!

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Yep, although I did not test it myself, all the reading material I found on layer shadows suggested offscreen compositing for cases where a path is impractical ([this, for example](https://swiftlemma.com/2017/08/20/better-shadow-performance-on-views/)). I can rewrite this comment to clarify.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -184,9 +192,26 @@ extension AttributedLabel {
...
+                    layer.shouldRasterize = true
+                    layer.rasterizationScale = layer.contentsScale
```
kylebshr, line 209: Does this need to be updated if the contents scale changes? Probably very much an edge case 🤔

kylebshr (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: When I tested this the `rasterizationScale` was already correct before being set, so tbh I'm not sure what the circumstances are that you'd need to do this. `BlueprintView` re-renders when it changes windows, so we should pick up scale changes that way. (If there's another way the scale can change, I dunno.)

PR #332, opened 2022-03-09
Fix doc script for Xcode 13

watt:
Seems like this broke a while ago because `instruments` went away. New version seems a bit more fragile, as the sim names don't quite line up with what `xcodebuild` thinks, but this works for me locally. 🤷


kyleve (approved):
Would simctl work here? https://nshipster.com/simctl/

watt:
> Would simctl work here? https://nshipster.com/simctl/

I don't think so — we'd still have to do some fragile text processing or add a JSON processing step. Kind of hoping this holds together with baling wire until we can switch to the docc plugin in Swift 5.6.

PR #334, opened 2022-03-10
Added AccessibilityAction closures 

RoyalPineapple:
Added AccessibilityAction closure variables to Accessibility Element. 
This allows for a richer and more customized accessibility experience.

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -31,6 +31,10 @@ public struct AccessibilityElement: Element {
     public var accessibilityFrameSize: CGSize?
     public var wrappedElement: Element
 
+    public var accessibilityIncrement: (() -> Void)? = nil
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -31,14 +31,19 @@ public struct AccessibilityElement: Element {
     public var accessibilityFrameSize: CGSize?
     public var wrappedElement: Element
 
+    public var accessibilityIncrement: (() -> Void)? = nil
```
kyleve, line 34: Can we document what these would be used for? I imagine most folks aren't familiar

kylebshr: Sorry to be so nit picky (and I know the rest of the repo doesn't set a good precedent) but could you update the comments to have triple slashes and copy it to decrement, so it shows up as a doc comment? Same for `accessibilityActivate` (you may also need to remove the white space)

kylebshr (approved):

PR #335, opened 2022-03-22
Aligned positioning for `Decorate`

watt:
This PR is a refactor of `Decorate.Position` to support a new `aligned` position.

Instead of an enum, `Position` is a now struct wrapping a closure. Each existing case is a static func. This means that custom and built-in positions are "made of the same stuff" (to quote @bencochran), and custom cases can do anything that the built-in ones can.

This change made it easy for me to build this new option:
```swift
public static func aligned(
    to alignment: Alignment,
    horizontalGuide: ((ElementDimensions) -> CGFloat)? = nil,
    verticalGuide: ((ElementDimensions) -> CGFloat)? = nil
) -> Self
```

Which uses the same alignment types and concepts from stacks.

The diff isn't bad, but you can also view this PR by commit to see how I refactored this step-by-step.

### DIFF
--- a/BlueprintUI/Sources/Layout/Alignment.swift
+++ b/BlueprintUI/Sources/Layout/Alignment.swift
@@ -1,5 +1,42 @@
 import CoreGraphics
 
+/// An alignment in both axes.
+public struct Alignment: Equatable {
+    /// A guide marking the center of the element.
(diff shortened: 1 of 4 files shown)


kyleve (approved):
Nice! Much cleaner.

bencochran (approved):
Love some good gardening

PR #337, opened 2022-03-29
Accessibility increment, decrement  as associated values

RoyalPineapple:
Suggested by @n8chur, I think this is a good tradeoff of adding slight complexity in dealing with the enum in favor of ensuring that the trait is never applied without the callbacks being available.

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -88,6 +88,9 @@ public struct AccessibilityElement: Element {
             {
                 config[\.increment] = incrementAction
                 config[\.decrement] = decrementAction
+            } else {
+                config[\.increment] = nil
+                config[\.decrement] = nil
             }
         }
     }
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -83,13 +83,12 @@ public struct AccessibilityElement: Element {
             config[\.accessibilityFrameSize] = accessibilityFrameSize
             config[\.activate] = accessibilityActivate
 
-            guard let adjustable = traits.first(where: { $0 == .adjustable({}, {}) }),
-                  case let .adjustable(incrementAction, decrementAction) = adjustable
-            else {
-                return
+            if let adjustable = traits.first(where: { $0 == .adjustable({}, {}) }),
+               case let .adjustable(incrementAction, decrementAction) = adjustable
+            {
+                config[\.increment] = incrementAction
+                config[\.decrement] = decrementAction
             }
-            config[\.increment] = incrementAction
-            config[\.decrement] = decrementAction
         }
     }
 
@@ -138,8 +137,9 @@ public struct AccessibilityElement: Element {
 }
 
 extension AccessibilityElement.Trait: Hashable, Equatable {
-
-    public var rawValue: Int {
+    /// This conformance to `Hashable` is provided to allow traits to be included in a `Set`.
+    /// - Important: ⚠️ This implementation does not take equality of associated values on `.adjustable` into account.  ⚠️
+    private var internalValue: Int {
         switch self {
         case .button: return 0
         case .link: return 1
@@ -162,15 +162,11 @@ extension AccessibilityElement.Trait: Hashable, Equatable {
     }
 
     public static func == (lhs: AccessibilityElement.Trait, rhs: AccessibilityElement.Trait) -> Bool {
-        lhs.rawValue == rhs.rawValue
-    }
-
-    public var hashValue: Int {
-        rawValue
+        lhs.internalValue == rhs.internalValue
     }
 
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(hashValue)
+        hasher.combine(internalValue)
     }
 }
 
--- a/BlueprintUICommonControls/Sources/AccessibilityElement.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityElement.swift
@@ -3,7 +3,15 @@ import UIKit
 
 public struct AccessibilityElement: Element {
 
-    public enum Trait: Hashable {
+    public enum Trait {
+        /// Used in conjunction with UIAccessibilityTrait.adjustable, these will be called to allow accessible adjustment of a value, for example in a slider or stepper control.
+        /// See [Accessibility Increment Documentation](https://developer.apple.com/documentation/objectivec/nsobject/1615076-accessibilityincrement) for further information.
+        public typealias IncrementAction = () -> Void
+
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -131,6 +137,42 @@ public struct AccessibilityElement: Element {
     }
 }
 
+extension AccessibilityElement.Trait: Hashable, Equatable {
```
kyleve, line 140: Does "hashable" still mean something if `adjustable` now has closures? I guess this is more a hash of the type, rather then the actual content, right?

n8chur: Yeah, was thinking the same. Do we need `Hashable`/`Equatable` conformance on this?

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -131,6 +137,42 @@ public struct AccessibilityElement: Element {
...
+        }
+    }
+
+    public static func == (lhs: AccessibilityElement.Trait, rhs: AccessibilityElement.Trait) -> Bool {
+        lhs.rawValue == rhs.rawValue
+    }
+
+    public var hashValue: Int {
+        rawValue
+    }
```
n8chur, line 170: Pretty sure this is deprecated and you only need to add `hash(into:)` for `Hashable` conformance

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: I think it's b/c we throw it into a `Set` and need to check traits in said set. But I think only the type of the trait matters, so it's OK that we don't care about the closures.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -131,6 +137,42 @@ public struct AccessibilityElement: Element {
     }
 }
 
+extension AccessibilityElement.Trait: Hashable, Equatable {
+
+    public var rawValue: Int {
```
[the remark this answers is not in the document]
kyleve: +1

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -81,9 +81,15 @@ public struct AccessibilityElement: Element {
             config[\.accessibilityTraits] = accessibilityTraits
             config[\.isAccessibilityElement] = true
             config[\.accessibilityFrameSize] = accessibilityFrameSize
-            config[\.increment] = accessibilityIncrement
-            config[\.decrement] = accessibilityDecrement
             config[\.activate] = accessibilityActivate
+
+            guard let adjustable = traits.first(where: { $0 == .adjustable({}, {}) }),
```
kylebshr, line 86: nit: I would use an `if` here since we don't really want an early return, it would be easy to add more configuration at the end of this closure without realizing it won't be executed

BlueprintUICommonControls/Sources/AccessibilityElement.swift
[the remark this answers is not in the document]
n8chur: Nit: We may also want to add a doc comment describing the `Hashable`/`Equatable` conformance subtleties here.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: Yep this is exactly what's going on.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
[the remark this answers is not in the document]
RoyalPineapple: Good suggestions on both.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: Added comment explaining the conformance.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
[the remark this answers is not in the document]
RoyalPineapple: added comment and renamed var

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: removed

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: changed to `if`

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -131,6 +136,39 @@ public struct AccessibilityElement: Element {
     }
 }
 
+extension AccessibilityElement.Trait: Hashable, Equatable {
+    /// This conformance to `Hashable` is provided to allow traits to be included in a `Set`.
+    /// - Important: ⚠️ This implementation does not take equality of associated values on `.adjustable` into account.  ⚠️
```
n8chur, line 141: This conformance is public, right? We should probably also get this note in the public API for this type.

n8chur (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: It was previously hashable as well. This isn't a change in that regard. I just moved the conformance to the extension.

RoyalPineapple: It was previously hashable as well. This isn't a change in that regard. I just moved the conformance to the extension.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -81,9 +81,14 @@ public struct AccessibilityElement: Element {
...
-            config[\.increment] = accessibilityIncrement
-            config[\.decrement] = accessibilityDecrement
             config[\.activate] = accessibilityActivate
+
+            if let adjustable = traits.first(where: { $0 == .adjustable({}, {}) }),
+               case let .adjustable(incrementAction, decrementAction) = adjustable
+            {
+                config[\.increment] = incrementAction
+                config[\.decrement] = decrementAction
+            }
```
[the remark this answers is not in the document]
RoyalPineapple: probably smart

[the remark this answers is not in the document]
RoyalPineapple: probably smart

PR #339, opened 2022-04-01
Allow optionals in builders.

ldstreet:


kylebshr (approved):

kyleve (approved):

kyleve:
Mind updating the changelog as well?

kyleve:
Also seeing some test failures – should make sure we can make this change source compatible

ldstreet:
> Also seeing some test failures – should make sure we can make this change source compatible

Fixed this. I had some extra code that I removed locally but failed to push up. I think it should pass now.

n8chur (approved):

PR #340, opened 2022-04-04
Add None type as an escape hatch for all builders.

ldstreet:
This adds a None type as an escape hatch when control flows such as Switch requires an output but none is desired.

### DIFF
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -44,4 +44,23 @@ public struct Builder<Child> {
     public static func buildExpression(_ children: [Child]) -> Children {
         children
     }
+
+    /// Allow an escape hatch when the control flow requires an output (such as a `switch`).
+    public static func buildExpression(_ children: None) -> Children {
+        []
+    }
+}
+
+/// Empty type that results in no children. This is useful when an expression requires output but no children are applicable (i.e. a switch statement).
+/// Example:
+/// ```
+/// switch item {
+/// case .first:
+///    None()
+/// case .second:
+///     "2"
+/// }
+/// ```
+public struct None {
+    public init() {}
 }
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -46,7 +46,7 @@ public struct Builder<Child> {
     }
 
     /// Allow an escape hatch when the control flow requires an output (such as a `switch`).
-    public static func buildExpression(_ children: None) -> Children {
+    public static func buildExpression(_ children: Nothing) -> Children {
         []
     }
 }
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/Builder.swift
```diff
@@ -44,4 +44,23 @@ public struct Builder<Child> {
...
+/// Example:
+/// ```
+/// switch item {
+/// case .first:
+///    None()
+/// case .second:
+///     "2"
+/// }
+/// ```
+public struct None {
```
kylebshr, line 64: Curious what other folks on the team think about this name; could also see `BuildNone` or `BuildNothing` being nice, since it hints at the result builder use-case and is perhaps more clear when reading?

ldstreet: I'm open to that! Definitely not sold on the name since it is a bit vague.

kyleve: I was also thinking `Nothing` could work

watt:
With optionals supported in builders now, can you return `nil` from a switch?

ldstreet:
> With optionals supported in builders now, can you return `nil` from a switch?

Nope 🙁 I think since `nil` is just syntactic sugar for `Optional<SomeType>.none` it is ambiguous to the builder what type     nil is referring too so it gives an error.

BlueprintUI/Sources/Layout/Builder.swift
ldstreet: I'm good with `Nothing`. I wonder if the prefix `Build` would be a little less intuitive because most other declarations inside the builder do not use a verb as a prefix, they just declare the noun. Though `BuildNothing` would be less likely to conflict with other libraries?

kyleve: No strong feeling here. It feels rare enough that it's not super critical either way

ldstreet: Changed to `Nothing`

watt:
> > With optionals supported in builders now, can you return `nil` from a switch?
> 
> Nope 🙁 I think since `nil` is just syntactic sugar for `Optional<SomeType>.none` it is ambiguous to the builder what type nil is referring too so it gives an error.

Dang. Is there any prior art for handling this in SwiftUI? I'm just hesitant to invent a pattern if we can avoid it.

kylebshr:
I'm actually seeing `nil` work here 🤔 Since the result builder is strongly typed, I was surprised it's ambiguous. What test case were you seeing fail?
Screen Shot 2022-04-08 at 4 01 44 PM

kylebshr:
Oh hmm this doesn't work for `ElementBuilder`..

kylebshr:
I think maybe we should leave this out; I think it'll work correctly for pure `Builder`s, since there's not an ambiguous optional `buildExpression`, and instead of this type, you can return a strongly typed optional (e.g., `Element?.none`)

kylebshr:
After some research I discovered that we can use `@_disfavoredOverload` to fix this; I'd like to propose that approach: https://github.com/square/Blueprint/pull/344

PR #344, opened 2022-04-08
Support nil in ElementBuilder

kylebshr:
Support `nil` in all cases in `ElementBuilder` (and any other such specializations of `Builder`) by disfavoring the optional build expression function in `Builder`.

### DIFF
--- a/BlueprintUI/Sources/Layout/Builder.swift
+++ b/BlueprintUI/Sources/Layout/Builder.swift
@@ -23,6 +23,10 @@ public struct Builder<Child> {
         [child]
     }
 
+    /// This function is disfavored in case a builder wants to offer additional `buildExpression` functions to support
+    /// multiple types in their builders. `@_disfavoredOverload` allows the compiler to choose the other function when
+    /// `nil` is built, instead of it being ambiguous.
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Layout/Builder.swift
```diff
@@ -23,6 +23,10 @@ public struct Builder<Child> {
         [child]
     }
 
+    /// This function is disfavored in case a builder wants to offer additional `buildExpression` functions to support
```
kyleve, line 26: TIL!

kyleve: (Also ugh, we have like 5 copies of these result builders. I wonder if we should put these into a pod somewhere and use them in market, blueprint, listable at some point...)

kyleve (approved):
Nice find!

ldstreet:
Awesome!

watt (approved):
- [ ] changelog

kylebshr:
The changelog has an entry that captures this due to another change (Added support for optionals in builders without unwrapping via `if let`.)

PR #345, opened 2022-04-22
Add date picker control

kylebshr:


DatePicker.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
+        NativeDatePicker.describe { config in
+            config.builder = {
+                NativeDatePicker(frame: context.bounds)
+            }
+
+            config.apply { view in
+                view.apply(model: self)
+//                print(view.frame)
```
kylebshr, line 70:
```suggestion
```

DatePicker.swift
```diff
@@ -0,0 +1,112 @@
+//
+//  DatePicker.swift
```
kyleve, line 2: File's in the wrong spot

DatePicker.swift
```diff
@@ -0,0 +1,112 @@
...
+        onChange: @escaping (Date) -> Void,
+        configure: ((inout DatePicker) -> Void)? = nil
+    ) {
+        self.date = date
+        self.onChange = onChange
+        configure?(&self)
+    }
+
+    public var content: ElementContent {
+        struct Measurer {
```
kyleve, line 41: Do you need this; can you just do it inline in the measure block?

kyleve (approved):
Changelog!

DatePicker.swift
kylebshr: dang it

DatePicker.swift
kylebshr: Need a place for the static instance, so it's nicely scoped with this type.

PR #348, opened 2022-05-20
Stack Layout Optimizations

kyleve:
This PR optimizes calculating the axis measurements in `Stack` elements (`Row` and `Column`) when there are N fixed and 1 flexible elements. This generally gets us a ~50% speedup in many cases where this sort of layout is prevalent (eg, a `MarketRow` with a leading accessory, one label, and then trailing content).

**Before (running `test_rows_with_one_flexible_element`):**
`Iterations: 6, Average Time: 0.3376373251279195`

**After:**
`Iterations: 12, Average Time: 0.17800241708755493`

This is done by optimizing the axis measurement algorithm to take into account the fixed-ness and flexible-ness of content in the stack. Before this change, we would (note: some simplifying done here)...

1. Measure every child element in the full axis size of the stack to determine their size.
2. Find the fixed width of the fixed width items in the stack, use this to re-measure the flexible elements (paraphrased).
3. Re-measure the flexible elements in the full width minus the fixed width items.

Notable, step 3 **always caused a second measurement pass on the flexible elements**. This got worse the deeper the element tree downstream from that flexible element, since 2 measurements could become 4, could become 8, etc...

To resolve, I changed the axis segment calculation to...

1. Find the fixed size elements.
2. Measure them in the full size.
3. Add up the total axis size of the fixed size elements.
4. Use the full axis size minus the axis size of fixed elements to measure the flexible element(s). 

In the cases that there is only 1 flexible element (quite common), this results in one less measurement pass on the flexible element, which can dramatically improve rendering times. This is because fewer measurements overall occur. Eg, in `test_rows_with_one_flexible_element_simple`, with 10 child elements, before this change, the labels are measured a total of 80 times. After it, they are measured only 30 times.

#### `CacheTree` rounding
Moved to a follow up PR: https://github.com/square/Blueprint/pull/362 ~Due to float addition and subtraction that happens throughout layout, it was common to end up with `210.0` getting turned into `209.999997` (etc) throughout. This was causing needless cache misses. To resolve, I introduced pixel-based rounding to the cache tree so the cached constraints land on a display pixel. I'm not 100% sure this is the best approach, but it's a starting point.~

#### Further optimization
There's some further optimization we could do here, since a lot the remaining measurements are semantically identical (with varying size constraints that all result in the same outcome), but optimizing this is a lot harder. We can also do some optimization of the cross segment calculation, but I will follow up with this one.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -206,10 +206,7 @@ public final class BlueprintView: UIView {
         let measurement = element.content.measure(
             in: constraint,
             environment: makeEnvironment(),
-            cache: CacheFactory.makeCache(
-                name: "sizeThatFits:\(type(of: element))",
-                screenScale: window != nil ? layer.contentsScale : UIScreen.main.scale
-            )
+            cache: CacheFactory.makeCache(name: "sizeThatFits:\(type(of: element))")
         )
 
         sizesThatFit[constraint] = measurement
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -206,7 +206,10 @@ public final class BlueprintView: UIView {
         let measurement = element.content.measure(
             in: constraint,
             environment: makeEnvironment(),
-            cache: CacheFactory.makeCache(name: "sizeThatFits:\(type(of: element))")
+            cache: CacheFactory.makeCache(
+                name: "sizeThatFits:\(type(of: element))",
+                screenScale: window != nil ? layer.contentsScale : UIScreen.main.scale
+            )
         )
 
         sizesThatFit[constraint] = measurement
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -12,6 +12,9 @@ protocol CacheTree: AnyObject {
     /// A reference to use for logging
     var signpostRef: AnyObject { get }
 
+    /// The scale of the screen scale, used for rounding to account for loss of float precision.
+    var screenScale: CGFloat { get }
+
     /// The sizes that are contained in this cache, keyed by size constraint.
     subscript(constraint: SizeConstraint) -> CGSize? { get set }
 
@@ -22,12 +25,31 @@ protocol CacheTree: AnyObject {
 extension CacheTree {
     /// Convenience method to get a cached size, or compute and store one if it is not in the cache.
     func get(_ constraint: SizeConstraint, orStore calculation: (SizeConstraint) -> CGSize) -> CGSize {
-        if let size = self[constraint] {
+
+        /// Due to various math upstream, we can end up with measurements like 210.0 getting
+        /// turned into numbers like 209.99999997 due to loss of precision.
+        ///
+        /// Because we cache by exact values in this cache,
+        /// let's round this up to the nearest screen-scale pixel.
+
+        let roundedConstraint = constraint.roundToNearestPixel(with: screenScale)
+
+        // TODO: Why does this break??
+        // let constraint = constraint.roundToNearestPixel(with: screenScale)
+
+        if let size = self[roundedConstraint] {
+            return size
+        } else {
+            let size = calculation(constraint)
+
+            /// 1) Cache the measured size for the given constraint.
+            self[roundedConstraint] = size
+
+            /// 2) Optimization: Cache the size itself as its own constraint.
+            self[SizeConstraint(size)] = size
+
             return size
         }
-        let size = calculation(constraint)
-        self[constraint] = size
-        return size
     }
 
     /// Gets a subcache for an element with siblings.
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -28,14 +28,7 @@ extension CacheTree {
         } else {
             let size = calculation(constraint)
 
-            /// 1) Cache the measured size for the given constraint.
             self[constraint] = size
-
-            /// 2) Optimization: Cache the measured size, so if we ask for this
-            /// exact size later on, we don't re-measure the element.
-            ///
-            /// This avoids a cache miss later on when a layout lays out an item with a measured size.
-            self[SizeConstraint(size)] = size
 
             return size
         }
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -12,9 +12,6 @@ protocol CacheTree: AnyObject {
     /// A reference to use for logging
     var signpostRef: AnyObject { get }
 
-    /// The scale of the screen scale, used for rounding to account for loss of float precision.
-    var screenScale: CGFloat { get }
-
     /// The sizes that are contained in this cache, keyed by size constraint.
     subscript(constraint: SizeConstraint) -> CGSize? { get set }
 
@@ -26,29 +23,18 @@ extension CacheTree {
     /// Convenience method to get a cached size, or compute and store one if it is not in the cache.
     func get(_ constraint: SizeConstraint, orStore calculation: (SizeConstraint) -> CGSize) -> CGSize {
 
-        /// Due to various math upstream, we can end up with measurements like 210.0 getting
-        /// turned into numbers like 209.99999997 due to loss of precision.
-        ///
-        /// Because we cache by exact values in this cache,
-        /// let's round this to the nearest screen-scale pixel to avoid a cache miss.
-
-        let roundedConstraint = constraint.roundToNearestPixel(with: screenScale)
-
-        // TODO: Why does this break??
-        // let constraint = constraint.roundToNearestPixel(with: screenScale)
-
-        if let size = self[roundedConstraint] {
+        if let size = self[constraint] {
             return size
         } else {
             let size = calculation(constraint)
 
             /// 1) Cache the measured size for the given constraint.
-            self[roundedConstraint] = size
+            self[constraint] = size
 
-            /// 2) Optimization: Cache the size itself as its own constraint.
+            /// 2) Optimization: Cache the measured size, so if we ask for this
+            /// exact size later on, we don't re-measure the element.
             ///
-            /// This avoids a cache miss later on when a layout
-            /// lays out an item with this same size.
+            /// This avoids a cache miss later on when a layout lays out an item with a measured size.
             self[SizeConstraint(size)] = size
 
             return size
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -30,7 +30,7 @@ extension CacheTree {
         /// turned into numbers like 209.99999997 due to loss of precision.
         ///
         /// Because we cache by exact values in this cache,
-        /// let's round this up to the nearest screen-scale pixel.
+        /// let's round this to the nearest screen-scale pixel to avoid a cache miss.
 
         let roundedConstraint = constraint.roundToNearestPixel(with: screenScale)
 
@@ -46,6 +46,9 @@ extension CacheTree {
             self[roundedConstraint] = size
 
             /// 2) Optimization: Cache the size itself as its own constraint.
+            ///
+            /// This avoids a cache miss later on when a layout
+            /// lays out an item with this same size.
             self[SizeConstraint(size)] = size
 
             return size
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -270,11 +270,6 @@ public struct StackLayout: Layout {
 
 }
 
-/// Temporary. Allows testing stack layout optimizations to compare performance.
-enum BlueprintStackUpdates {
-
-    static var isOptimizedLayoutEnabled: Bool = true
-}
 
 extension StackLayout {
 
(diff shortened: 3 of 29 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+"""
```
kyleve, line 1: Delete?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -206,7 +206,10 @@ public final class BlueprintView: UIView {
         let measurement = element.content.measure(
             in: constraint,
             environment: makeEnvironment(),
-            cache: CacheFactory.makeCache(name: "sizeThatFits:\(type(of: element))")
+            cache: CacheFactory.makeCache(
+                name: "sizeThatFits:\(type(of: element))",
+                screenScale: window != nil ? layer.contentsScale : UIScreen.main.scale
```
kyleve, line 211: TODO: Is this the right fallback?

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -22,12 +25,27 @@ protocol CacheTree: AnyObject {
 extension CacheTree {
     /// Convenience method to get a cached size, or compute and store one if it is not in the cache.
     func get(_ constraint: SizeConstraint, orStore calculation: (SizeConstraint) -> CGSize) -> CGSize {
+
+        /// Due to various math upstream, we can end up with measurements like 210.0 getting
+        /// turned into numbers like 209.99999997 due to loss of precision.
+        ///
+        /// Because we cache by exact values in this cache,
+        /// let's round this up to the nearest screen-scale pixel to work around.
+        let constraint = constraint.roundUpToNearestPixel(with: screenScale)
```
kyleve, line 34: I think instead of up, we want to closest value.

BlueprintUI/Tests/StackTests.swift
```diff
@@ -1183,6 +1183,125 @@ class StackTests: XCTestCase {
         )
     }
 
+    func test_measurementCounts() {
```
kyleve, line 1186: Finish me

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -96,11 +96,15 @@ public struct AttributedLabel: Element, Hashable {
         configure(&self)
```
kyleve, line 96: Revert before merge

BlueprintUICommonControls/Tests/Sources/PerformancePlayground.swift
```diff
@@ -24,7 +24,7 @@ class PerformancePlayground: XCTestCase {
 
     override func invokeTest() {
         // Uncomment this line to run performance metrics, eg in Instruments.app.
-        // super.invokeTest()
```
kyleve, line 27: Revert before merge

Package.swift
```diff
@@ -24,7 +24,8 @@ let package = Package(
     targets: [
```
kyleve, line 24: Revert before merge

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -22,12 +25,31 @@ protocol CacheTree: AnyObject {
...
+
+        /// Due to various math upstream, we can end up with measurements like 210.0 getting
+        /// turned into numbers like 209.99999997 due to loss of precision.
+        ///
+        /// Because we cache by exact values in this cache,
+        /// let's round this up to the nearest screen-scale pixel.
+
+        let roundedConstraint = constraint.roundToNearestPixel(with: screenScale)
+
+        // TODO: Why does this break??
```
kyleve, line 37: This breaks one test; which is interesting; I _think_ because test elements are relying on their measurement constraint to be unrounded (even unrounded off of screen pixel values, eg 0.333). This feels "wrong", but didn't change it.

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -270,6 +270,12 @@ public struct StackLayout: Layout {
 
 }
 
+/// Temporary. Allows testing stack layout optimizations to compare performance.
+enum BlueprintStackUpdates {
+
+    static var isOptimizedLayoutEnabled: Bool = true
```
kyleve, line 276: Remove before merging.

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -448,21 +454,88 @@ extension StackLayout {
...
+
+                let fixedMagnitude: CGFloat = measurements.reduce(0) { magnitude, measurement in
+                    magnitude + (measurement.size?.axis(on: axis) ?? 0)
+                }
+
+                // From that, determine how much space the flexible items will have
+
+                let flexibleMagnitude: SizeConstraint.Axis = constraint.axis(on: axis) - fixedMagnitude - minimumTotalSpacing
+
+                // TODO: What if flexibleMagnitude is negative??
```
kyleve, line 504: Looking for input on what to do here – eg if there's one or more fixed elements that take up all the space. I was thinking we may just want to skip measuring the flexible elements entirely, since they basically can't lay out.

BlueprintUI/Tests/StackTests.swift
```diff
@@ -1176,13 +1176,136 @@ class StackTests: XCTestCase {
             ],
             expectedSizes: [
```
kyleve, line 1177: These changes b/c the flexible elements are now measured in the _available_ space, not _all_ the space.

BlueprintUI/Sources/Layout/Stack.swift
n8chur: I might be misunderstanding the issue, but I'd expect we could skip measurement of those flexible items and just use a size of `0`, right?

n8chur (commented):
Would love to get a review on this from @watt—these look like reasonable changes on the surface but having someone with more experience here would be really helpful.

BlueprintUI/Sources/Layout/Stack.swift
kyleve: I think we could – let me see what happened before at layout time.

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -22,12 +25,34 @@ protocol CacheTree: AnyObject {
 extension CacheTree {
```
kyleve, line 25: - [ ] Move `CacheTree` changes to their own PR

kyleve:
Moved CacheTree changes to a new PR: https://github.com/square/Blueprint/pull/362

BlueprintUI/Sources/Layout/Stack.swift
kyleve: I added a test (`test_fixed_elements_overflow_behavior`), and the behavior here mirrors the old behavior – which seems wrong; but at least it's consistent. I'll follow up with lower bounding this to zero.

kyleve: https://github.com/square/Blueprint/pull/363

kyleve:
Ran the Register snapshot tests over this again, and there's a bunch of tiny lil 1px rounding changes, but I only found a couple meaningful differences, which seem to be due to incorrect configuration of the Rows/Columns in the first place, and will be easy to fix!

BlueprintUI/Tests/StackTests.swift
```diff
@@ -1176,13 +1176,162 @@ class StackTests: XCTestCase {
...
+        XCTAssertEqual(fixed3.count, 1)
+
+        XCTAssertEqual(
+            fixed3,
+            [
+                SizeConstraint(width: .atMost(59), height: .atMost(30)),
+            ]
+        )
+
+        XCTAssertEqual(flexible1.count, 1)
```
n8chur, line 1296: what are the counts for each of these before you made these changes?

n8chur (approved):
Looks reasonable to me but would really like to get a review from @watt before this lands if possible!

BlueprintUI/Tests/StackTests.swift
kyleve: Some were 4, and some were 2.

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -448,21 +454,90 @@ extension StackLayout {
...
+                // Measure the flexible items within that constraint
 
-            return totalMeasuredAxis + minimumTotalSpacing
+                measurements.forEach { measurement in
+                    guard measurement.isFlexible else { return }
+
+                    measurement.size = measurement
+                        .item
+                        .content
+                        .measure(in: flexibleConstraint)
```
watt, line 525: Ahh OK, I was wondering how you reduced measurement counts entirely within `_axisSegments`, but it seems to be the same number of calls. If I'm understanding right, it looks like this optimization really hinges on that `CacheTree` change in order to shortcut measurements from `_crossSegments` before they happen. The change to flexible child constraints makes the input values match for the 1-flexible-child case.

watt (commented):
Seeing how this works by relying on pre-caching measurements, rather than explicitly skipping measurements within the stack layout itself, I'm a bit nervous about this change. Gotta think about it a bit more 🤔

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -22,12 +22,23 @@ protocol CacheTree: AnyObject {
...
+            let size = calculation(constraint)
+
+            /// 1) Cache the measured size for the given constraint.
+            self[constraint] = size
+
+            /// 2) Optimization: Cache the measured size, so if we ask for this
+            /// exact size later on, we don't re-measure the element.
+            ///
+            /// This avoids a cache miss later on when a layout lays out an item with a measured size.
+            self[SizeConstraint(size)] = size
```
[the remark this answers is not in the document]
kyleve: Yeahhhhh, I was thinking about this too. It felt rare / weird enough that the tradeoff was worth it, but your point is valid – we can't guarantee folks _don't_ rely on this, so I am also fine just removing it... It _does_ cut the measurement cost in half once again though... Trying to think about how we could keep the "good" parts here...

BlueprintUI/Tests/StackTests.swift
```diff
@@ -1176,13 +1176,162 @@ class StackTests: XCTestCase {
             ],
             expectedSizes: [
                 StackLayout.Vector(axis: 10, cross: 10),
-                StackLayout.Vector(axis: 20, cross: 50), // 2 x 5
-                StackLayout.Vector(axis: 60, cross: 20), // 6 x 2
+                StackLayout.Vector(axis: 27, cross: 50), // 2 x 5
+                StackLayout.Vector(axis: 53, cross: 20), // 6 x 2
```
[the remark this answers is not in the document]
kyleve: Agree. My view on current system:

(a) I do not like the current system, and dare I say no one (outside of like 3 people, and 2 of us are on this PR) understand it (I do not like green eggs and ham)
(b) Every time I see someone trying to use a non-binary priority (eg 0 or 1), it's used wrong
(c) I think this change actually makes things more understandable / behave as you expect for the common case – eg because we're now measuring flexible elements in the actual size they have, it fixes many of the issues called out in https://github.com/square/Blueprint/pull/265 – which you can see from running this in Market – the weird "huh, that row layout sure is wrong" bugs are fixed.

I will update the tests here, but I'd also like to follow up with turning growing and shrinking into bools vs ints, since, yeah...

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -448,21 +454,90 @@ extension StackLayout {
...
         }
 
+        let basisSizes = basisSizes()
+
+        let unconstrainedAxisSize: CGFloat = basisSizes.reduce(0.0, +) + minimumTotalSpacing
+
         switch vectorConstraint.axis {
         case .exactly(let axisSize):
-            if unconstrainedAxisSize() >= axisSize {
+            if unconstrainedAxisSize >= axisSize {
```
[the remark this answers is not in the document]
kyleve: - [ ] Bring back optimization for unconstrained cases

BlueprintUI/Sources/Internal/CacheTree.swift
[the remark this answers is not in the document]
kyleve: (I was also trying to figure out how to lift this level of caching up to this type only, but not really a doable thing rn..)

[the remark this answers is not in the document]
kyleve: Oh I have an idea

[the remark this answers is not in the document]
kyleve: Hmm, thinking about this more... This is OK? Eg you don't have to respect the size returned here. And if you still stretch the returned size, the aforementioned layout will still work... I think?

(Ok we've returned to where I was when I first wrote this code and had the same thought. Trying to figure out how to prove this case beyond "the snapshots in SPOS still pass".)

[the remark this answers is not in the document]
watt: After sleeping on it a bit, I guess I'm OK with this.

I do think that lifting this logic up to to be done explicitly in stack layout is still probably the more "right" thing, for a couple reasons:
1. the previously mentioned correctness argument, since you could definitely build an element that this breaks
2. the way that this interacts with stack layout is not very obvious. If someone else tweaked stack layout a bit we could lose this optimization without knowing why. Or tests would fail for frustratingly unclear reasons.

But practically speaking, I can only come up with hypothetical cases that cause (1), and I'm less worried about (2) knowing that a layout rewrite is likely right around the corner to support SPL. Worth the tradeoffs.

watt (approved):

BlueprintUI/Sources/Internal/CacheTree.swift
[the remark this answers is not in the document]
kyleve: I'm gonna hold off on merging now, and I would like to talk through this a bit more next week – I think these are very valid points and wanna make sure I don't paint us into a corner with this change. I also had a couple other ideas for how to lift / manage the caching a bit better for _only_ the stack layout which I need to play around with.

BlueprintUI/Sources/Layout/Stack.swift
kyleve: Sorry, just saw this comment as I come back to this PR. The cache tree changes get us from 2 measurements to 1 – if you run the `test_measurementCounts` test and turn off `BlueprintStackUpdates.isOptimizedLayoutEnabled`, you can also see the main optimization, of some measurements happening 3, 4, times, etc.

BlueprintUI/Sources/Internal/CacheTree.swift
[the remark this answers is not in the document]
kyleve: I think I have a way to do this on the _get_ end, not the set end. Testing to see what happens.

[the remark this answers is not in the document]
kyleve: Actually wait, that won't work unless we have the concept of cached measurements and if they're natural sizes or not. I might just remove this for now, gonna play with it a bit more.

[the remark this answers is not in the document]
kyleve: Yeah, gonna just get rid of this for now. It reduces the benefit a bit, but it still cuts many measurements down to 2 passes instead of 2 * depth.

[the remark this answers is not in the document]
kyleve: Eg, with the flag disabled: 
image

BlueprintUI/Sources/Layout/Stack.swift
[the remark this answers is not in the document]
kyleve: Done

BlueprintUI/Sources/Layout/Stack.swift
kyleve: Removed

PR #349, opened 2022-05-27
Add .grows and .shrinks to StackLayout.Child.Priority

kyleve:
There's no way to find-tune the grow and shrink priorities with the `.stackLayoutChild(...)` APIs, so adding this so we also don't need to directly let people fiddle with the direct priorities.

### DIFF
--- a/BlueprintUI/Sources/Layout/Stack.swift
+++ b/BlueprintUI/Sources/Layout/Stack.swift
@@ -992,13 +992,28 @@ extension StackLayout {
         public let key: AnyHashable?
 
         public enum Priority {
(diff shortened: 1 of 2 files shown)


watt:
We [intentionally didn't do this](https://github.com/square/Blueprint/pull/220#discussion_r710661281) the first time around. Got a new use case in mind?

kyleve:
Yeah, the main case I wanted was to have a spacer with a min width but will also grow to fill available space. Especially when you're nesting rows in rows setting up this sort of thing becomes useful to distribute them during layout.

kylebshr:
That does seem like a pretty good use case, I'm fine adding these

kylebshr (approved):

PR #352, opened 2022-06-06
Introduce accessibilityFrameCornerStyle onto AccessibilityElement

kyleve:
Need this for some `MarketPill` a11y improvements

IMG_8822


n8chur (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -38,6 +38,7 @@ public struct AccessibilityElement: Element {
     public var identifier: String?
     public var traits: Set<Trait>
     public var accessibilityFrameSize: CGSize?
+    public var accessibilityFrameCornerStyle: Box.CornerStyle
```
kyleve, line 41: You'll likely always want to just match a box's corner style, so using the same value.

PR #354, opened 2022-06-07
Aligned will now constrain content to its max layout width

kyleve:
This fixes issues like this, where wrapping content in a `.center` means it grows outside of the bounds of the view (expected on left, actual before this fix on right)

image

### DIFF
--- a/BlueprintUI/Sources/Layout/Aligned.swift
+++ b/BlueprintUI/Sources/Layout/Aligned.swift
@@ -75,17 +75,25 @@ public struct Aligned: Element {
         }
 
         func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
-            let contentSize = child.measure(in: SizeConstraint(size))
 
-            var attributes = LayoutAttributes(size: contentSize)
+            let measurement = child.measure(in: SizeConstraint(size))
+
+            let constrainedMeasurement = CGSize(
+                width: min(size.width, measurement.width),
+                height: min(size.height, measurement.height)
+            )
+
+            var attributes = LayoutAttributes(
+                size: constrainedMeasurement
+            )
 
             switch verticalAlignment {
             case .top:
                 attributes.frame.origin.y = 0
             case .center:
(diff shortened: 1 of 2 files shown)


n8chur (approved):
Do we need to be concerned about this breaking existing layouts that depend on the existing behavior?

I feel like it's probably worth it, and hopefully covered by snapshot tests if so, but it does seem likely that a consumer may be (perhaps unknowingly) depending on this behavior.

kyleve:
Ugh sorry, turned on automatic merge...

I guess we'll find out. I think this is OK b/c without this wrapper being applied, the wrapper+1 would respect width – the only thing I know that doesn't respect this is `Stack` itself; which has always seemed weird tbh.

kyleve:
> In principal, I don't like this change. I think elements should follow the unix philosophy of doing exactly 1 job and favoring composability, and I don't think it's Aligned's job to apply size constraints.

Perhaps, though I don't really have a better suggestion for you – several elements already do something like this, since it is a weird behavior for this to happen where children can expand outside their parents bounds (usually, except in some unique cases).

I think the main case where this happens / why this happens is due to `Label` (and more specifically, how `UILabel` measures under the hood), so perhaps another way to resolve is to have `Label` constrain its returned measurements to the provided bounds.

> If this is a behavior we want universally, it might make more sense to bake that into the layout engine itself instead of requiring individual elements to do it?

Perhaps – though I don't think it's truly universal: Eg, the `Decorate` element intentionally lets things expand outside its bounds. Maybe something to consider after SPL to see if there's something universal that falls out of common use cases.

watt:
> several elements already do something like this

Yeah, that's really what I want to avoid. That means it's inconsistent _and_ we're duplicating responsibility.

> Perhaps – though I don't think it's truly universal: Eg, the `Decorate` element intentionally lets things expand outside its bounds.

I'm not suggesting that elements universally can't grow beyond their bounds, just that elements should not have to do this clamping behavior manually. It could be built into the `Measurable` vended to layouts, for example.

PR #355, opened 2022-06-08
Prepare 0.43.0

kyleve:


kyleve:
> Did you run an integration test with changes from #354? I think we should.

I tested it in Market and everything worked as expected, but I can test in SPOS too!

kyleve:
Couple small things to fix, but nothing too major! I'll work on an integration PR.

nononoah (approved):

PR #357, opened 2022-06-28
Allow `Image` to block accessibility descriptions of images

RoyalPineapple:
iOS 14 introduced [image descriptions using VoiceOver](https://support.apple.com/en-mn/guide/iphone/iph37e6b3844/ios). This feature uses machine learning magic to provide accurate descriptions of arbitrary images, it's pretty damn impressive.

Unfortunately this is not always desirable, particularly when the app's UI makes heavy use of arbitrary user provided images. Voiceover is particularly aggressive in finding images to describe. 
Any view with `UIAccessibilityTraits.image` will trigger the description regardless of its status as an accessibility element. What's more, the entire view hierarchy containing the view with the `.image` trait  will be analysed by the ML model, including any text that might be overlaid. This can result in a long and duplicative description. 

This PR provides an escape hatch to allow an image to be excluded from accessibility descriptions. 

Check out this screen recording for an example. 

https://user-images.githubusercontent.com/219578/176280560-4bd5905b-ecab-46ef-b2fb-2c617c4b476f.mov

### DIFF
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -15,15 +15,21 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
+    /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always apropreate.
+    /// Set this to `true` to prevent VoiceOver from discribing the displayed image.
+    public var blockAccessibilityDiscription: Bool
+
     /// Initializes an image element with the given `UIImage` instance.
     public init(
         image: UIImage?,
         contentMode: ContentMode = .aspectFill,
-        tintColor: UIColor? = nil
+        tintColor: UIColor? = nil,
+        blockAccessibilityDiscription: Bool = false
     ) {
         self.image = image
         self.contentMode = contentMode
         self.tintColor = tintColor
+        self.blockAccessibilityDiscription = blockAccessibilityDiscription
     }
 
     public var content: ElementContent {
@@ -37,6 +43,10 @@ public struct Image: Element {
             config[\.contentMode] = contentMode.uiViewContentMode
             config[\.layer.minificationFilter] = .trilinear
             config[\.tintColor] = tintColor
+            if blockAccessibilityDiscription {
+                // Seting `isAccessibilityElement = false` isn't enough here, VoiceOver is very aggressive in finding images to discribe. We need to explicitly remove the `.image` trait.
+                config[\.accessibilityTraits] = UIAccessibilityTraits.none
+            }
         }
     }
 
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -15,7 +15,7 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
-    /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always apropreate.
+    /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always appropriate.
     /// Set this to `true` to prevent VoiceOver from discribing the displayed image.
     public var blockAccessibilityDiscription: Bool
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Image` now provides an override to prevent VoiceOver from generating accessibility descriptions.
+
 ### Removed
 
 ### Changed


BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -15,15 +15,21 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
+    /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always apropreate.
```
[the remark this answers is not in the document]
RoyalPineapple: I'll learn to spell one day

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -15,15 +15,21 @@ public struct Image: Element {
...
+    /// Set this to `true` to prevent VoiceOver from discribing the displayed image.
+    public var blockAccessibilityDiscription: Bool
+
     /// Initializes an image element with the given `UIImage` instance.
     public init(
         image: UIImage?,
         contentMode: ContentMode = .aspectFill,
-        tintColor: UIColor? = nil
+        tintColor: UIColor? = nil,
+        blockAccessibilityDiscription: Bool = false
```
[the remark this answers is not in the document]
RoyalPineapple: Unfortunately, no. It just doesn't work. Seems it searches through the view hierarchy completely independently of the accessibility hierarchy.

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -15,15 +15,21 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
+    /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always appropriate.
+    /// Set this to `true` to prevent VoiceOver from discribing the displayed image.
+    public var blockAccessibilityDiscription: Bool
```
kyleve, line 20: I almost wonder if this should be the default and you should have to _opt in_ to this – it feels really unexpected to me generally for it to do this magically.

kyleve: (Especially since it'll vary by OS version – it seems like the cases where we want the ML Magic are few and far in between)

RoyalPineapple: Ok, so I did a deep dive into this and figured out the following.  My research leads me to believe this is a relatively rare situation and we should not be turning this off by default. 
https://docs.google.com/document/d/1nmIoJbVvnde0jJISiJYb-ZPG0kHXJWroP2O7ShMdK-c/edit?usp=sharing

BlueprintUICommonControls/Sources/Image.swift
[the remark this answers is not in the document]
RoyalPineapple: I did some deeper investigation into what causes this, turns out it is triggered by overlapping an image and text.  https://docs.google.com/document/d/1nmIoJbVvnde0jJISiJYb-ZPG0kHXJWroP2O7ShMdK-c/edit?usp=sharing

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -37,6 +43,10 @@ public struct Image: Element {
             config[\.contentMode] = contentMode.uiViewContentMode
             config[\.layer.minificationFilter] = .trilinear
             config[\.tintColor] = tintColor
+            if blockAccessibilityDiscription {
+                // Seting `isAccessibilityElement = false` isn't enough here, VoiceOver is very aggressive in finding images to discribe. We need to explicitly remove the `.image` trait.
+                config[\.accessibilityTraits] = UIAccessibilityTraits.none
```
RoyalPineapple, line 48: i wish i could do `.remove(.image)` here but that's not how this works, in this case its fine because its backed by a native `UIImageView` which has `.image` as its only default trait.

BlueprintUICommonControls/Sources/Image.swift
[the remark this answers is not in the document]
kyleve: Oh wow that's wild. Thanks for digging in!

kyleve (approved):

kylebshr (approved):

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -15,15 +15,21 @@ public struct Image: Element {
...
         image: UIImage?,
         contentMode: ContentMode = .aspectFill,
-        tintColor: UIColor? = nil
+        tintColor: UIColor? = nil,
+        blockAccessibilityDiscription: Bool = false
     ) {
         self.image = image
         self.contentMode = contentMode
         self.tintColor = tintColor
+        self.blockAccessibilityDiscription = blockAccessibilityDiscription
```
kyleve, line 32: I just noticed this is spelled wrong: `blockAccessibilityDiscription` should be `blockAccessibilityDescription`

PR #359, opened 2022-07-11
Fix AccessibilityContainer with children that set accessibilityElementsHidden to true

n8chur:
`AccessibilityContainer` now omits accessibility elements where `.accessibilityElementsHidden` is `true`.

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -61,7 +61,9 @@ extension AccessibilityContainer {
 extension UIView {
     func recursiveAccessibleSubviews() -> [Any] {
         subviews.flatMap { subview -> [Any] in
-            if let accessibilityElements = subview.accessibilityElements {
+            if subview.accessibilityElementsHidden {
+                return []
+            } else if let accessibilityElements = subview.accessibilityElements {
                 return accessibilityElements
             } else if subview.isAccessibilityElement {
                 return [subview]
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -61,7 +61,7 @@ extension AccessibilityContainer {
 extension UIView {
     func recursiveAccessibleSubviews() -> [Any] {
         subviews.flatMap { subview -> [Any] in
-            if subview.accessibilityElementsHidden {
+            if subview.accessibilityElementsHidden || subview.isHidden {
                 return []
             } else if let accessibilityElements = subview.accessibilityElements {
                 return accessibilityElements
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -61,7 +61,9 @@ extension AccessibilityContainer {
 extension UIView {
     func recursiveAccessibleSubviews() -> [Any] {
         subviews.flatMap { subview -> [Any] in
-            if let accessibilityElements = subview.accessibilityElements {
+            if subview.accessibilityElementsHidden {
```
kylebshr, line 64:
```suggestion
            if subview.accessibilityElementsHidden || subview.isHidden {
```

I'm thinking we should check this too

kylebshr (approved):

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
watt: `isHidden` doesn't normally block subviews from accessibility, does it?

watt (approved):

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
kylebshr: Hmm I thought it did! We should definitely match the system

kyleve: IMO I would _expect_ it do (since they're otherwise completely visible from most users). To not definitely feels like a foot gun

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
kyleve: (Does `accessibilityElementsHidden` default to `isHidden`?)

RoyalPineapple (commented):
great catch

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
watt: I agree with KB that we should match the system, so I checked: it looks like `isHidden` does normally suppress nested a11y elements.

RoyalPineapple: This method is also in listable, I'll open a PR to add it there once this lands.

n8chur: So we _do_ want to add the `|| subview.isHidden` check to this?

- [ ] Add `isHidden` check

watt: > So we _do_ want to add the `|| subview.isHidden` check to this?

Yeah, I think so.

PR #361, opened 2022-08-04
Cascade rounding correction to nested layouts

watt:
This is a fix for a new twist on an old classic: frame rounding that causes labels to be truncated.

This time the issue issue appears in nested BlueprintViews. In a "flat" element composition, labels' frame rounding behavior prioritizes sizes over edge alignment to avoid this issue. However, when we have a nested BlueprintView, it acts as a leaf node to the outer BlueprintView, and the layout is effectively broken into two stages. This breaks the rounding error accumulation/correction, and prevents the size prioritization from working.

The fix works like this:

1. Rounding correction info is extended to include size correction in addition to origin.
2. Correction info is cascaded across nested BlueprintViews through internal environment keys.
3. Nested BlueprintViews apply the correction:
    a. the size is applied during layout, so that layout happens in the same bounds it would have in a flat composition
    b. the origin is applied during rounding, so that rounding behavior can also match a flat composition

New pixel boundary snapshot tests validate the expected behavior: both existing tests now run in a "flat" mode and a "nested" mode, where the latter has intermediate BlueprintViews added to the tree. We expect to get the same result in both cases.

There's also a new test for size priority frame rounding, since there didn't seem to be one before.

See UI-2676


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -324,25 +324,33 @@ public final class BlueprintView: UIView {
 
         let environment = makeEnvironment()
 
+        let rootCorrection = environment.roundingCorrection
+        let rootOrigin = environment.roundingOrigin
+        let rootFrame = CGRect(
```
kyleve, line 329: OHHHH. Nice

BlueprintUI/Sources/Internal/RoundingCorrectionKey.swift
```diff
@@ -0,0 +1,13 @@
+import CoreGraphics
+
+/// This key is used internally to cascade rounding information to nested BlueprintViews.
+enum RoundingCorrectionKey: EnvironmentKey {
+    static let defaultValue: CGRect = .zero
+}
+
+extension Environment {
```
kyleve, line 8: Should we stick this on the env, or should we change the inherited (thing) type to something other than env?

kyleve (approved):
This is great – nice work!

BlueprintUI/Sources/Internal/RoundingCorrectionKey.swift
kyleve: (Eg, change this https://github.com/square/Blueprint/blob/main/BlueprintUI/Sources/Environment/Environment.swift#L84 to an `InheritedProperties` or something)

n8chur: @watt would this be preferable?

It does seem like it might be nice to share the `inherited` infra that already exists—it's a pattern that's being used to bridge the same space that this PR is bridging (I think?).

n8chur (approved):
Nice!

BlueprintUI/Sources/Internal/RoundingCorrectionKey.swift
watt: I don't have a strong preference. I just put it on the environment out of convenience, since it already existed. Splitting out would require duplicating some infra, and it doesn't seem like there's much advantage.

PR #362, opened 2022-08-05
Fix Cache Tree Cache Misses

kyleve:
Due to float addition and subtraction that happens throughout layout, it was common to end up with `210.0` getting turned into `209.999997` (etc) throughout. This was causing needless cache misses. To resolve, I introduced pixel-based rounding to the cache tree so the cached constraint keys land on a display pixel.

### DIFF
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -12,6 +12,9 @@ protocol CacheTree: AnyObject {
     /// A reference to use for logging
     var signpostRef: AnyObject { get }
 
+    /// The scale of the screen scale, used for rounding to account for loss of float precision.
+    var screenScale: CGFloat { get }
+
     /// The sizes that are contained in this cache, keyed by size constraint.
     subscript(constraint: SizeConstraint) -> CGSize? { get set }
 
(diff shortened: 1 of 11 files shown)


BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -23,18 +26,29 @@ extension CacheTree {
...
-        if let size = self[constraint] {
+        /// Due to various math upstream, we can end up with measurements like 210.0 getting
+        /// turned into numbers like 209.99999997 due to loss of precision.
+        ///
+        /// Because we cache by exact values in this cache,
+        /// let's round this to the nearest screen-scale pixel to avoid a cache miss.
+
+        let roundedConstraint = constraint.roundToNearestPixel(with: screenScale)
+
+        // TODO: Why does this break??
```
kyleve, line 37: I know why this breaks; just rounding the key we use is correct; we don't want to round the constraint itself. Remove this TODO and rename `roundedConstraint` to `key`

n8chur (approved):
Looks reasonable to me, but would appreciate a review from @watt before this lands!

BlueprintUI/Tests/RenderPassCacheTests.swift
```diff
@@ -3,7 +3,7 @@ import XCTest
 
 final class RenderPassCacheTests: XCTestCase {
     func test_caching() {
```
n8chur, line 5: Should we add a test that validates the rounding logic is cached appropriately?

kyleve: Probably a good idea! Will do.

nononoah (approved):
Seems like a nice easy win here, unless of course rounding fails to pay for itself in cache hits.

I'll defer to Andrew as the expert here, but it might be nice to show as part of PRs like this one that performance has improved as a result.

BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -23,18 +26,29 @@ extension CacheTree {
...
+        /// Due to various math upstream, we can end up with measurements like 210.0 getting
+        /// turned into numbers like 209.99999997 due to loss of precision.
+        ///
+        /// Because we cache by exact values in this cache,
+        /// let's round the key to the nearest screen-scale pixel to avoid a cache miss.
+        ///
+        /// ⚠️ Important! We **do not** want to round the constraint itself, just the key.
+        /// Blueprint allows (and expects) unrounded values – they're rounded / snapped during the view layout pass.
+
+        let key = constraint.roundToNearestPixel(with: screenScale)
```
nononoah, line 38: Since we don't actually update the constraint, how important is the `screenScale` rounding?

PR #365, opened 2022-08-18
[WIP DNR] [Prototype] ViewController-Backed Elements

kyleve:


n8chur (commented):
It'd be worth digging through AirBnB's Epoxy implementation of this concept if we do pursue this path (if you haven't already): https://github.com/airbnb/epoxy-ios/blob/bb59fcc0a7ad08d657226809dc97df5c1c4a2371/Sources/EpoxyCore/SwiftUI/EpoxySwiftUIHostingView.swift#L305C16-L305C49

PR #366, opened 2022-08-30
[logging]: add some logging customization options

jamieQ:


BlueprintUI/Sources/Internal/Logger.swift
```diff
@@ -71,13 +71,40 @@ extension Logger {
...
+    static func logSizeThatFitsStart(
+        view: BlueprintView,
+        description: String
+    ) {
+        guard BlueprintLogging.isEnabled else { return }
+
+        os_signpost(
+            .begin,
+            log: .active,
+            name: "View Sizing",
```
kyleve, line 84: Do we want to call this `BlueprintView Measurement` or otherwise?

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
```diff
@@ -12,6 +12,20 @@ extension OSLog {
 /// Namespace for logging config.
 public enum BlueprintLogging {
 
+    /// Configuration for logging options
+    public struct Config {
+
+        /// When `true`, measurement data will be collected for all `Element`s
+        public var recordMeasurePasses: Bool
```
kyleve, line 19: Maybe this would be better as an enum?

```
enum LogLevel {
    case fullHierarchy
    case completeTimeOnly
}
```
?

watt, line 19: Yeah, "measure pass" is kind of vague in this context. Enum is OK, or if you want a fine-grained config I'd call this something like `recordElementMeasures`.

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
```diff
@@ -25,3 +39,14 @@ public enum BlueprintLogging {
...
 }
+
+extension BlueprintLogging.Config {
+    /// Logging configuration that will not record measurement data for all `Element`s.
+    /// This provides a reasonable balance between data collection and performance impact.
+    /// This is the default logging configuration.
+    public static let lite = Self(recordMeasurePasses: false)
+
+    /// Logging configuration that includes measurement data.
+    /// This provides the most granular information, but has a noticeable impact on performance.
```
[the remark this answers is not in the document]
jamieQ: it's like ~50% slower when all element measures are logged. here's some instruments data collected when swapping b/w checkout & the 'more' tab (debug build & in simulator though, so discount the absolute timings).

| | |
| --- | --- |
| verbose | Screen Shot 2022-08-31 at 10 39 58 AM |
| lite | Screen Shot 2022-08-31 at 10 40 09 AM |

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
jamieQ: that seems reasonable, but since at the moment it just governs whether to emit the element measurement logs, a bool seems sufficient.

BlueprintUI/Sources/Internal/Logger.swift
kyleve: Fair enough!

BlueprintUI/Sources/Internal/OSLog+Blueprint.swift
kyleve: Got it, sounds good then

PR #367, opened 2022-09-02
[WIP DNR] Caching Again

kyleve:


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -510,12 +555,28 @@ extension BlueprintView {
...
+                viewDescription.apply(to: view)
+                appliedViewDescription = true
+            }
 
             // After this view's children are updated, allow it to run a layout pass.
             // This ensures backing view layout changes are contained in animation blocks.
             defer {
-                view.layoutIfNeeded()
+                // TODO: also need to do if we changed the size
+                if appliedViewDescription {
```
kyleve, line 577: Actually just remove this, `ifneeded` will already no-op if we can.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,125 @@ public struct ElementContent {
...
     }
 
-    // MARK: Measurement & Children
-
-    /// Measures the required size of this element's content.
-    /// - Parameters:
-    ///   - constraint: The size constraint.
-    ///   - environment: The environment to measure in.
-    /// - returns: The layout size needed by this content.
-    public func measure(in constraint: SizeConstraint, environment: Environment) -> CGSize {
```
kyleve, line 33: Bring this back.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,82 @@ import UIKit
 ///             }
 ///         }
 ///     }
-public struct Environment {
+public struct Environment: Equatable {
```
kyleve, line 37: Probably shouldn't be `Equatable` itself

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -140,10 +167,29 @@ final class ElementState {
...
+        }
 
         if isEquivalent == false {
             clearAllCachedData()
+
+            if comparability == .comparableElement {
+                recursiveForEach {
+                    $0.clearAllCachedDataIfNotComparable()
+                }
+            }
```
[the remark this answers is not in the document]
kyleve: > This might be expensive, do we need to recurse here?

We do need to, because children of comparable elements always cache measurements.

> Can we guarantee that we always compare parents before their children?

I believe so, since the tree iterates top down; since it's a tree, not a graph.

>  In what case will have a child with isEquivalent == true and a parent who is false?

Thinking about eg, a `Label` within a parent element that contains many `Labels`. Only one of them may have changed!

PR #370, opened 2022-09-09
Improve label performance by optimizing string normalization

kyleve:
An easy optimization; we should only need to do this when laying out – not when measuring.

For the `test_repeated_layouts` case, it provides significant benefit:

```
Before: Iterations: 13, Average Time: 0.24918307707859919
After:  Iterations: 21, Average Time: 0.14575852098919095
```


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Improve `AttributedLabel` measurement performance by avoiding normalizing it's string when not needed.
```
jamieQ, line 12:
```suggestion
- Improve `AttributedLabel` measurement performance by avoiding normalizing its string when not needed.
```

kyleve: One day I'll remember which is which

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -217,9 +217,14 @@ extension AttributedLabel {
             activeLinkAttributes = model.activeLinkAttributes
             linkDetectionTypes = model.linkDetectionTypes ?? []
 
-            attributedText = model
-                .attributedText
-                .normalizingForView(with: model.numberOfLines)
+            if isMeasuring {
+                attributedText = model
```
nononoah, line 221: For any other readers, I'll summarize `normalizingForView(numberOfLines:)`:

- Replace link attributes with custom link attribute to fully control link styling
- Update truncation mode to sensible defaults when bad values are provided. These updates are purported to not impact sizing.

I'm wondering if the first point can change sizing, especially when the default `UILabel` styling for links varies greatly from ours, but if we have snapshot / unit test coverage here that remained untouched by these changes, I would imagine we have little to worry about.

n8chur (approved):

kyleve:
The snapshot tests changed a little bit (pushed them here); though not in a way that seems to particularly matter. Going to try to dig in more to verify..

watt:
> The snapshot tests changed a little bit (pushed them here); though not in a way that seems to particularly matter. Going to try to dig in more to verify..

Some of these do seem kinda significant. I'm not sure that `normalizingForView` is safe to entirely skip during measurement.

Screen Shot 2022-09-19 at 4 40 27 PM

kyleve:
> > The snapshot tests changed a little bit (pushed them here); though not in a way that seems to particularly matter. Going to try to dig in more to verify..
> 
> Some of these do seem kinda significant. I'm not sure that `normalizingForView` is safe to entirely skip during measurement.
> 
> Screen Shot 2022-09-19 at 4 40 27 PM

It does seem that way! Going to poke at a couple things:
- Can we skip it quickly if there's no links – I think that's "cheap" to figure out at the Label level, but this isn't helpful because Market uses AttributedLabel directly.
- I think it's not _too_ hard to directly cache the `content` of elements once and only once during the layout cycle – going to poke at this bit.
- I'm also not 100% sure these matter "in practice" in actual use cases we see in POS – so maybe we just ignore it?

kylebshr:
I think they do matter in practice, since the sizing is now wrong - labels certainly get truncated in SPOS, right?

kylebshr:
I wonder if you'd see a performance increase by removing the use of `AttributedText` in `normalizingForView`, and using the NSAttributedString API directly? I think that version of the function may even be in the git history

watt:
> > > The snapshot tests changed a little bit (pushed them here); though not in a way that seems to particularly matter. Going to try to dig in more to verify..
> > 
> > 
> > Some of these do seem kinda significant. I'm not sure that `normalizingForView` is safe to entirely skip during measurement.
> > Screen Shot 2022-09-19 at 4 40 27 PM
> 
> It does seem that way! Going to poke at a couple things:
> 
> * Can we skip it quickly if there's no links – I think that's "cheap" to figure out at the Label level, but this isn't helpful because Market uses AttributedLabel directly.
> * I think it's not _too_ hard to directly cache the `content` of elements once and only once during the layout cycle – going to poke at this bit.
> * I'm also not 100% sure these matter "in practice" in actual use cases we see in POS – so maybe we just ignore it?

Another option I think would be pretty safe is to cache the result of `normalizingForView` so we only call it once per `AttributedLabel`.

kyleve:
> Another option I think would be pretty safe is to cache the result of normalizingForView so we only call it once per AttributedLabel.

I tried this, and it ends up not materially helping a lot, because the `content` of the element is still called many many times – which happens if you're in a "lazy" element like an `EnvironmentReader`, eg:

```
EnvironmentReader { env in
   MarketLabel(...)
}
```

Each time the `EnvironmentReader` is measured, etc, the label is re-created.

kyleve:
I think I got per pass content caching working, which resolves this issue by moving the caching up a layer – closing this. Sorry for the noise!

PR #374, opened 2022-09-28
[SPL] `LayoutMode` flag

watt:
A new `LayoutMode` type to switch between standard and single-pass.

- You can set this on a `BlueprintView`. It doesn't do anything yet.
- If it's not set, the view will inherit the mode of its nearest ancestor (via the current plumbing for environments)
- If there's no ancestor, the global default is `standard`.
- The global default cannot be changed at runtime (yet?).

The layout mode also needs to be bridged to "out-of-band" operations (e.g. measuring elements in a GeometryReader). To facilitate that, there's also a new `RenderContext` type that is available statically during a render pass. Currently it only holds the layout mode but it might also be useful later to bridge the CacheTree or other stuff.

Note that this PR targets a feature branch.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -99,6 +99,16 @@ public final class BlueprintView: UIView {
             invalidateIntrinsicContentSize()
         }
     }
+    
+    /// An optional explicit layout mode for this view. If `nil`, this view will inherit the layout
+    /// mode of its nearest ancestor, or use ``LayoutMode/default``.
+    public var layoutMode: LayoutMode? {
+        didSet {
+            if layoutMode != oldValue {
+                setNeedsViewHierarchyUpdate()
+            }
+        }
+    }
 
     /// An optional name to help identify this view
     public var name: String?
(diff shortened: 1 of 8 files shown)


BlueprintUI/Sources/Internal/LayoutModeKey.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+enum LayoutModeKey: EnvironmentKey {
+    static let defaultValue: LayoutMode = .default
+}
+
+extension Environment {
+    /// This mode will be inherited by descendant BlueprintViews that do not have an explicit
+    /// mode set.
+    var layoutMode: LayoutMode {
```
watt, line 10: I put this here for now to make the inheritance convenient, may wind up removing it from Environment so that we're not managing two sources of truth (this and RenderContext).

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -99,6 +99,16 @@ public final class BlueprintView: UIView {
             invalidateIntrinsicContentSize()
         }
     }
+    
+    /// An optional explicit layout mode for this view. If `nil`, this view will inherit the layout
+    /// mode of its nearest ancestor, or use ``LayoutMode/default``.
+    public var layoutMode: LayoutMode? {
```
kyleve, line 105: Thinking outloud:
- Do we want to persist a previous "global" default value in user defaults? Eg imagining without persisting it somewhere, for a bit as the app starts up (or if you're logged out and we drive this from an account/status flag), we'll be using the "standard" value – which could also result in layouts being laid out one time with standard layout, and another time with SPL

kyleve (approved):

n8chur (approved):
Nice!

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
watt: > Thinking outloud:
> 
> * Do we want to persist a previous "global" default value in user defaults? Eg imagining without persisting it somewhere, for a bit as the app starts up (or if you're logged out and we drive this from an account/status flag), we'll be using the "standard" value – which could also result in layouts being laid out one time with standard layout, and another time with SPL

Maybe. One reason I made the global a constant is that if you change it at runtime you need to invalidate _everything_, and there's currently no way to do that. We'd need every BlueprintView to listen to a global notification or something. It's much easier to set individual views to whatever value the app wants.

PR #377, opened 2022-10-20
[SPL] Debug and test helpers

watt:
I pulled these out of some other work, because I think they'll be handy.

First a `Metatype` wrapper that makes element IDs and paths readable. Given a path like this:
```swift
let path = ElementPath()
    .appending(identifier: .init(elementType: Row.self, key: nil, count: 0))
    .appending(identifier: .init(elementType: Column.self, key: nil, count: 1))
    .appending(identifier: .init(elementType: Empty.self, key: "k", count: 0))
```
Before:
`ObjectIdentifier(0x0000000103407fc0).0ObjectIdentifier(0x0000000103406210).1ObjectIdentifier(0x0000000103406c60).k.0`
After:
`Row.0/Column.1/Empty.k.0`

Second, a couple methods that we can hook into for external tests: `BlueprintView.forceSynchronousLayout` and `LayoutResultNode.dump`. Both of these are behind `@_spi(Debugging)`.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -99,7 +99,7 @@ public final class BlueprintView: UIView {
             invalidateIntrinsicContentSize()
         }
     }
-    
+
     /// An optional explicit layout mode for this view. If `nil`, this view will inherit the layout
     /// mode of its nearest ancestor, or use ``LayoutMode/default``.
     public var layoutMode: LayoutMode? {
@@ -224,7 +224,7 @@ public final class BlueprintView: UIView {
         let environment = makeEnvironment()
         let layoutMode = environment.layoutMode
         let renderContext = RenderContext(layoutMode: layoutMode)
-        
+
         let measurement = renderContext.perform {
             element.content.measure(
                 in: constraint,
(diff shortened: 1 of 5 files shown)


kyleve (approved):
Omg thank you for fixing the path printing!

BlueprintUI/Sources/Internal/Metatype.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+/// A wrapper to make metatypes easier to work with, providing Equatable, Hashable, and
+/// CustomStringConvertible.
+struct Metatype: Hashable, CustomStringConvertible {
```
nononoah, line 5: Love the way this pans out.

nononoah (approved):

BlueprintUI/Sources/Internal/Metatype.swift
kyleve: (Overnight thought: I do wonder if it's worth comparing the old and new hashing and == speeds; since this is in some pretty hot paths to verify nothing much changed)

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -423,6 +432,13 @@ public final class BlueprintView: UIView {
         return rootController.children
     }
 
+    /// Forces a synchronous layout, for testing purposes.
+    @_spi(Debugging)
```
n8chur, line 436: Ah, hm, I wonder if we should/could drop the contextual prefixes from our SPI annotations without causing conflicts (e.g. you have two frameworks with a `Debugging` SPI namespace that you want to import in a single file).

n8chur (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: I personally prefer the prefix of `Market`, `Blueprint`, etc since it's easier to cmd + f for all the spis for a single module, as well.

watt: > Ah, hm, I wonder if we should/could drop the contextual prefixes from our SPI annotations without causing conflicts (e.g. you have two frameworks with a `Debugging` SPI namespace that you want to import in a single file).

From a technical perspective that shouldn't be a problem, since these are scoped to each import.

I don't have a strong preference on naming, since this will be used in like 2 places. ¯\_(ツ)_/¯

PR #382, opened 2022-11-03
[Caching] Adds changes to Environment and EnvironmentKey needed for cross-render caching

kdaker:
This PR adds updates to `Environment` and `EnvironmentKey` that needed for [cross-render caching](https://docs.google.com/document/d/1ozD6XaZYZ2a8xRudJrZ-4m3tcH6d8N4bBdtFTjks4QM/edit#heading=h.qmr9cyd5p0i7). 

Summary of changes:
* Makes `Environment` equatable and adds equivalency to `EnvironmentKey`
* Adds default equivalency implementations to `Environment` and `EnvironmentKeys`
* Adds ability to get a `Subset` of keys from `Environment` and  compare an `Environment` with a `Subset` 
* We can now subscribe to `Environment` reads through `subscribeToReads()`

### DIFF
--- a/BlueprintUI/Sources/Environment/Environment.swift
+++ b/BlueprintUI/Sources/Environment/Environment.swift
@@ -34,33 +34,103 @@ import UIKit
 ///             }
 ///         }
 ///     }
-public struct Environment {
+public struct Environment: Equatable {
+
+    typealias OnDidRead = (StorageKey) -> Void
+
+    enum LayoutPass: Equatable {
+        case measurement
+        case layout
+    }
 
     /// A default "empty" environment, with no values overridden.
     /// Each key will return its default value.
     public static let empty = Environment()
 
-    private var values: [ObjectIdentifier: Any] = [:]
-
-    /// Gets or sets an environment value by its key.
-    public subscript<Key>(key: Key.Type) -> Key.Value where Key: EnvironmentKey {
-        get {
-            let objectId = ObjectIdentifier(key)
-
-            if let value = values[objectId] {
-                return value as! Key.Value
-            }
-
-            return key.defaultValue
-        }
-        set {
-            values[ObjectIdentifier(key)] = newValue
-        }
-    }
+    /// When enabled, notify any subscribers when a value is read
+    /// from the environment
+    var readNotificationsEnabled: Bool = true
 
     /// If the `Environment` contains any values.
     var isEmpty: Bool {
         values.isEmpty
+    }
+
+    private var values: [StorageKey: ValueBox] = [:]
+
+    private var measurementDidRead: OnDidRead?
+    private var layoutDidRead: [OnDidRead] = []
+
+    /// Gets or sets an environment value by its key.
+    public subscript<KeyType: EnvironmentKey>(key: KeyType.Type) -> KeyType.Value {
+        get {
+            let storageKey = StorageKey(key)
+
+            if readNotificationsEnabled {
+                measurementDidRead?(storageKey)
+                for reader in layoutDidRead {
+                    reader(storageKey)
+                }
+            }
+
+            if let value = values[storageKey] {
+                return value.base as! KeyType.Value
+            }
+            return key.defaultValue
+        }
+        set {
+            values[StorageKey(key)] = ValueBox(newValue)
+        }
+    }
+
+    public static func == (lhs: Environment, rhs: Environment) -> Bool {
+        guard lhs.values.count == rhs.values.count else { return false }
+
+        for (key, value) in lhs.values {
+            if key.valuesEqual(value.base, rhs.values[key]?.base) == false {
+                return false
+            }
+        }
+
+        return true
+    }
+
+    func valuesEqual(to subset: Environment.Subset?) -> Bool {
+
+        guard let subset = subset else { return true }
+
+        for (key, value) in subset.values {
+            if key.valuesEqual(value.base, values[key]?.base) == false {
+                return false
+            }
+        }
+
+        return true
+    }
+
+    func subset(with keys: Set<StorageKey>) -> Environment.Subset? {
+
+        if keys.isEmpty { return nil }
+
+        var subset = Subset()
+        subset.values.reserveCapacity(keys.count)
+
+        for key in keys {
+            if let value = values[key] {
+                subset.values[key] = value
+            }
+        }
+
+        return subset
+    }
+
+    mutating func subscribeToReads(for layoutPass: LayoutPass, callback: @escaping OnDidRead) {
+        switch layoutPass {
+        case .measurement:
+            measurementDidRead = callback
+        case .layout:
+            layoutDidRead.append(callback)
+        }
     }
 
     /// Returns a new `Environment` by merging the values from `self` and the
@@ -73,6 +143,49 @@ public struct Environment {
     }
 }
 
+extension Environment {
+
+    /// A subset of `Environment.value`s that can be used for matching a set of
+    /// key-value pairs with other `Environment`s
+    struct Subset {
+        var values: [StorageKey: ValueBox] = [:]
+    }
+
+    /// Place values in the environment into a reference box, so that their storage is only allocated once
+    /// even when placed into an environment subset for change tracking.
+    final class ValueBox {
+        let base: Any
+
+        init(_ base: Any) {
+            self.base = base
+        }
+    }
+
+    struct StorageKey: Hashable {
+
+        private let identifier: ObjectIdentifier
+
+        private let isEqual: (Any?, Any?) -> Bool
+
+        init<KeyType: EnvironmentKey>(_ key: KeyType.Type) {
+            identifier = ObjectIdentifier(key)
+
+            isEqual = KeyType.areValuesEqual
+        }
+
+        fileprivate func valuesEqual(_ lhs: Any?, _ rhs: Any?) -> Bool {
+            isEqual(lhs, rhs)
+        }
+
+        func hash(into hasher: inout Hasher) {
+            hasher.combine(identifier)
+        }
+
+        static func == (lhs: Self, rhs: Self) -> Bool {
+            lhs.identifier == rhs.identifier
+        }
+    }
+}
 
 extension UIView {
 
@@ -84,7 +197,7 @@ extension UIView {
     var inheritedBlueprintEnvironment: Environment? {
         if let environment = nativeViewNodeBlueprintEnvironment {
             return environment
-        } else if let superview = self.superview {
+        } else if let superview = superview {
             return superview.inheritedBlueprintEnvironment
         } else {
             return nil
@@ -105,3 +218,4 @@ extension UIView {
 
     private static var environmentKey = NSObject()
 }
+
--- a/BlueprintUI/Sources/Environment/Environment.swift
+++ b/BlueprintUI/Sources/Environment/Environment.swift
@@ -36,20 +36,9 @@ import UIKit
 ///     }
 public struct Environment: Equatable {
 
-    typealias OnDidRead = (StorageKey) -> Void
-
-    enum LayoutPass: Equatable {
-        case measurement
-        case layout
-    }
-
     /// A default "empty" environment, with no values overridden.
     /// Each key will return its default value.
     public static let empty = Environment()
-
-    /// When enabled, notify any subscribers when a value is read
-    /// from the environment
-    var readNotificationsEnabled: Bool = true
 
     /// If the `Environment` contains any values.
     var isEmpty: Bool {
@@ -58,19 +47,13 @@ public struct Environment: Equatable {
 
     private var values: [StorageKey: ValueBox] = [:]
 
-    private var measurementDidRead: OnDidRead?
-    private var layoutDidRead: [OnDidRead] = []
-
     /// Gets or sets an environment value by its key.
     public subscript<KeyType: EnvironmentKey>(key: KeyType.Type) -> KeyType.Value {
         get {
             let storageKey = StorageKey(key)
 
-            if readNotificationsEnabled {
-                measurementDidRead?(storageKey)
-                for reader in layoutDidRead {
-                    reader(storageKey)
-                }
+            for reader in readSubscriptions {
+                reader(storageKey)
             }
 
             if let value = values[storageKey] {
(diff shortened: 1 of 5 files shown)


BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,111 @@ import UIKit
...
+
+    enum LayoutPass: Equatable {
+        case measurement
+        case layout
+    }
 
     /// A default "empty" environment, with no values overridden.
     /// Each key will return its default value.
-    public static let empty = Environment()
+    public static var empty: Environment {
```
kyleve, line 48: Nit (this was my bad sorry); I think this can go back to how it was.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,111 @@ import UIKit
...
+        .init()
+    }
+
+    var readNotificationsEnabled: Bool = true
```
kyleve, line 52: Can we add some docs for this explaining how it's used?

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,111 @@ import UIKit
...
+        values.isEmpty
+    }
+
+    private var values: [StorageKey: ValueBox] = [:]
+
+    private var measurementDidRead: OnDidRead?
+    private var layoutDidRead: [OnDidRead] = []
 
-    private var values: [ObjectIdentifier: Any] = [:]
+    private var onDidReadSubscriptions: [OnDidRead] {
```
kyleve, line 64: Nit: Rather than building an array, I'd just call the two sequentially inline in the code; since there's a small but non-zero cost to building this array.

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -73,6 +151,51 @@ public struct Environment {
     }
 }
 
+extension Environment {
+
+    struct Subset {
```
kyleve, line 156: Nit: Document me

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -73,6 +151,51 @@ public struct Environment {
...
+            self.base = base
+        }
+    }
+
+    struct StorageKey: Hashable {
+
+        private let identifier: ObjectIdentifier
+
+        private let isEqual: (Any?, Any?) -> Bool
+        private let keyTypeName: () -> String
```
kyleve, line 175: This is my bad (sorry), but I think we should just pass the type in here instead, probably.

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,27 @@ public protocol EnvironmentKey {
     /// The default value that will be vended by an `Environment` for this key if no other value
     /// has been set.
     static var defaultValue: Self.Value { get }
+
+    /// Equivalency check on the `Value`s of `EnvironmentKey`s
```
kyleve, line 30: It would be good to explain what "equivalency" means here; eg "return false if the difference in values could affect the measurement or layout of a consuming element..."

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,27 @@ public protocol EnvironmentKey {
...
+
+extension EnvironmentKey where Value: Equatable {
+
+    public static func isEquivalent(_ lhs: Value, _ rhs: Value) -> Bool {
+        lhs == rhs
+    }
+}
+
+extension EnvironmentKey {
+    static func anyEquals(_ lhs: Any?, _ rhs: Any?) -> Bool {
```
kyleve, line 42: Maybe `anyValuesEqual`?

BlueprintUI/Tests/EnvironmentKeyTests.swift
```diff
@@ -0,0 +1,14 @@
+import XCTest
+@testable import BlueprintUI
```
kyleve, line 2: Would be good to also test `anyEquals`

kyleve (approved):

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,103 @@ import UIKit
...
+    }
 
     /// A default "empty" environment, with no values overridden.
     /// Each key will return its default value.
     public static let empty = Environment()
 
-    private var values: [ObjectIdentifier: Any] = [:]
+    /// When enabled, notify any subscribers when a value is read
+    /// from the environment
+    var readNotificationsEnabled: Bool = true
```
n8chur, line 52: When would you want to disable this? Could we just rely on the presence of subscriptions instead?

kyleve: We turn this off when adapting the environment so erroneous reads aren't fired; which is coming in a later PR: https://github.com/square/Blueprint/pull/367/files#diff-07c71c5dcdc5da4c2ddf51b560695e562af396d4afafe3908b35656a0907459aR568

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -73,6 +143,49 @@ public struct Environment {
...
+    /// even when placed into an environment subset for change tracking.
+    final class ValueBox {
+        let base: Any
+
+        init(_ base: Any) {
+            self.base = base
+        }
+    }
+
+    struct StorageKey: Hashable {
```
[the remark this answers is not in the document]
kyleve: I don't _think_ so, since `.Type`s aren't `Hashable`. We can likely heavily simplify this type though!

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,28 @@ public protocol EnvironmentKey {
...
+
+extension EnvironmentKey where Value: Equatable {
+
+    public static func isEquivalent(_ lhs: Value, _ rhs: Value) -> Bool {
+        lhs == rhs
+    }
+}
+
+extension EnvironmentKey {
+    static func areValuesEqual(_ lhs: Any?, _ rhs: Any?) -> Bool {
```
n8chur, line 43: Nit/thought:
```suggestion
    @_disfavoredOverload
    static func isEquivalent(_ lhs: Any?, _ rhs: Any?) -> Bool {
```

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,103 @@ import UIKit
...
 ///     }
-public struct Environment {
+public struct Environment: Equatable {
+
+    typealias OnDidRead = (StorageKey) -> Void
+
+    enum LayoutPass: Equatable {
+        case measurement
+        case layout
+    }
```
n8chur, line 44: We can make this change later if we decide it's worth it, but if we wanted to promote these changes to `MarketEnvironment` it would be nice if this was an extendible "scope" instead of being tied to measure and layout.

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
```diff
@@ -15,16 +15,25 @@ import UIKit
...
+    func isEquivalent(to other: URLHandler) -> Bool
 }
 
 struct DefaultURLHandler: URLHandler {
     func onTap(url: URL) {
         UIApplication.shared.open(url, options: [:], completionHandler: nil)
     }
+
+    func isEquivalent(to other: URLHandler) -> Bool {
+        true
```
n8chur, line 27: Do we need to ensure the type hasn't changed?
```suggestion
        other is Self
```

If not, why not just always return `true` in the `URLHandlerEnvironmentKey`'s `isEquivalent`?

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
```diff
@@ -41,4 +50,8 @@ struct ClosureURLHandler: URLHandler {
     func onTap(url: URL) {
         onTap(url)
     }
+
+    func isEquivalent(to other: URLHandler) -> Bool {
```
n8chur, line 54: If I change the logic for the `onTap` between updates, will this handler still be re-applied to the backing view (even though it's considered equal)?

BlueprintUI/Sources/Environment/EnvironmentKey.swift
kyleve: We think it's important to include the word `Value` in here, since the type signature is just two `Any`s, which could be construed to be the keys. Perhaps `areValuesEquivalent`?

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
kdaker: I don't it matters right? Given these URL handlers changing don't affect layout or measurement. Is there a case where we would want to remeasure if this handler changes?

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
kdaker: Yes it will get re-applied. The equality check is only used to cache the `ElementState` however the `viewDescription` will always be called regardless of the element's cache state.

BlueprintUI/Sources/Environment/Environment.swift
kyleve: Had the same thought; it should be easy enough to do this when we extract the environment, but for now I think this is sufficient.

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
n8chur: I suppose it shouldn't matter, but then when would it ever? Why have `isEquivalent` on the URLHandler protocol?

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,28 @@ public protocol EnvironmentKey {
     /// The default value that will be vended by an `Environment` for this key if no other value
     /// has been set.
     static var defaultValue: Self.Value { get }
+
+    /// Equivalency check on the `Value`s of `EnvironmentKey`s. This should return false if the
+    /// difference in values affects the measurement or layout of consuming elements.
+    static func isEquivalent(_ lhs: Value, _ rhs: Value) -> Bool
```
n8chur, line 32: Thought: Should we name these functions so that they describe that they're comparing only for layout/measurement purposes and not for strict content equability? `isEquivalentForLayout(...)`? I suppose not as I can't think of a good name...

n8chur: I suppose this question might be more relevant in [the next PR](https://github.com/square/Blueprint/pull/383/files#diff-a8aa53da00aac8da1ca838dad2feccf6c13b44e7e1d336db033af8e76aeab8ecR29-R32) (for the `ComparableElement` function)

BlueprintUI/Sources/Environment/Environment.swift
kyleve: Update: Since this a relatively small optimization; we just removed this in a later PR.

BlueprintUI/Tests/EnvironmentTests.swift
```diff
@@ -179,6 +179,188 @@ class EnvironmentTests: XCTestCase {
...
+
+            environment.subscribeToReads(for: .layout) { key in
+                readKey = key
+                exp.fulfill()
+            }
+
+            environment[TestKey.self] = .right
+            _ = environment[TestKey.self]
+
+            wait(for: [exp], timeout: 0.1)
```
kyleve, line 199: Fwiw, this logging is synchronous, we shouldn't need this at all. Once `_ = environment[TestKey.self]` is performed, so are the tracking callbacks.

kyleve: - [ ] Remove test expectations

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
kyleve: Removing for now. If we bring back `appliesViewDescriptionIfNotEquivalent` we'll need some finessing, but we can figure that out then.

n8chur (approved):
Approving merge to feature branch to help make iterating on further work easier, but I'd like to do a full pass at the feature branch before it lands!

BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -34,33 +34,103 @@ import UIKit
...
+
+    /// If the `Environment` contains any values.
+    var isEmpty: Bool {
+        values.isEmpty
+    }
+
+    private var values: [StorageKey: ValueBox] = [:]
+
+    private var measurementDidRead: OnDidRead?
+    private var layoutDidRead: [OnDidRead] = []
```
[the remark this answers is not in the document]
kyleve: A very good question!

We differentiate these because measurements are "flat", eg we only care about the dependencies read during the direct element's measurement – because every layer of the tree that can caches its own measurements. However for layouts, we cache entire trees of layout – including our children – so we want to record and track the environment dependencies of our children as well, and re-perform the layouts if those values change as well.

Does that make sense? I feel like I should probably describe this both better in the code docs here, and in some explanatory tests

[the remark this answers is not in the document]
n8chur: > Does that make sense?

I think so!

> I feel like I should probably describe this both better in the code docs here, and in some explanatory tests

+ 1

---

If we used an array to describe the measurement subscriptions (just to simplify the implementation a bit within environment and perhaps prepare the shape better for supporting "scopes" rather than `enum LayoutPass`) would we need to manually clear them out for some reason, or would there just end up being only 1 subscription of that scope?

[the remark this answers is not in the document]
kyleve: I think we can simplify this to a single array!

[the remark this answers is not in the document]
kyleve: Thinking forward through this more, we might actually want to simplify measurement to be append as well. In theory, once we allow measurements and layouts to read the environment, downstream elements could pull a value from the environment, which when the parent measures its children, would affect their size.

PR #383, opened 2022-11-03
[Caching] Add ElementState and ComparableElement types

kyleve:
- Copies over `ElementState` and `ComparableElement` types, and adds documentation for both of them.
- Adds `ComparableElement` tests.
- Adds tests for `ElementStateTree`.
- Adds skeleton tests for `ElementState`. To keep the PR from gaining another 1k lines, these tests will follow in the next PR.

### DIFF
--- a/BlueprintUI/Sources/Element/ComparableElement.swift
+++ b/BlueprintUI/Sources/Element/ComparableElement.swift
@@ -1,0 +1,86 @@
+//
+//  ComparableElement.swift
+//  BlueprintUI
+//
+//  Created by Kyle Van Essen on 7/3/21.
+//
+
+import Foundation
+import UIKit
+
+
+/// An `ComparableElement` is an element which can cache its
+/// measurements and layouts across layout passes in order to dramatically improve
+/// performance of repeated layouts.
+///
+/// Some element types within Blueprint already conform to `ComparableElement`.
+/// If your element can be easily compared for equality, consider making your
+/// it conform to `ComparableElement` for improved performance across layouts.
+public protocol ComparableElement: AnyComparableElement {
+
+    ///
+    /// Indicates if the element is equivalent to the other provided element.
+    /// Return true if the elements are the same, or return false if something about
+    /// the element changed, such as attributes that affect measurement or layout.
+    ///
+    /// Note that even if this method returns true, the `ViewDescription`
+    /// backing the element will still be re-applied to the on-screen view.
+    ///
+    /// A `ComparableElementContext` parameter is provided, in case
+    /// you need access to the `Environment`, etc, to perform your comparison.
+    /// This is useful, for example, if your `Element` contains builders that are used to
+    /// build the final element structure.
+    ///
+    /// If your element conforms to `Equatable`, this method is synthesized automatically.
+    ///
+    func isEquivalent(to other: Self, in context: ComparableElementContext) -> Bool
+}
+
+
+/// A context object pased to `isEquivalent(to:in:)` that contains
+/// contextual information in which the `Element` is being compared. This
+/// is useful, for example, if you need information from the `Environment`
+/// to perform the comparison.
+public struct ComparableElementContext {
+
+    /// The `Environment` in which the comparison is being performed.
+    public var environment: Environment
+
+    /// Creates a new context object.
+    public init(
+        environment: Environment
+    ) {
+        self.environment = environment
+    }
+}
+
+
+/// Provides a default `ComparableElement` implementation for `Equatable` elements.
+extension ComparableElement where Self: Equatable {
+
+    public func isEquivalent(to other: Self, in context: ComparableElementContext) -> Bool {
+        self == other
+    }
+}
+
+
+/// A type-erased version of `ComparableElement`, allowing the comparison
+/// of two arbitrary elements, and allowing direct access to methods, without self or associated type constraints.
+///
+/// ## Note
+/// You do not need to implement this protocol yourself. It is implemented by Blueprint on `ComparableElement`.
+public protocol AnyComparableElement: Element {
+
+    /// Returns true if the two elements are the same type, and are equivalent.
+    func anyIsEquivalent(to other: AnyComparableElement, in context: ComparableElementContext) -> Bool
+}
+
+
+extension ComparableElement {
+
+    public func anyIsEquivalent(to other: AnyComparableElement, in context: ComparableElementContext) -> Bool {
+        guard let other = other as? Self else { return false }
+
+        return isEquivalent(to: other, in: context)
+    }
+}
--- a/BlueprintUI/Sources/Element/ComparableElement.swift
+++ b/BlueprintUI/Sources/Element/ComparableElement.swift
@@ -20,45 +20,24 @@ public protocol ComparableElement: AnyComparableElement {
 
     ///
     /// Indicates if the element is equivalent to the other provided element.
-    /// Return true if the elements are the same, or return false if something about
-    /// the element changed, such as attributes that affect measurement or layout.
+    /// Return `true` if the elements are the same, or return `false` if something about
+    /// the element changed that will affect measurement or layout.
     ///
-    /// Note that even if this method returns true, the `ViewDescription`
+    /// ## Note
+    /// Even if this method returns true, the `ViewDescription`
     /// backing the element will still be re-applied to the on-screen view.
     ///
-    /// A `ComparableElementContext` parameter is provided, in case
-    /// you need access to the `Environment`, etc, to perform your comparison.
-    /// This is useful, for example, if your `Element` contains builders that are used to
-    /// build the final element structure.
-    ///
+    /// ## Equatable
     /// If your element conforms to `Equatable`, this method is synthesized automatically.
     ///
-    func isEquivalent(to other: Self, in context: ComparableElementContext) -> Bool
-}
-
-
-/// A context object pased to `isEquivalent(to:in:)` that contains
-/// contextual information in which the `Element` is being compared. This
-/// is useful, for example, if you need information from the `Environment`
-/// to perform the comparison.
-public struct ComparableElementContext {
-
-    /// The `Environment` in which the comparison is being performed.
-    public var environment: Environment
-
-    /// Creates a new context object.
-    public init(
-        environment: Environment
-    ) {
-        self.environment = environment
-    }
+    func isEquivalent(to other: Self) -> Bool
 }
 
 
 /// Provides a default `ComparableElement` implementation for `Equatable` elements.
 extension ComparableElement where Self: Equatable {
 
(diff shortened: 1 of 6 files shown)


BlueprintUI/Sources/Element/ComparableElement.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// If your element conforms to `Equatable`, this method is synthesized automatically.
+    ///
+    func isEquivalent(to other: Self, in context: ComparableElementContext) -> Bool
+}
+
+
+/// A context object pased to `isEquivalent(to:in:)` that contains
+/// contextual information in which the `Element` is being compared. This
+/// is useful, for example, if you need information from the `Environment`
+/// to perform the comparison.
```
n8chur, line 43: A more concrete example might be helpful here, like when you'd need an `Environment` to actually build a `Builder` in order to compare the result (right?).

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+                    environment: newEnvironment
+                )
+            )
+
+        case .childOfComparableElement:
+            /// We're always, equivalent, because our parent
+            /// determines our equatability from its state.
+            /// It will also invalidate and throw away _our_ caches if its
+            /// equivalency changes.
+            isEquivalent = true
```
n8chur, line 234: Nit/thought: I understand that this is correct according to the logic, but it's a bit misleading in terms of the naming and how it gets set on `wasUpdateEquivalent`. Perhaps `wasUpdateEquivalent` and this local val should be optional or a tri-state enum to represent the fact that no update was made?

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+    /// or return a cached layout tree if one exists.
+    func layout(
+        in size: CGSize,
+        with environment: Environment,
+        using layout: (Environment) -> [LayoutResultNode]
+    ) -> [LayoutResultNode] {
+
+        /// Layout is only invoked once per cycle. If we're not a `ComparableElement`,
+        /// there's no point in caching this value, so just return the `layout` directly.
+        guard comparability == .comparableElement else {
```
n8chur, line 326: Could we add an explicit note here about why it's not necessary to check for `comparability == .childOfComparableElement`?

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+
+        switch layoutPass {
+        case .measurement:
+            if comparability.isComparable {
+                break
+            } else {
+                return (toTrack(environment), nil)
+            }
+        case .layout:
+            if comparability == .comparableElement {
```
n8chur, line 368: Same here about noting why `.childOfComparableElement` is treated differently from `.comparableElement` (or just move this logic into the call site so we can document this behavior and reason about it in one place).

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+        in toTrack: (Environment) -> Output
+    ) -> (Output, Environment.Subset?) {
+
+        switch layoutPass {
+        case .measurement:
+            if comparability.isComparable {
+                break
+            } else {
+                return (toTrack(environment), nil)
+            }
```
n8chur, line 366: IMO these conditionally ignored paths would be easier to reason about at the call site even if it does add a few more lines there.

From the callsite it looks like it's going to track reads, then you have to dig into this function to discover that it's completely ignored.

I suppose I'd be OK with adding `IfNeeded` to the function name if we feel strongly about this shape.

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+        with identifier: ElementIdentifier
+    ) -> ElementState {
+
+        if let existing = children[identifier] {
+
+            if existing.wasVisited == false {
+                existing.update(with: child, in: environment, identifier: identifier)
+            }
+
+            existing.wasVisited = true
```
n8chur, line 404: Nit:
```suggestion
                existing.update(with: child, in: environment, identifier: identifier)
                existing.wasVisited = true
            }
```

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+        recursiveForEach {
+            $0.wasVisited = false
+            $0.wasUpdateEquivalent = false
+        }
+    }
+
+    /// To be called at the end of the layout cycle by the owner
+    /// of the `ElementStateTree`, to tear down any old state,
+    /// or throw out caches which should not be maintained across layout cycles.
+    func finishedLayout() {
```
n8chur, line 434: Nit:
```suggestion
    func cleanupAfterLayout() {
```
?

BlueprintUI/Tests/XCTestCaseAdditions.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+import XCTest
+
+extension XCTestCase {
+
+    /// Provides scoping within a single `func test...()` method, eg
+    /// if you want to test multiple cases / permutations with a label.
+    public func testcase(
```
n8chur, line 15: Why not just use `do {` with a comment?

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+    /// The signpost ref used when logging to `os_log`.
+    let signpostRef: AnyObject
+
+    /// The name of the owning tree. Useful for debugging.
+    let name: String
+
+    /// The element represented by this object.
+    /// This value is a reference type – the inner element value
+    /// will change after updates.
+    let element: ElementSnapshot
```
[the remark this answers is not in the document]
kyleve: We pass this through to the `LayoutResultNode`, which is a `struct`.

BlueprintUI/Tests/XCTestCaseAdditions.swift
kyleve: I have this in Market and Listable. It's a nicer way to define sub-cases vs the comment and IMO somewhat abuse of the `do {`.

BlueprintUI/Sources/Element/ElementState.swift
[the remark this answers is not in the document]
kyleve: Another option is to pass the whole `ElementState` through to the `LayoutResultNode` and avoid the inner type.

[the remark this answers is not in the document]
n8chur: Or just define the box on `LayoutResultNode`, right?

This seems fine, I was mostly just unclear on how this was used when I asked 👍.

BlueprintUI/Tests/XCTestCaseAdditions.swift
n8chur: IMO it's weird that the `String` is unused—it looks like it should do more than just create a scope.

I don't feel strongly, but I try to avoid stuff like this since it interrupts code review while you spin up on what this does, where just using the native patterns/tools is instantly obvious.

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+    let name: String
+
+    init(name: String) {
+        self.name = name
+    }
+
+    /// Updates, replaces, or removes the root node depending on the type of the new element.
+    func update(with element: Element?, in environment: Environment) {
+
+        func makeRoot(with element: Element) {
```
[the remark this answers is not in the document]
kyleve: This is going to change in https://github.com/square/Blueprint/pull/389, so not going to touch for now.

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -0,0 +1,597 @@
...
+            }
+        }
+    }
+}
+
+
+/// A reference type passed through to `LayoutResultNode` and `NativeViewNode`,
+/// so that even when returning cached layouts, we can get to the latest version of the element
+/// that their originating `ElementState` represents.
+final class ElementSnapshot {
```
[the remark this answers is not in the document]
kyleve: Yeah good point, the name ain't quite it. `LatestElement`? `UpdatingElement`? `LatestElementReference`?

BlueprintUI/Sources/Element/ComparableElement.swift
kyleve: Sure, I'll add an example to the protocol's func definition itself.

kyleve: Alternatively, we might just wanna remove this and then add it back once we find a real need where it matters in production – after thinking about this use case more (opaque builders that return elements); if the elements returned there are `ComparableElement`, _their_ layout will still be cached, so the upstream element that contains both known _and_ unknown element values can still benefit from caching.

BlueprintUI/Sources/Element/ElementState.swift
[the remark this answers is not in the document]
kyleve: I've moved this to `ElementSnapshot.LatestElement`.

BlueprintUI/Sources/Element/ElementState.swift
kyleve: I am going to leave this for now; since I think I can make both prepare and finish a single function call; will try it in https://github.com/square/Blueprint/pull/389.

BlueprintUI/Sources/Element/ElementState.swift
kyleve: I like appending `ifNeeded`, it will make testing the desired behavior a bit easier without having to change the shape of the caller

BlueprintUI/Sources/Element/ElementState.swift
kyleve: Sure, will do

BlueprintUI/Sources/Element/ElementState.swift
kyleve: Committing locally!

BlueprintUI/Sources/Element/ElementState.swift
kyleve: Done

BlueprintUI/Sources/Element/ElementState.swift
kyleve: Done

BlueprintUI/Sources/Element/ElementState.swift
kyleve: I'm just going to remove `wasUpdateEquivalent` for now, given it was only added for debugging. We could always bring it back if needed

PR #385, opened 2022-11-04
chore: [Blueprint] Advance minimum deployment target to iOS 14

nononoah:
### Summary

This PR advances Blueprint's minimum deployment target from iOS 12 to iOS 14. Along the way, it removes now defunct availability clauses and corrects deprecation warnings where they were observed.


BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -424,7 +424,7 @@ extension ScrollView {
             finalContentInset.bottom -= safeAreaInsets.bottom
         }
 
-        if #available(iOS 13, *) {
+        if #available(iOS 14, *) {
```
nononoah, line 427: [This radar](http://www.openradar.appspot.com/35866834) has been marked as resolved. It might be possible to remove this altogether, but this appears to be the behavior we're relying on at the moment so I did not wish to take that step as part of this PR.

SampleApp/SampleApp.xcodeproj/project.pbxproj
```diff
@@ -754,7 +754,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
```
nononoah, line 757: Note - we could eliminate the need to maintain this with a post install action (thereby reducing the number of sources of truth), but it seemed unnecessary to introduce as much in this PR.

BlueprintUI.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = SWIFT_VERSION
 
-  s.ios.deployment_target = '12.0'
+  s.ios.deployment_target = '14.0'
```
kyleve, line 16: Should we move this to version.rb?

BlueprintUICommonControls.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = SWIFT_VERSION
 
-  s.ios.deployment_target = '12.0'
+  s.ios.deployment_target = '14.0'
```
kyleve, line 16: Same Q; that way the version is only in one place.

kyleve (approved):

BlueprintUI/Tests/BuilderTests.swift
```diff
@@ -174,7 +174,7 @@ class BuilderTests: XCTestCase {
                 "2"
             }
 
-            if #available(iOS 11, *) {
+            if #available(iOS 14, *) {
```
n8chur, line 177: Why is it still useful to keep any of these `if #available(iOS 14, *)` checks if you cannot deploy on a version below iOS 14?

n8chur (approved):

BlueprintUI/Tests/BuilderTests.swift
kyleve: This is part of ensuring our function builder syntax for availability checks is working properly.

kyleve:
```
  /// If declared, this will be called on the partial result of an 'if
  /// #available' block to allow the result builder to erase type
  /// information.
  static func buildLimitedAvailability(_ component: Component) -> Component { ... }
```

n8chur: ohhh, got it.

PR #386, opened 2022-11-04
[Caching] ElementIdentifier caching

kdaker:
This PR adds optimizations to `ElementIdentifier` that are needed cross-render caching.
With [cross-render caching](https://docs.google.com/document/d/1ozD6XaZYZ2a8xRudJrZ-4m3tcH6d8N4bBdtFTjks4QM/edit#heading=h.si9yctdme18g) we will be using `ElementIdentifier` more frequently than we currently do which results in many unnecessary memory allocations for identifiers that are essentially the same.

We now cache `ElementIdentifier` by type and count. This should significantly reduce memory allocation of identifiers during layout passes. Key-based identifiers are never cached.

### DIFF
--- a/BlueprintUI/Sources/Internal/ElementIdentifier.swift
+++ b/BlueprintUI/Sources/Internal/ElementIdentifier.swift
@@ -4,13 +4,13 @@
  and update operations.
 
  The identifier has three parts:
- 1) The type of the underlying element, represented by `elementType`.
+ 1) **The type** of the underlying element, represented by `elementType`.
 
- 2) The key. This is an optional value provided by developers using Blueprint to further
+ 2) **The key.** This is an optional value provided by developers using Blueprint to further
  disambiguate identical elements within a hierarchy. This is optional, and is rarely provided.
  When it is provided, `elementType` + ` key` is used to disambiguate elements.
 
- 3) The occurrence count of that type of `elementType` + `key` in the hierarchy.
+ 3) **The occurrence count** of that type of `elementType` + `key` in the hierarchy.
  For example, if I have a hierarchy of [A, B, B] the counts will be [1, 1, 2] respectively.
 
  A fully constructed `ElementIdentifier` would look like this:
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -41,42 +41,82 @@
...
+    private let hash: Int
+    private static var cachedIdentifiers: [ObjectIdentifier: [Int: ElementIdentifier]] = [:]
 
-        self.elementType = ObjectIdentifier(elementType)
+    static func identifier(for element: Element, key: AnyHashable?, count: Int) -> ElementIdentifier {
+
+        let elementType = type(of: element)
+
+        guard key == nil else {
+            let result = ElementIdentifier(elementType: elementType, key: key, count: count)
```
kyleve, line 60: (My fault again sorry) can just be: `return ElementIdentifier(elementType: elementType, key: key, count: count)`

kyleve (approved):

BlueprintUI/Tests/ElementIdentifierTests.swift
```diff
@@ -9,58 +9,82 @@ class ElementIdentifierTests: XCTestCase {
...
+        set.insert(id3.objectIdentifier)
+        XCTAssertFalse(set.contains(id4.objectIdentifier))
+        set.insert(id4.objectIdentifier)
+        XCTAssertTrue(set.contains(id5.objectIdentifier))
+        set.insert(id5.objectIdentifier)
+        XCTAssertTrue(set.contains(id6.objectIdentifier))
+
+        XCTAssertEqual(id1.objectIdentifier, id5.objectIdentifier)
+        XCTAssertEqual(id2.objectIdentifier, id6.objectIdentifier)
+        XCTAssertNotEqual(id5.objectIdentifier, id6.objectIdentifier)
```
n8chur, line 60: Should we also check hashes now that `Hashable` and `Equatable` use different codepaths?

kyleve:
Going to merge this one into the feature branch to clean up some later diffs that depend on it in our PR tree – happy to address any follow ups!

PR #388, opened 2022-11-04
[Caching] Add test fixtures for element state testing

kyleve:


BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -542,6 +632,68 @@ extension Dictionary {
...
+
+    func treeDidPerformCachedLayout(
+        _ layout: ElementState.CachedLayout,
+        for state: ElementState
+    )
+}
+
+
+extension Optional where Wrapped == ElementStateTreeDelegate {
+    func perform(_ block: (Wrapped) -> Void) {
```
n8chur, line 688: Could we name this function in a way where it's clear that it's only performed in `DEBUG` mode? `ifDebugPerform`?

n8chur (commented):
Should this PR be targetting `kareem/elementidentifier-changes` as the base branch?

BlueprintUI/Sources/Element/ElementState.swift
kyleve: How about just `ifDebug`?

kyleve: This is done in a future branch

kdaker (approved):

PR #389, opened 2022-11-05
[Caching] Update ElementContent to consume the ElementStateTree

kyleve:


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,146 @@ public struct ElementContent {
...
+    /// you should pass through the `LayoutContext` provided to you to ensure the measured elements
+    /// downstream have full access to their measurement caches, environment, etc.
+    /// ```
+    /// public var content: ElementContent {
+    ///     ElementContent { constraint, context -> CGSize in
+    ///         self.wrapped.detachedMeasure(in: constraint, with: context)
+    ///     }
+    /// }
+    /// ```
+    public func detachedMeasure(in constraint: SizeConstraint, with environment: Environment) -> CGSize {
```
kyleve, line 102: Remove this for now.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -391,6 +392,44 @@ public final class BlueprintView: UIView {
...
+        rootState.root?.prepareForLayout()
+
+        defer {
+            self.rootState.root?.finishedLayout()
+        }
+
+        states.update(with: element, in: environment)
+
+        return LayoutResultNode(
+            root: ElementSnapshot(element),
```
kyleve, line 425: Should be `states.root!.element`

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,76 @@ public struct ElementContent {
         storage = builder
     }
 
-    // MARK: Measurement & Children
+    fileprivate init(storage: ContentStorage) {
```
kyleve, line 26: Double check if these are needed.

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -10,10 +10,23 @@ extension Element {
     ///
     /// - Returns: A layout result
     func layout(layoutAttributes: LayoutAttributes, environment: Environment) -> LayoutResultNode {
```
kyleve, line 12: Hmm while we're in here, can we move this to the test module? I think we can.

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -23,42 +36,58 @@ extension Element {
...
+        children: [LayoutResultNode]
     ) {
         self.element = element
+        self.identifier = identifier
         self.layoutAttributes = layoutAttributes
         self.environment = environment
+        self.state = state
         self.children = children
+
+        precondition(type(of: element.value) == type(of: state.element.value))
```
kyleve, line 68: Convert this (and others) to an `assert` to avoid paying a cost in release.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,75 @@ public struct ElementContent {
...
-            in: constraint,
-            environment: environment,
-            cache: CacheFactory.makeCache(name: "ElementContent")
-        )
-    }
 
-    func measure(in constraint: SizeConstraint, environment: Environment, cache: CacheTree) -> CGSize {
-        storage.measure(in: constraint, environment: environment, cache: cache)
-    }
+        let element = MeasurementElement(content: self)
```
n8chur, line 68: Why is this created outside of the `autoreleasepool`?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -391,6 +409,43 @@ public final class BlueprintView: UIView {
...
+        state: ElementStateTree
+    ) -> LayoutResultNode {
+
+        rootState.root?.prepareForLayout()
+
+        defer {
+            self.rootState.root?.finishedLayout()
+        }
+
+        state.update(with: element, in: environment)
```
n8chur, line 439: Could we break this up into two functions, one where `element` is non-optional and returns a non-optional `root`, and another that takes an optional so we can avoid the force unwrapping?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,75 @@ public struct ElementContent {
...
+    fileprivate init(storage: ContentStorage) {
+        self.storage = storage
+    }
+
+    func measure(
+        in constraint: SizeConstraint,
+        with environment: Environment,
+        state: ElementState
+    ) -> CGSize {
+        autoreleasepool {
```
n8chur, line 35: Are these `autoreleasepool`s exclusively to avoid memory bloating? Do they cause any negative performance impact?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -356,87 +387,119 @@ private struct EnvironmentAdaptingStorage: ContentStorage {
...
             )
         }
     }
 
     private func adapted(environment: Environment) -> Environment {
         var environment = environment
+
+        environment.readNotificationsEnabled = false
         adapter(&environment)
+        environment.readNotificationsEnabled = true
```
n8chur, line 443: Micronit (fine leaving as is): Should we add a function to `Environment` that allows you to customize it without sending read notifications so we can keep this property/behavior `private`?

Something like:
```
private func adapted(environment: Environment) -> Environment {
    environment.adaptedWithoutReadNotifications(adapter)
}
```

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -41,55 +41,87 @@
 
  You will note that the identifiers remain stable, which ultimately ensures that views are reused.
  */
-struct ElementIdentifier: Hashable, CustomDebugStringConvertible {
+final class ElementIdentifier: Hashable, CustomDebugStringConvertible {
```
n8chur, line 44: Looks like the stack might be out of order or a need a rebase, as these changes are in #386.

kyleve: This is expected, we have a tree of PRs, not a single stack, so I had to pull this in to avoid more thrash back and forth in this PR then another one. We should be merging the ID one soon so this will go away

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: That's a good idea!

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: Yeah, avoid big spikes. No perf impact that I can tell since it'd just be released next runloop cycle in aggregate anyway.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: I went back and forth on this vs the current shape, so happy to bring this one back

BlueprintUI/Sources/Internal/ElementIdentifier.swift
n8chur: Ah, got it.

Why not just stack them on top of each other?

kyleve: So Kareem and I were able to independently work on pulling in separate pieces last week

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: Actually, this is a pain because it leads to ambiguous method resolution, having two methods with the same name, where one has `Element?` and another has `Element`; and giving them different names is a bit odd... Let's see if making the optional one just a `remove` works..

kyleve: That seems to work pretty well

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: We might also just kill this entirely; it's a pretty tiny micro-optimization.

kyleve: I'm just going to kill this for now.

BlueprintUI/Sources/Internal/LayoutResultNode.swift
kyleve: Yep

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: Probably shouldn't be, moving in

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
n8chur: Would `@_disfavoredOverload` help?

kdaker (approved):

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -1,64 +1,58 @@
...
-            root: self,
-            layoutAttributes: layoutAttributes,
-            environment: environment
-        )
-    }
-
-}
-
 /// Represents a tree of elements with complete layout attributes
 struct LayoutResultNode {
```
kyleve, line 4: Consider turning this whole thing into a reference type?

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -145,7 +145,10 @@ class BlueprintViewTests: XCTestCase {
             ]
         )
 
+        blueprintView.layoutIfNeeded()
+
         tags = [1, 2]
+
         for index in blueprintView.currentNativeViewControllers.indices {
```
kyleve, line 152: Calling `currentNativeViewControllers` implicitly lays out the hierarchy, which is a bit non-obvious; so I added the explicit `layoutIfNeeded` calls.

BlueprintUI/Tests/Element+Testing.swift
```diff
@@ -11,6 +12,32 @@ extension Element {
...
     }
+
+    /// Build a fully laid out element tree with complete layout attributes
+    /// for each element.
+    ///
+    /// - Parameter layoutAttributes: The layout attributes to assign to the
+    ///   root element.
+    ///
+    /// - Returns: A layout result
+    func layout(layoutAttributes: LayoutAttributes, environment: Environment) -> LayoutResultNode {
```
kyleve, line 23: This moved from production code into the tests.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -23,42 +23,94 @@ public struct ElementContent {
...
-            in: constraint,
-            environment: environment,
-            cache: CacheFactory.makeCache(name: "ElementContent")
-        )
-    }
+        autoreleasepool {
+            let element = MeasurementElement(content: self)
+            let root = ElementStateTree(name: "ElementContent.measure")
+
+            let (_, size) = root.performUpdate(with: element, in: environment) { state in
```
kdaker, line 89: Given `performUpdate` calls `finishedLayout`, we now call that on every measurement. It's a recursive call that has to have some overhead.

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -55,39 +68,62 @@ final class ElementStateTree {
...
                 root.update(with: element, in: environment, identifier: root.identifier)
 
-                delegate.perform {
+                let output = updates(root)
+
+                delegate.ifDebug {
                     $0.tree(self, didUpdateRootState: root)
                 }
+
+                root.finishedLayout()
```
kdaker, line 84: Related to my other comment, this recursive call gets called on all measurements. Can we optimize this to only run in certain conditions where we know we need to clean up? Looking at what it calls, we don't always skip when `wasVisited` is true, meaning it's recursive all the time.

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: This is for the "detached" measurement method; not for the regular measurement tree – eg where you're just invoking `element.content.measure` somewhere.

PR #390, opened 2022-11-05
[Caching] Add some tests for ElementState

kdaker:
* Adds some tests for `ElementState` methods
* Adds assertions for `ElementStateTreeDelegate` calls
* Adds `ContainerElement` for testing `Element` hierarchies in later tests

### DIFF
--- a/BlueprintUI/Tests/ElementStateTests.swift
+++ b/BlueprintUI/Tests/ElementStateTests.swift
@@ -14,6 +14,7 @@ import XCTest
 class ElementStateTreeTests: XCTestCase {
 
     func test_update() throws {
+        let delegate = TestDelegate()
 
         let tree = ElementStateTree(name: "Testing")
         let element1 = Element1(text: "1")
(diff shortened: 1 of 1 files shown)


BlueprintUI/Tests/ElementStateTests.swift
```diff
@@ -13,6 +13,16 @@ import XCTest
 
 class ElementStateTreeTests: XCTestCase {
 
+    fileprivate var delegate: TestDelegate! = nil
```
kyleve, line 16: Fwiw, rather than assigning this in setUp and tearDown, I think it's better to just create it within each test – that avoids more global state, and will eventually make it easier to have multiple cases in a single test method – this is generally the pattern we recommend in SPOS as well – have a fixture or object you create in each test method, vs having instance vars and setting them in setup and teardown.

kyleve (approved):
Looks good to verify things are working right now! Per discussion, I think we should likely remove these when we've got the fuller suite running, since these are a bit of a synthetic test at that point.

PR #391, opened 2022-11-10
[SPL] SwiftUI-mimicking layout

watt:
This is the work myself and @n8chur have done so far in pairing. This is sort of a "checkpoint" merge, so that we can make more iterations on this while also bringing in the previous prototype on top of it.


BlueprintUI/Sources/Element/Element+Debug.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+
+extension Element {
+
+    @_spi(Debugging)
+    public func debugScope(_ scope: String) -> Element {
```
n8chur, line 7: We should probably come back to this at some point—I don't know if we gave it enough though (maybe it's fine though?).

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -45,8 +45,19 @@ public struct ElementContent {
...
-        // TODO: switch on layoutMode
-        storage.measure(in: constraint, environment: environment, cache: cache)
+        switch layoutMode {
+        case .standard:
+            return storage.measure(in: constraint, environment: environment, cache: cache)
+        case .singlePass:
+            return storage.sizeThatFits(
+                proposal: ProposedViewSize(constraint),
+                context: .init(
+                    // TODO: Hoist upward?
```
n8chur, line 55: Should we do this before we get too deep into performance optimizations? Might this have a significant impact?

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -18,8 +18,34 @@ extension Element {
...
+        switch layoutMode {
+        case .standard:
+            return layout(layoutAttributes: LayoutAttributes(frame: frame), environment: environment)
+        case .singlePass:
+            return singlePassLayout(
+                proposal: .init(frame.size),
+                context: SPLayoutContext(
+                    attributes: LayoutAttributes(frame: frame),
+                    environment: environment,
+                    // TODO: Hoist up?
```
n8chur, line 30: Same here regarding potentially doing this before we get too deep into performance optimizations.

n8chur (approved):

BlueprintUI/Sources/Element/ElementContent.swift
watt: Yeah, I'll make a note in our planning doc

PR #394, opened 2022-11-15
Adding custom actions to accessibility element

RoyalPineapple:


BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -3,6 +3,24 @@ import UIKit
...
 
+    /// Used to provide addtional functionality to assistive technogies beyond your accessible UI.
+    public struct CustomAction {
+        public typealias OnActivation = (String) -> Bool
+
+        /// A localized name that discribes the action.
+        public var name: String
+        /// An image representing the action to be shown with some assistive technologies such as Switch Control.
+        public var image: UIImage?
+        /// A Callback for when the action is activated. The action's name will be passed as a parameter. This should return a `bool` indicating success or failure of the action.
```
n8chur, line 14: Why do we need to pass the name in here? Isn't it declared alongside a name?

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -3,6 +3,24 @@ import UIKit
...
+        public var onActivation: OnActivation
+
+        public init(name: String, Image: UIImage? = nil, onActivation: @escaping OnActivation) {
+            self.name = name
+            image = Image
```
n8chur, line 19:
```suggestion
        public init(name: String, image: UIImage? = nil, onActivation: @escaping OnActivation) {
            self.name = name
            self.image = image
```

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: I guess we don't need to, Apple's implementation passes the whole object back, so i was just extracting the relevant bits from that but you're right, it can pretty much be ignored.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: 👍🏻

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -3,6 +3,24 @@ import UIKit
 
 public struct AccessibilityElement: Element {
 
+    /// Used to provide addtional functionality to assistive technogies beyond your accessible UI.
+    public struct CustomAction {
+        public typealias OnActivation = () -> Bool
```
ldstreet, line 8: Not sure on all the use cases, but should we input a `CustomAction` here to mirror `UIAccessibilityCustomAction.Handler`?

RoyalPineapple: It would more closely mirror apple's implementation, but all it would gain you is the name and the callback itself. You'll have the name when creating the action and I can't see why having access to the callback within itself would be useful unless you wanna do some weird recursive something that would probably be unwise anyway,

ldstreet (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
ldstreet: I'm good with omitting. We can always add later if there is a good use case.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -45,6 +63,9 @@ public struct AccessibilityElement: Element {
     /// See [Accessibility Activate Documentation](https://developer.apple.com/documentation/objectivec/nsobject/1615165-accessibilityactivate) for further information.
     public var accessibilityActivate: (() -> Bool)? = nil
 
+    /// The array contains one or more `Custom Action`s, defining additional supported actions. Assistive technologies, such as VoiceOver, will display your custom actions to the user at appropriate times.
+    public var customActions: [CustomAction]? = nil
```
kyleve, line 67: We need to add this to the `public func accessibilityElement(` below.

RoyalPineapple: I was imagining that this would be such a rare occurrence you'd be fine using the configure block in the  big initializer, but if you don't think adding yet another property to the convenience method is overkill I've got no pushback.

ldstreet:
Do we also need to add this to `Label`? If we were to layer this onto a Label with markdown links then we'd lose all the rotor accessibility. Ideally we'd be able to add custom actions directly on the label, no?

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: I think adding it and giving it an optional `= nil` is fine, but we tend to avoid folks using the `init` based builders to avoid the "pyramid of doom" of intending. Another option is to throw a `configure` block on the `.accessibilityElement` builder.

n8chur: +1 to adding this to the `public func` modifier.

BlueprintUICommonControls/Sources/AccessibilityElement.swift
```diff
@@ -3,6 +3,24 @@ import UIKit
 
 public struct AccessibilityElement: Element {
 
+    /// Used to provide addtional functionality to assistive technogies beyond your accessible UI.
```
vng-sq, line 6: Typos: additional, technologies

vng-sq (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve, line 67: Let's just keep this as an `Array`, empty would mean there's no actions – right?

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityElement.swift
RoyalPineapple: 👍🏻  thanks!

BlueprintUICommonControls/Sources/AccessibilityElement.swift
kyleve: It's just an extra state that doesn't really mean anything – eg what is the difference between `nil` and `[]`?

RoyalPineapple: 👍🏻

PR #395, opened 2022-11-15
[Caching] Simplify how storage and layout works for singular child elements

kyleve:
An optimization we found while working on caching, this improvements performance for basic single child elements with a passthrough layout.


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -599,6 +639,7 @@ fileprivate struct MeasurableLayout: Layout {
     }
 }
 
+
```
nononoah, line 642: Intentional add?

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -570,18 +622,6 @@ fileprivate struct SingleChildLayoutHost: Layout {
...
-
-    func measure(in constraint: SizeConstraint, child: Measurable) -> CGSize {
-        child.measure(in: constraint)
-    }
-
-    func layout(size: CGSize, child: Measurable) -> LayoutAttributes {
-        LayoutAttributes(size: size)
-    }
-
-}
```
nononoah, line 584: I wonder if some portion of this optimization - which, I think, improves on `SingleChildLayout` by introducing caching - could be introduced to all `SingleChildLayout` adopters, perhaps by updating `SingleChildLayoutHost`?

nononoah (approved):

kdaker (approved):

BlueprintUI/Sources/Element/ElementContent.swift
kyleve: The trick here is that we can skip the layout entirely; because the child is always exactly the same size as the parent – I don't think we can do that for other non-passthrough layouts

kyleve: Though perhaps: This could take an optional `Layout`, which would then allow us to alternatively pass through vs perform the layout. Would be a good follow up to play with

kyleve:
Needs a merge w/ the feature branch to get tests passing

PR #396, opened 2022-11-16
[Caching] Update PerformancePlayground w/ tests

kyleve:


BlueprintUICommonControls/Tests/Sources/PerformancePlayground.swift
```diff
@@ -224,17 +284,15 @@ class PerformancePlayground: XCTestCase {
...
+            let iterationEnd = Date()
+            let duration = iterationEnd.timeIntervalSince(iterationStart)
 
             iterations += 1
 
-            if Date().timeIntervalSince(lastUpdateDate) >= 1 {
-                lastUpdateDate = Date()
-                print("Continuing Test: \(iterations) Iterations...")
-            }
+            print("Iteration: \(iterations), Duration : \(duration)")
```
nononoah, line 295: 🐴 

I would suggest a rewrite of this method to better scope measured time:

```swift
func determineAverage(for seconds: TimeInterval, using block: () -> Void) {
    var duration: TimeInterval = 0
    var measurements: [TimeInterval] = []
    
    while duration < seconds {
        let start = Date()
        block()
        let end = Date()
        
        let elapsed = end.timeIntervalSince(start)
        duration += elapsed
        measurements.append(elapsed)

        print("Iteration: \(measurements.count), Duration : \(elapsed)")
    }

    let average = duration / TimeInterval(measurements.count)
    print("Iterations: \(measurements.count), Average Time: \(average)")
}
```

This approach

- excludes time not measuring the block
- sets us up to print other stats if we want, such as median and standard deviation

PR #397, opened 2022-11-16
[Caching] Add new storage type elements that create child BlueprintViews

kdaker:
Adds new init for `ElementContent` that takes and measures a child element. It uses a new storage type that only caches the measurement for the child element but does not return any layout nodes.

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -155,6 +155,18 @@ extension ElementContent {
         ) -> [LayoutResultNode] {
             []
         }
+
+        func forEachElement(
+            in size: CGSize,
+            with environment: Environment,
+            children childNodes: [LayoutResultNode],
+            state: ElementState,
+            forEach: (ElementState, Element, LayoutResultNode) -> Void
+        ) {
+            let childState = state.childState(for: element, in: environment, with: .identifierFor(singleChild: element))
+
+            forEach(childState, element, childNodes[0])
+        }
     }
 }
 
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -54,6 +54,24 @@ public struct ElementContent {
             )
         }
     }
+
+    func forEachElement(
+        in size: CGSize,
+        with environment: Environment,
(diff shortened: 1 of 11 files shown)


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -91,6 +91,53 @@ extension ElementContent {
...
+    /// Creates a new `ElementContent` which uses the provided element to measure its
+    /// size, but does not place the element as a child in the final, laid out hierarchy.
+    ///
+    /// This is useful if you are placing the element in a nested `BlueprintView`, for example (eg
+    /// to create a stateful element) and just need this element to be correctly sized.
+    public init(byMeasuring element: Element) {
+        storage = MeasureNestedElementStorage(element: element)
+    }
+
+    struct MeasureNestedElementStorage: ContentStorage {
```
kyleve, line 105: Nit: private?

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -235,12 +251,14 @@ final class ElementState {
         identifier: ElementIdentifier,
         element: Element,
         signpostRef: AnyObject,
-        name: String
+        name: String,
+        kind: Kind = .regular
```
kyleve, line 255: IMO given this is an internal-only method, I think requiring folks to pass a value is best, vs them forgetting and getting it wrong

kyleve (approved):

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -111,6 +111,65 @@ extension ElementContent {
...
+
+        func performLayout(
+            in size: CGSize,
+            with environment: Environment,
+            state: ElementState
+        ) -> [LayoutResultNode] {
+            []
+        }
+
+        func forEachElement(
```
n8chur, line 159: How is this used?

kyleve: We use this to enumerate the element tree, to make sure that the newest element makes it into the live state; even if we return a cached layout, eg: https://github.com/square/Blueprint/pull/404/files#diff-87a9ff4712c4ac0a865f6587fa32cecfb76335e0767e569de804727956baffc4L489

PR #401, opened 2022-11-18
Add perf test for incrementally changing state

kdaker:


BlueprintUICommonControls/Tests/Sources/PerformancePlayground.swift
```diff
@@ -219,6 +219,61 @@ class PerformancePlayground: XCTestCase {
         }
     }
 
+    func test_increment_state_changes() {
```
kyleve, line 222: I think it's worth waiting for #389 to land, and then re-using `LabeledBox` for an apples to apples comparison with test_deep_and_wide_grid.

PR #406, opened 2022-12-09
[SPL] Strict SPL layout mode

watt:
This is an implementation of "strict" SPL, based on the prototype in #328.

### DIFF
--- a/BlueprintUI/Sources/Internal/CacheTree.swift
+++ b/BlueprintUI/Sources/Internal/CacheTree.swift
@@ -23,10 +23,25 @@ extension CacheTree {
     /// Convenience method to get a cached size, or compute and store one if it is not in the cache.
     func get(_ constraint: SizeConstraint, orStore calculation: (SizeConstraint) -> CGSize) -> CGSize {
 
+//        print("measure \(name)")
+//        if (name == "Row/Column.0/Spacer.0") {
+//            print("xxx")
+//        }
         if let size = self[constraint] {
+            Logger.logCacheHit(object: self.signpostRef, description: name, constraint: constraint)
             return size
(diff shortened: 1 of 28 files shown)


BlueprintUI/Sources/Internal/CacheTree.swift
```diff
@@ -23,11 +23,26 @@ extension CacheTree {
     /// Convenience method to get a cached size, or compute and store one if it is not in the cache.
     func get(_ constraint: SizeConstraint, orStore calculation: (SizeConstraint) -> CGSize) -> CGSize {
 
+//        print("measure \(name)")
+//        if (name == "Row/Column.0/Spacer.0") {
+//            print("xxx")
+//        }
```
n8chur, line 29:
```suggestion
```

BlueprintUI/Sources/Layout/LayoutWriter.swift
```diff
@@ -285,6 +285,24 @@ extension LayoutWriter {
                     )
                 }
             }
+            
+            func layout(in context: StrictLayoutContext, children: [StrictLayoutChild]) -> StrictLayoutAttributes {
+//                let options = StrictLayoutOptions(mode: .init(horizontal: .fill, vertical: .fill))
```
n8chur, line 290: Do we need to add a todo here and/or remove this commented code?

n8chur (approved):

BlueprintUI/Sources/Layout/LayoutWriter.swift
watt: Will remove — I think this cruft came along with the original prototype.

PR #407, opened 2022-12-12
Extension Safe Methods: LayoutDirectionKey

lickel:
This was accessing UIApplication.main, which is unsafe.
The _modern_ way to access this would be via a UITraitCollection.

### DIFF
--- a/BlueprintUI/Sources/Environment/Keys/LayoutDirectionKey.swift
+++ b/BlueprintUI/Sources/Environment/Keys/LayoutDirectionKey.swift
@@ -15,12 +15,24 @@ extension Environment {
                 self = .leftToRight
             }
         }
+
+        init(_ direction: UITraitEnvironmentLayoutDirection) {
+            switch direction {
+            case .leftToRight:
+                self = .leftToRight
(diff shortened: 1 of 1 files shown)


BlueprintUI/Sources/Environment/Keys/LayoutDirectionKey.swift
```diff
@@ -19,7 +19,8 @@ extension Environment {
 
     private enum LayoutDirectionKey: EnvironmentKey {
         static var defaultValue: LayoutDirection {
-            LayoutDirection(UIApplication.shared.userInterfaceLayoutDirection)
+            // This will be updated in BlueprintView.makeEnvironment()
+            .leftToRight
```
kyleve, line 23: Could we use `.current` (https://developer.apple.com/documentation/uikit/uitraitcollection/3238080-current) here, or is that unsafe?

lickel: Good call!

kyleve (approved):

PR #408, opened 2022-12-12
Extension Safe Methods: AttributedLabel

lickel:
This tricks Xcode into thinking AttributedLabel is "extension safe".
It also makes `URLHandlerEnvironmentKey.defaultValue` settable.

In an extension target you could then make it _actually_ safe by providing your own default.

In a significant version change, I would like to change the default to be a no-op.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
@@ -29,7 +29,16 @@ class DefaultURLHandler: NullURLHandler {
 }
 
 public struct URLHandlerEnvironmentKey: EnvironmentKey {
-    public static var defaultValue: URLHandler = DefaultURLHandler()
+    public static let defaultValue: URLHandler = {
+        // This is our best guess for "is this executable an extension?"
+        if let _ = Bundle.main.infoDictionary["NSExtension"] {
+            return NullURLHandler()
+        } else if Bundle.main.bundlePath.hasSuffix(".appex") {
+            return NullURLHandler()
+        } else {
+            return DefaultURLHandler()
+        }
+    }()
 }
 
 extension Environment {
--- a/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
@@ -17,6 +17,7 @@ public protocol URLHandler {
     func onTap(url: URL)
 }
 
+@available(iOSApplicationExtension, unavailable, message: "This cannot be used in application extensions")
 struct DefaultURLHandler: URLHandler {
     func onTap(url: URL) {
         UIApplication.shared.open(url, options: [:], completionHandler: nil)
@@ -24,12 +25,12 @@ struct DefaultURLHandler: URLHandler {
 }
 
 public struct URLHandlerEnvironmentKey: EnvironmentKey {
-    public static let defaultValue: URLHandler = DefaultURLHandler()
+    public static let defaultValue: URLHandler? = nil
 }
 
 extension Environment {
     /// The link handler to use to open links tapped in an `AttributedLabel`.
-    public var urlHandler: URLHandler {
+    public var urlHandler: URLHandler? {
         get { self[URLHandlerEnvironmentKey.self] }
         set { self[URLHandlerEnvironmentKey.self] = newValue }
     }
--- a/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
@@ -17,14 +17,19 @@ public protocol URLHandler {
     func onTap(url: URL)
 }
 
-struct DefaultURLHandler: URLHandler {
-    func onTap(url: URL) {
+class NullURLHandler: URLHandler {
+    func onTap(url: URL) {}
+}
+
+class DefaultURLHandler: NullURLHandler {
+    @available(iOSApplicationExtension, unavailable)
+    override func onTap(url: URL) {
         UIApplication.shared.open(url, options: [:], completionHandler: nil)
     }
 }
 
 public struct URLHandlerEnvironmentKey: EnvironmentKey {
-    public static let defaultValue: URLHandler = DefaultURLHandler()
+    public static var defaultValue: URLHandler = DefaultURLHandler()
 }
 
 extension Environment {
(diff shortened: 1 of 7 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -96,8 +96,25 @@ public struct AttributedLabel: Element, Hashable {
         .foregroundColor: UIColor.systemBlue.withAlphaComponent(0.3),
     ]
 
+    /// This is the fallback value used when the environment has no URL handler
+    private var defaultURLHandler: URLHandler
+
+    @available(iOSApplicationExtension, unavailable, message: "This cannot be used in application extensions")
     public init(attributedText: NSAttributedString, configure: (inout Self) -> Void = { _ in }) {
```
kyleve, line 103: I assume we can't then do something like: 

```
    @available(iOSApplicationExtension, available)
    public init(attributedText: NSAttributedString, configure: (inout Self) -> Void = { _ in }) {
        self.init(
            attributedText: attributedText,
            defaultURLHandler: AnExtensionSafeURLHandler(),
            configure: configure
        )
    }
```

And avoid the new init?

n8chur: I _think_ you'd need to inject a `NSExtensionContext` somehow in order for that to work. Is there a way to get the `NSExtensionContext` without piping it through?

n8chur (commented):
I responded in a Slack thread about this but also posting here:

I don't think we can go down this path because then anything that uses an `AttributedLabel` would need a separate initializer in Market. It would start with `MarketLabel`, and then all components that compose a `MarketLabel`, etc.

BlueprintUICommonControls/Sources/AttributedLabel.swift
lickel: Best I can tell, the only way to get an `NSExtensionContext` is via a UIViewController.
Further, that API is heavily restricted such that you can only open your host app.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -96,8 +96,25 @@ public struct AttributedLabel: Element, Hashable {
         .foregroundColor: UIColor.systemBlue.withAlphaComponent(0.3),
     ]
 
+    /// This is the fallback value used when the environment has no URL handler
+    private var defaultURLHandler: URLHandler
```
lickel, line 100: - [ ] This breaks Equatable/Hashable

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -96,8 +96,25 @@ public struct AttributedLabel: Element, Hashable {
...
+    @available(iOSApplicationExtension, unavailable, message: "This cannot be used in application extensions")
     public init(attributedText: NSAttributedString, configure: (inout Self) -> Void = { _ in }) {
+        self.init(
+            attributedText: attributedText,
+            defaultURLHandler: DefaultURLHandler(),
+            configure: configure
+        )
+    }
+
+    public init(
```
lickel, line 111: - [ ] Label would need this same init pattern

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
```diff
@@ -17,19 +17,20 @@ public protocol URLHandler {
     func onTap(url: URL)
 }
 
+@available(iOSApplicationExtension, unavailable, message: "This cannot be used in application extensions")
```
watt, line 20: Does this work?
```swift
private class NullURLHandler: URLHandler {
    func onTap(url: URL) { }
}

private final class DefaultURLHandler: NullURLHandler {
    @available(iOSApplicationExtension, unavailable)
    override func onTap(url: URL) {
        UIApplication.shared.open(url, options: [:], completionHandler: nil)
    }
}
```

lickel:
Cross posting:

Between Equatable and Label, I’m guessing the init() world is a no go…

Different build flavors would definitely work, but would ultimately result in duplicate symbols.

We could make a façade non-extension-safe module that depends upon something like ObjC +load to set URLHandlerEnvironmentKey.defaultValue
Alternatively we could make URLHandlerEnvironmentKey.defaultValue a public settable method and/or add a BlueprintUI.Environment.configure(application) and require API consumers to explicitly set the default (a breaking change)

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
lickel: Assuming that the compiler allowed it (I don't think so), it would still actually trigger the URL in a share extension ~> App Store rejection

lickel: OK this does actually work hah.
There's just the caveat that you should _always_ provide a url handler in an extension if you expect a URL

n8chur: Nice!

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
```diff
@@ -17,14 +17,19 @@ public protocol URLHandler {
...
+class DefaultURLHandler: NullURLHandler {
+    @available(iOSApplicationExtension, unavailable)
+    override func onTap(url: URL) {
         UIApplication.shared.open(url, options: [:], completionHandler: nil)
     }
 }
 
 public struct URLHandlerEnvironmentKey: EnvironmentKey {
-    public static let defaultValue: URLHandler = DefaultURLHandler()
+    public static var defaultValue: URLHandler = DefaultURLHandler()
```
lickel, line 32: In a future version, we could change the default to be a no-op

n8chur (approved):
Thanks for working through this!

- [x] 2 approvals

kyleve (approved):
Nice!

BlueprintUICommonControls/Sources/AttributedLabel+Environment.swift
watt: Hmm, I don't feel great about having a mutable global like this — I think if you need to have clickable links in an app extension context, it should be on you to accomplish that through normal environment means, i.e. adding an  `.adaptedEnvironment()` modifier.

kyleve: Ah shit, I missed this was globally mutable like this – it does seem easy to unintentionally break. I agree with Andrew's take here.

watt: I spoke to Adam on Slack and suggested we add a runtime check here (by [inspecting the main bundle](https://stackoverflow.com/a/29322061)), and do something like:

```swift
public static let defaultValue: URLHandler = runningInAppContext ? NullURLHandler() ? DefaultURLHandler()
```

watt (approved):
approved pending the `let defaultValue` with runtime check described above 👍

PR #409, opened 2022-12-13
Extension Safe Methods: Add Extension-Safe flag to podspecs

lickel:
This marks these targets as explicitly extension safe.

Depends upon https://github.com/square/Blueprint/pull/408

### DIFF
--- a/BlueprintUI.podspec
+++ b/BlueprintUI.podspec
@@ -19,6 +19,10 @@ Pod::Spec.new do |s|
 
   s.weak_framework = 'SwiftUI'
 
+  s.pod_target_xcconfig = {
+    'APPLICATION_EXTENSION_API_ONLY' => 'YES',
+  }
+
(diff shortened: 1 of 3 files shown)


BlueprintUI.podspec
```diff
@@ -19,6 +19,10 @@ Pod::Spec.new do |s|
 
   s.weak_framework = 'SwiftUI'
 
+  s.pod_target_xcconfig = {
+    'APPLICATION_EXTENSION_API_ONLY' => 'YES',
```
lickel, line 23: As far as I can tell, there's no equivalent for SPM. 
It's supposed to inherit this flag from its dependencies.

lickel: The _closest_ is an unsafe linker flag, as documented [here](https://danielsaidi.com/blog/2022/05/18/how-to-suppress-linking-warning)

PR #412, opened 2023-01-06
Cross-layout caching feature flag

kdaker:
Adds a flag to control cross-layout caching. The flag is a static member on BlueprintView. When turned on, all blueprint view instances will use the flag.

When the flag is off:
* We reset measurement caches and measurement caching behaves the same as it did with `CacheTree`.
* We do not cache layout nodes.

The feature flag is off by default. Because of this we turn it on for all tests that use `BlueprintView`.

Below is a comparison of performance tests between `main` and this branch with the flag turned off. Performance is not exact but close.

Test | Feature flag off w/ caching (seconds) | Main (seconds)
-- | -- | --
test_deep_and_wide_grid | 1.446 avg | 1.657 avg
test_repeated_layouts | 0.146 avg | 0.141 avg
test_row_with_one_flexible_element | 0.080 avg | 0.077 avg
test_deep_element_hierarchy | 0.254 avg | 0.245 avg
test_deep_stacks | 7.646 total | 8.136 total
test_incremental_state_change | 0.046 avg | 0.055 avg

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -27,9 +27,9 @@ import UIKit
 /// ```
 public final class BlueprintView: UIView {
     
-    public static var useCrossLayoutCaching: Bool = false {
+    public static var enableCrossLayoutCaching: Bool = false {
         didSet {
-            ElementState.useCaching = useCrossLayoutCaching
+            ElementState.enableCrossLayoutCaching = enableCrossLayoutCaching
         }
     }
 
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -26,7 +26,7 @@ import UIKit
 /// }
 /// ```
 public final class BlueprintView: UIView {
-    
+
     public static var enableCrossLayoutCaching: Bool = false {
         didSet {
             ElementState.enableCrossLayoutCaching = enableCrossLayoutCaching
@@ -427,7 +427,7 @@ public final class BlueprintView: UIView {
         rootFrame: CGRect
     ) -> [(path: ElementPath, node: NativeViewNode)] {
 
-        if let element = self.element {
+        if let element = element {
             let (_, node) = rootState.performUpdate(with: element, in: environment) { state in
                 LayoutResultNode(
                     identifier: .identifierFor(singleChild: element),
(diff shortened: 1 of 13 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -420,7 +427,7 @@ public final class BlueprintView: UIView {
         rootFrame: CGRect
     ) -> [(path: ElementPath, node: NativeViewNode)] {
 
-        if let element = self.element {
+        if let element = element {
```
kdaker, line 430: linter fix

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -460,7 +467,7 @@ public final class BlueprintView: UIView {
             }
         }()
 
-        var environment = inherited.merged(prioritizing: self.environment)
+        var environment = inherited.merged(prioritizing: environment)
```
kdaker, line 470: linter fix

BlueprintUI/Tests/BlueprintViewTests.swift
```diff
@@ -382,15 +390,15 @@ class BlueprintViewTests: XCTestCase {
             var child: Element?
 
             var content: ElementContent {
-                if let child = self.child {
+                if let child = child {
```
kdaker, line 393: linter fix

BlueprintUI/Sources/Element/ElementState.swift
```diff
@@ -622,6 +634,9 @@ final class ElementState {
             $0.wasVisited = false
             $0.hasUpdatedChildrenDuringLayout = false
             $0.clearOrderedChildren()
+            if !Self.enableCrossLayoutCaching {
```
kyleve, line 637: Should we do this at the _end_ of the layout, so we clear out data and reduce memory usage?

BlueprintUI/Tests/ElementStateTests.swift
```diff
@@ -32,6 +32,14 @@ extension ElementStateTree {
 
 class ElementStateTreeTests: XCTestCase {
 
+    override func setUp() {
```
kyleve, line 35: I wonder if we should add a test base class, and run each test twice; once with caching on, and once with it off?

kyleve: (Eg, override invoke test:)

```
override func invokeTest() {
   BlueprintView.enableCrossLayoutCaching = true
   super.invokeTest()

   BlueprintView.enableCrossLayoutCaching = false
   super.invokeTest()
}
```

kdaker: Did you mean for all the touched tests? This won't work for the `ElementStateTree` tests since some will break. We can do it for the other tests though.

BlueprintUI/Sources/Element/ElementState.swift
kdaker: Done

PR #416, opened 2023-01-14
Updates cross layout feature flag to support individual views

kdaker:
Allows setting the cross layout feature flag on individual views. The local value can be overridden by the global flag. This supports nested `BlueprintView`s but also assumes that we only layout on the main thread.


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -341,6 +344,25 @@ public final class BlueprintView: UIView {
...
+        if ElementState.enableCrossLayoutCaching == nil {
+            shouldClearCrossLayoutCaching = true
+            ElementState.enableCrossLayoutCaching = isCrossLayoutCachingEnabled
+        }
+
+        defer {
+            if shouldClearCrossLayoutCaching {
+                ElementState.enableCrossLayoutCaching = nil
+            }
+        }
```
kdaker, line 362: I might actually move this to `updateViewHierarchyIfNeeded`. Less elegant but safer, we don't want to call `updateViewHierarchyIfNeeded` without this code and we aren't wrapping any other methods with it.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -341,6 +344,25 @@ public final class BlueprintView: UIView {
         setNeedsLayout()
     }
 
+    /// Updates the layout caching flag for `ElementState` only from root level `BlueprintView`s
+    /// For out-of-band renders, this should re-use the flag value used by the parent view.
+    /// This assumes all blueprint layout calls happen on the main thread.
+    private func updateLayoutCachingFlag(_ closure: () -> Void) {
```
kdaker, line 350: I might remove this method and move the code to `updateViewHierarchyIfNeeded`. Less elegant but safer, we don't want to call `updateViewHierarchyIfNeeded` without this code and we aren't wrapping any other methods with it.

PR #419, opened 2023-02-02
[SPL] Remaining strict layouts

watt:
- Implemented `EqualStack`
- fixed bugs in `ConstrainedSize` and stacks
- fixed bugs due to subtree caching, by removing state from `StrictLayoutNode` and adding another intermediary node to capture the last proposed mode & size

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent+StrictSPL.swift
+++ b/BlueprintUI/Sources/Element/ElementContent+StrictSPL.swift
@@ -266,11 +266,11 @@ class StrictLayoutNode: StrictLayoutable {
 //            fatalError("\(type(of: element)) layout called \(layoutCount) times")
         }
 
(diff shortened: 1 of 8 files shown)


BlueprintUI/Sources/Element/ElementContent+StrictSPL.swift
```diff
@@ -176,21 +191,51 @@ public struct StrictNeutralLayout: SingleChildLayout {
 }
 
 
-class StrictLayoutNode: StrictLayoutable {
+class StrictProposalCaptureNode: StrictLayoutable {
```
n8chur, line 194: It might be nice to add a quick doc comment here, but we can do that later if you're worried about rebasing the stack!

n8chur (approved):

PR #420, opened 2023-02-02
[SPL] Associated cache

watt:
This is the cache you can add to any layout type, a.k.a. "phase cache", in the SwiftUI-mimicking API. By default, it is `Void` for all layouts.

This PR is stacked on #419

### DIFF
--- a/BlueprintUI/Sources/Layout/SPL/SPLayout.swift
+++ b/BlueprintUI/Sources/Layout/SPL/SPLayout.swift
@@ -6,14 +6,26 @@ public protocol SPLayout {
 
     typealias Subviews = LayoutSubviews
 
+    associatedtype Cache = Void
+
     func sizeThatFits(
         proposal: SizeConstraint,
-        subviews: Subviews
+        subviews: Subviews,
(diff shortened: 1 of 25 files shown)


BlueprintUI/Sources/Layout/SPL/SPLayout.swift
```diff
@@ -6,14 +6,26 @@ public protocol SPLayout {
 
     typealias Subviews = LayoutSubviews
 
+    associatedtype Cache = Void
+
     func sizeThatFits(
         proposal: SizeConstraint,
-        subviews: Subviews
+        subviews: Subviews,
+        cache: inout Self.Cache
```
n8chur, line 14: is the `Self.` here intentional/needed?

n8chur (approved):

BlueprintUI/Sources/Layout/SPL/SPLayout.swift
watt: Hm, doesn't seem needed, but I don't recall.

kyleve:
This is really cool!

PR #421, opened 2023-02-02
[SPL] Aggressive cache hinting

watt:
targeting #420

### DIFF
--- a/BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
+++ b/BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
@@ -80,6 +80,40 @@ extension SPValueCache where Key == SizeConstraint, Value == CGSize {
             return size
         }
         
+        // TODO:
+        // This has a negative perf impact on deep stack tests. Need to determine what the impact
+        // is on a wider variety of tests, and if there is a way we can avoid this for cases
+        // where it doesn't help (e.g. not in a scroll view).
+        if
+            case .atMost(let maxHeight) = key.height,
+            let size = values[.init(width: key.width, height: .unconstrained)],
+            size.height <= maxHeight
+        {
+            values[key] = size
+            return size
+        }
+        
+        if
+            case .atMost(let maxWidth) = key.width,
+            let size = values[.init(width: .unconstrained, height: key.height)],
+            size.width <= maxWidth
+        {
+            // TODO: test with and without caching this
+            values[key] = size
+            return size
+        }
+
+        if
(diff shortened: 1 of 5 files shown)


BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
```diff
@@ -55,38 +55,136 @@ final class SPCacheTree<Key, Value, SubcacheKey> where Key: Hashable, SubcacheKe
...
     func get(key: Key, or create: (Key) -> Value) -> Value {
         if let size = values[key] {
             Logger.logCacheHit(object: self, description: path, constraint: key)
             return size
         }
         
+        // TODO:
+        // This has a negative perf impact on deep stack tests. Need to determine what the impact
+        // is on a wider variety of tests, and if there is a way we can avoid this for cases
+        // where it doesn't help (e.g. not in a scroll view).
```
n8chur, line 95: Does this TODO need to be updated? It looks like we addressed this by introducing an option that's only true during layout.

BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
```diff
@@ -55,38 +55,136 @@ final class SPCacheTree<Key, Value, SubcacheKey> where Key: Hashable, SubcacheKe
...
+        
         Logger.logCacheMiss(object: self, description: path, constraint: key)
         
         Logger.logMeasureStart(object: self, description: path, constraint: key)
         let size = create(key)
         Logger.logMeasureEnd(object: self)
 
         values[key] = size
+        
+        if options.hintRangeBoundaries {
```
n8chur, line 154: Do we anticipate a use case for disabling this? Are we thinking about feature flagging this?

n8chur (approved):

BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
n8chur: Er, hm, maybe we do still want to think through scoping this further, I'm not entirely sure.

watt: It's not yet addressed for the condition of "are we in a scroll view or other layout that measures unconstrained", which could be important.

BlueprintUI/Sources/Internal/SPL/SPCacheTree.swift
watt: The case I had in mind was to be able to run performance tests with it turned on & off.

PR #422, opened 2023-02-04
[SPL] Regression testing options

watt:
- expose layout result for "snapshot" comparison
- hoist SwiftUI-SPL layout options up to the enum
- fix strict stack layout in some cases I found in regression tests

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -48,17 +48,14 @@ public struct ElementContent {
         switch layoutMode {
         case .standard:
             return storage.measure(in: constraint, environment: environment, cache: cache)
-        case .singlePass:
+        case .singlePass(let options):
             return storage.sizeThatFits(
                 proposal: constraint,
                 context: .init(
                     // TODO: Hoist upward?
                     cache: SPCacheNode(
                         path: "m",
-                        options: SPCacheOptions(
-                            hintRangeBoundaries: true,
-                            searchUnconstrainedKeys: false
-                        )
+                        options: options
                     ),
                     environment: environment
                 )
--- a/BlueprintUI/Sources/Layout/LayoutMode.swift
+++ b/BlueprintUI/Sources/Layout/LayoutMode.swift
@@ -6,4 +6,6 @@ public enum LayoutMode: Equatable {
     case standard
     case singlePass(options: SPCacheOptions = .default)
     case strictSinglePass
+    
+    public static let singlePass: Self = .singlePass()
 }
--- a/BlueprintUI/Sources/Layout/LayoutMode.swift
+++ b/BlueprintUI/Sources/Layout/LayoutMode.swift
@@ -4,6 +4,6 @@ public enum LayoutMode: Equatable {
     public static let `default`: Self = .strictSinglePass
 
     case standard
-    case singlePass
+    case singlePass(options: SPCacheOptions = .default)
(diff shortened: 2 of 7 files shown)


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -48,17 +48,14 @@ public struct ElementContent {
...
+        case .singlePass(let options):
             return storage.sizeThatFits(
                 proposal: constraint,
                 context: .init(
                     // TODO: Hoist upward?
                     cache: SPCacheNode(
                         path: "m",
-                        options: SPCacheOptions(
-                            hintRangeBoundaries: true,
-                            searchUnconstrainedKeys: false
```
n8chur, line 60: Are we sure we want to use a valid of `false` for `searchCnconstrainedKeys` here? Do we have a TODO list item to dig deeper into the performance trade-off here?

BlueprintUI/Sources/Layout/Stack.swift
```diff
@@ -1012,22 +1012,17 @@ extension StackLayout {
...
-        }()
-        
-        switch (alignment, crossPressure) {
-        case (.fill, .natural):
+        // If we're in fill mode, we'll do an additional pass to ensure we fill the assigned space,
+        // in case:
+        // - cross pressure is natural, so we still need to fill the calculated size
+        // - we skipped a pass during cross segments, and didn't fill along axis yet
+        switch alignment {
+        case .fill:
```
n8chur, line 1025: oof, this is another hit to Strict SPL performance, right?

n8chur (approved):

BlueprintUI/Sources/Layout/LayoutMode.swift
```diff
@@ -4,6 +4,6 @@ public enum LayoutMode: Equatable {
     public static let `default`: Self = .strictSinglePass
 
     case standard
-    case singlePass
+    case singlePass(options: SPCacheOptions = .default)
     case strictSinglePass
```
n8chur, line 8: Feel free to discard if you don't like this pattern:

```suggestion
    case strictSinglePass
    
    public static let singlePass: Self = .singlePass()
```

BlueprintUI/Sources/Element/ElementContent.swift
watt: Hm, I thought you had convinced me previously that it didn't matter during measurement, but I could be misremembering. Anyhow, I added an item to come back to this.

BlueprintUI/Sources/Layout/Stack.swift
watt: Yeah. There might be a tighter condition we could apply, but this much fixed the regression tests.

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -48,17 +48,14 @@ public struct ElementContent {
...
         case .standard:
             return storage.measure(in: constraint, environment: environment, cache: cache)
-        case .singlePass:
+        case .singlePass(let options):
             return storage.sizeThatFits(
                 proposal: constraint,
                 context: .init(
                     // TODO: Hoist upward?
                     cache: SPCacheNode(
                         path: "m",
```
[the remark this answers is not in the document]
watt: > For my own education, what does "m" refer to here?

"measure". A little (temporary) trick to make it easier to distinguish out of band measurements during layout in Instruments.

PR #424, opened 2023-02-08
Fix Image aspectFit measurement

n8chur:
## Overview

`Image`'s `aspectFit` `contentMode` currently reports measured sizes that correspond to the expected result of an `aspectFill` measurement. This leads to some weird bugs in layout.

For example, take this simple layout as an example:
```swift
        Column(
            alignment: .fill,
            minimumSpacing: 16
        ) {
            Image(
                image: {
                    let size = CGSize(width: 60, height: 30)
                    return UIGraphicsImageRenderer(size: size).image { rendererContext in
                        UIColor.red.withAlphaComponent(0.2).setFill()
                        rendererContext.fill(CGRect(origin: .zero, size: size))
                    }
                }(),
                contentMode: .aspectFit
            )
            .stackLayoutChild(priority: .fixed)

            Label(text: "Hello World")
        }
        .scrollable(.fittingHeight) { $0.centersUnderflow = true }
```

| Current | With Fix |
| - | - |
| Current | Fixed |

## Integration PRs

https://github.com/squareup/ios-register/pull/80822
https://github.com/squareup/market/pull/5562

### DIFF
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -141,15 +141,17 @@ extension Image {
                     mode = .fitHeight(height)
                 }
 
+            case (.aspectFill, .atMost(let width), _),
+                 (.aspectFit, .atMost(let width), _):
+                mode = .fitWidth(width)
+
+            case (.aspectFill, _, .atMost(let height)),
+                 (.aspectFit, _, .atMost(let height)):
+                mode = .fitHeight(height)
+
             case (.aspectFill, _, _),
                  (.aspectFit, _, _):
-                if case .atMost(let width) = constraint.width {
-                    mode = .fitWidth(width)
-                } else if case .atMost(let height) = constraint.height {
-                    mode = .fitHeight(height)
-                } else {
-                    mode = .useImageSize
-                }
+                mode = .useImageSize
             }
 
             switch mode {
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -122,17 +122,28 @@ extension Image {
 
             let mode: Mode
 
-            switch contentMode {
-            case .center, .stretch:
+            switch (contentMode, constraint.width, constraint.height) {
+            case (.center, _, _),
+                 (.stretch, _, _):
                 mode = .useImageSize
-            case .aspectFit, .aspectFill:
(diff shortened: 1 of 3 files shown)


n8chur:
Note for Square folks reviewing this—please review [the register integration PR](https://github.com/squareup/ios-register/pull/80822) as well!

kyleve (approved):
I didn't review all the tests in depth, but the change makes sense to me. Thanks for doing this!

BlueprintUICommonControls/Tests/Sources/ImageTests.swift
```diff
@@ -46,4 +37,355 @@ class ImageTests: XCTestCase {
...
+
+        validate(
+            size: .init(width: 20, height: 10),
+            constraint: .init(width: .atMost(5), height: .atMost(5)),
+            expectedValue: .init(width: 10, height: 5)
+        )
+
+        validate(
+            size: .init(width: 20, height: 10),
+            constraint: .init(width: .atMost(40), height: .atMost(40)),
```
watt, line 84: I'd like to see a case where the constraint is fixed but is has a "longer" aspect ratio than the image, e.g. 40 x 10. This only exercises one side of the `CGSize(width: width, height: height).aspectRatio > imageSize.aspectRatio` comparison.

Same request for the tall image, and for both wide & tall in `aspectFit`

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -122,17 +122,28 @@ extension Image {
...
+            case (.aspectFit, .atMost(let width), .atMost(let height)):
+                if CGSize(width: width, height: height).aspectRatio < imageSize.aspectRatio {
+                    mode = .fitWidth(width)
+                } else {
+                    mode = .fitHeight(height)
+                }
+
+            case (.aspectFill, _, _),
+                 (.aspectFit, _, _):
+                if case .atMost(let width) = constraint.width {
```
watt, line 146: Can you inline these into the outer switch cases?

watt (approved):

BlueprintUICommonControls/Tests/Sources/ImageTests.swift
n8chur: 47adc98

BlueprintUICommonControls/Sources/Image.swift
n8chur: d4cc65a

PR #425, opened 2023-02-08
Allow lifecycle callbacks to safety trigger re-renders

kyleve:
This fixes an issue where folks would in some cases have to `Dispatch.async` the work they perform in their `onAppear` (or `onDisappear) blocks, because the lifecycle callbacks were triggered inside the update. They're not really parts of the render semantically (they more mean "the layout is done, now the element is visible), so this seems more correct.

To properly test this, I had to move some stuff around:
1) I copied our test host helper over from Market/Listable/Workflow.
2) I moved LifecycleObserver up into `BlueprintUI`, since it's not really a "control".
3) I SPI'd `PassthroughView` and moved it into `BlueprintUI` as well.


BlueprintUI/Tests/XCTestCase+AppHost.swift
```diff
@@ -0,0 +1,94 @@
...
+    ///
+    public func show<ViewController: UIViewController>(
+        vc viewController: ViewController,
+        loadAndPlaceView: Bool = true,
+        test: (ViewController) throws -> Void
+    ) rethrows {
+
+        var temporaryWindow: UIWindow? = nil
+
+        func rootViewController() -> UIViewController {
```
kyleve, line 24: The two cases in here ensure we can run the tests in both cocoapods world, which provides an app host, and SPM, which does not (there is no option in SPM to provide an app host).

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -370,17 +370,21 @@ public final class BlueprintView: UIView {
             )
         )
 
+        hasUpdatedViewHierarchy = true
+        isInsideUpdate = false
+
+        /// We intentionally deliver our lifecycle callbacks (eg, `onAppear`,
+        /// `onDisappear`, etc, _after_ we've marked our view as updated.
+        /// This is inc ase the `onAppear` callback triggers a re-render,
+        /// we don't hit our recurisve update precondition.
```
watt, line 379: I dunno, this makes me nervous. If we allow this then you can get into weird scenarios where there's a render pass with callbacks, and it forces another layout which also has callbacks, and now they're out of order — the recursive render gets to complete its callbacks before the first one. In the worst case you could kick off a loop that never allows the render to complete.

Fundamentally I'm not sure there's a strong justification for anyone needing to force an explicit re-render during `onAppear`. Why did the element need to change as a result of appearing, instead of just using that element to begin with? Why did it need to change immediately and synchronously? Seems like there are some holes in this story.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -370,17 +370,21 @@ public final class BlueprintView: UIView {
             )
         )
 
+        hasUpdatedViewHierarchy = true
+        isInsideUpdate = false
+
+        /// We intentionally deliver our lifecycle callbacks (eg, `onAppear`,
+        /// `onDisappear`, etc, _after_ we've marked our view as updated.
+        /// This is inc ase the `onAppear` callback triggers a re-render,
```
robmaceachern, line 378: Nit:
```suggestion
        /// This is in case the `onAppear` callback triggers a re-render,
```

robmaceachern (approved):
I think this change is good. I'm not sure if the debug warning we were talking about has a ton of value, so I can live without it.

The nested callback case pointed out by Andrew is interesting but I'm not 100% sure if we need to address that.

n8chur (approved):

kyleve:
Once again, annoyed about the extension thing

```
/Users/runner/work/Blueprint/Blueprint/BlueprintUI/Tests/XCTestCase+AppHost.swift:25:43: error: 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
            if let rootVC = UIApplication.shared.delegate?.window??.rootViewController {
                                          ^~~~~~
UIKit.UIApplication:4:20: note: 'shared' has been explicitly marked unavailable here
    open class var shared: UIApplication { get }
```

watt (approved):
just discovered I had this pending review from a while ago

BlueprintUI/Tests/XCTestCase+AppHost.swift
kyleve: I'm trying to make it work/exist as close to as possible to a "real" UI – hence pulling in the app host changes. While it passed before, there's enough special UIKit behavior that exists when being in a real app that I'd rather we just mirror real apps as much as possible.

BlueprintUI/Sources/Internal/PassthroughView.swift
```diff
@@ -1,16 +1,16 @@
 import UIKit
 
 /// View which does not personally receive touches but allows its subviews to receive touches.
-final class PassthroughView: UIView {
+@_spi(BlueprintPassthroughView) public final class PassthroughView: UIView {
```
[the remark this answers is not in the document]
kyleve: Not unrelated; moving `LifecycleObserver`, which depends on `PassthroughView`, means we need to expose this.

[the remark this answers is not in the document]
kyleve: (Also sorry, to be clear, one of the reasons I moved this was to get this running under our cocoapods tests – which right now does not run any of the CommonControls tests...)

nononoah (approved):

PR #427, opened 2023-02-10
Measure `Image` in `aspectFill` `contentMode` the same as `aspectFit`

n8chur:
This prevents aggressively taking up space when using the `aspectFill` `contentMode`. Doing so will end up creating layouts that are more likely to show the entire image contents without cropping when it could fit the proposed space.

Additionally adds `stretch` measurement unit tests.


kyleve (approved):

watt (approved):
:

PR #430, opened 2023-02-15
Ensure nested BlueprintViews inherit their environment when inside a UIViewElement

kyleve:
Because these aren't placed into a view hierarchy anywhere, we need to manually perform this wiring.

### DIFF
--- a/BlueprintUI/Sources/Element/UIViewElement.swift
+++ b/BlueprintUI/Sources/Element/UIViewElement.swift
@@ -154,6 +154,13 @@ private final class UIViewElementMeasurer {
 
         let view = measurementView(for: element)
 
+        /// Ensure that during measurement / sizing, the inherited `Environment` is available
+        /// to any child `BlueprintView`s. We must manually wire this property up, as the
+        /// measurement views are not in the view hierarchy.
+
+        view.nativeViewNodeBlueprintEnvironment = environment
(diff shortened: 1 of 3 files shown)


BlueprintUI/Sources/Element/UIViewElement.swift
```diff
@@ -154,6 +154,13 @@ private final class UIViewElementMeasurer {
 
         let view = measurementView(for: element)
 
+        /// Ensure that during measurement / sizing, the inherited `Environment` is available
+        /// to any child `BlueprintView`s. We must manually wire this property up, as the
+        /// measurement views are not in the view hierarchy.
+
+        view.nativeViewNodeBlueprintEnvironment = environment
+        defer { view.nativeViewNodeBlueprintEnvironment = nil }
```
n8chur, line 162: Not sure if this matters or is necessary, but should we store the previous value and then restore it in `defer` rather than assuming it's `nil`?

n8chur (approved):

BlueprintUI/Sources/Element/UIViewElement.swift
kyleve: Blueprint is the only thing that sets this property or reads it; so I think nil-ing out is correct / better; also to avoid retaining anything stored in the environment in the static prototype views.

PR #436, opened 2023-03-06
Fix documentation generation by setting explicit ruby version

robmaceachern:
Our documentation generation step [started failing](https://github.com/square/Blueprint/actions/runs/4256655333/jobs/7405802101) after the update to Xcode 14 / macos-12. This PR configures the docs workflow to use an explicit Ruby version (the same as the tests workflow was configured to use).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -77,10 +77,6 @@ jobs:
         ruby-version: ${{ env.ruby_version }}
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
 
-    - name: Bundle Install
-      run: |
-        bundle check || bundle install --path .bundle
-
     - name: Pod Install
       run: |
         bundle exec pod install --project-directory=SampleApp/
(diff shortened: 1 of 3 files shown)


kyleve (approved):

.github/workflows/tests.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
     - name: Setup Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6
+        ruby-version: ${{ env.ruby_version }}
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
```
watt, line 78: Dang, I'm just noticing the whole step below this one is unnecessary. Feel free to remove that while you're in here, if you want.

watt (approved):

.github/workflows/tests.yaml
robmaceachern: The `Bundle Install` step is unnecessary? Doesn't that install cocoapods?

watt: Yeah. [setup-ruby does that automatically](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically), and better, since it uses a cache.

robmaceachern: TIL! Cool will do.

PR #438, opened 2023-03-08
Update Cocoapods

kyleve:


Gemfile.lock
```diff
@@ -1,41 +1,40 @@
 GEM
```
kyleve, line 1: Do we care about the internal changelog for this one?

robmaceachern:
I see you're hitting the same ruby version error I'm hitting [over here](https://github.com/square/Blueprint/actions/runs/4369380230/jobs/7643128830) while attempting to [fix the doc generation](https://github.com/square/Blueprint/pull/439).

```
Fetching gem metadata from https://rubygems.org/..............
activesupport-7.0.4.2 requires ruby version >= 2.7.0, which is incompatible with
the current version, ruby 2.6.10p210
Error: The process '/Users/runner/hostedtoolcache/Ruby/2.6.10/x64/bin/bundle' failed with exit code 5
```

It looks like the jazzy update I'm doing has already updated Cocoapods to 1.12 so we can probably close this unless you updated other gems too.

kyleve:
sg, closing!

PR #439, opened 2023-03-08
Fix `generate_docs` not finding sourcekitten

robmaceachern:
https://github.com/square/Blueprint/pull/436 fixed one of the errors with doc generation but then revealed another issue with the script unable to find the sourcekitten binary in the `generate_docs.sh` script.

This PR:
- Executes the `generate_docs.sh` script with `bundle exec` so that it can properly reference the installed gems.
- Updates the jazzy gem to 0.14.3 (required because the previous version was generating a `Illegal instruction: 4` error on the new macOS version)
- Updates cocoapods to 1.12 (it is a dependency of jazzy).
- Updates the Ruby version to 2.7 (to avoid a compatibility error caused by the cocoapods update: `activesupport-7.0.4.2 requires ruby version >= 2.7.0, which is incompatible with the current version, ruby 2.6.10p210`)

Verified the doc generation works [here](https://github.com/square/Blueprint/actions/runs/4375279743/jobs/7655796185).


kyleve (approved):
LGTM; at some point we should likely just crib over the market DocC script and deprecate Jazzy usage entirely.

PR #440, opened 2023-03-10
1) Introduce mode control and debugging types

watt:
This is PR 1 in a stack targeting `feature/caffeinated-layout`. You can see how it fits into the larger picture by looking at the staging branch in #434 .

It introduces the `LayoutMode` enum and `RenderContext` type, along with some setting up some related debugging changes. Largely the same as it was in #374, targeting the SPL branch.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -36,6 +36,8 @@ public final class BlueprintView: UIView {
 
     private let rootController: NativeViewController
 
+    private var layoutResult: LayoutResultNode?
+
     private var sizesThatFit: [SizeConstraint: CGSize] = [:]
 
     /// A base environment used when laying out and rendering the element tree.
@@ -97,6 +99,16 @@ public final class BlueprintView: UIView {
             guard oldValue.size != bounds.size else { return }
 
             invalidateIntrinsicContentSize()
+        }
+    }
+
+    /// An optional explicit layout mode for this view. If `nil`, this view will inherit the layout
+    /// mode of its nearest ancestor, or use ``LayoutMode/default``.
+    public var layoutMode: LayoutMode? {
+        didSet {
+            if layoutMode != oldValue {
+                setNeedsViewHierarchyUpdate()
+            }
         }
     }
 
@@ -211,11 +223,18 @@ public final class BlueprintView: UIView {
         )
         defer { Logger.logSizeThatFitsEnd(view: self) }
 
-        let measurement = element.content.measure(
-            in: constraint,
-            environment: makeEnvironment(),
-            cache: CacheFactory.makeCache(name: cacheName)
-        )
+        let environment = makeEnvironment()
+        let layoutMode = environment.layoutMode
+        let renderContext = RenderContext(layoutMode: layoutMode)
+
+        let measurement = renderContext.perform {
+            element.content.measure(
+                in: constraint,
+                environment: environment,
+                cache: CacheFactory.makeCache(name: cacheName),
+                layoutMode: layoutMode
+            )
+        }
 
         sizesThatFit[constraint] = measurement
 
@@ -339,10 +358,21 @@ public final class BlueprintView: UIView {
             size: bounds.size + rootCorrection.size
         )
 
-        /// Grab view descriptions
-        let viewNodes = element?
-            .layout(layoutAttributes: LayoutAttributes(frame: rootFrame), environment: environment)
-            .resolve() ?? []
+        let layoutMode = environment.layoutMode
+        let renderContext = RenderContext(layoutMode: layoutMode)
+
+        // Perform layout
+        let layoutResult = renderContext.perform {
+            element?.layout(
+                frame: rootFrame,
+                environment: environment,
+                layoutMode: layoutMode
+            )
+        }
+        self.layoutResult = layoutResult
+
+        // Flatten into tree of view descriptions
+        let viewNodes = layoutResult?.resolve() ?? []
 
         let layoutEndTime = CACurrentMediaTime()
         Logger.logLayoutEnd(view: self)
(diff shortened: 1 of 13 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -401,6 +431,27 @@ public final class BlueprintView: UIView {
...
+    @_spi(BlueprintDebugging)
+    public func forceSynchronousLayout() {
+        setNeedsViewHierarchyUpdate()
+        updateViewHierarchyIfNeeded()
+    }
+
+    /// Dumps the result of the most recent layout, by recursing through the layout tree and calling
+    /// the provided visitor on each node. By default, this prints to `stdout`.
+    @_spi(BlueprintDebugging)
+    public func dumpLayoutResult(
```
kyleve, line 444: I think this is somewhat duplicated with the impl on `LayoutResultNode`, can we just call into that impl?

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -41,24 +41,24 @@
 
  You will note that the identifiers remain stable, which ultimately ensures that views are reused.
  */
-struct ElementIdentifier: Hashable, CustomDebugStringConvertible {
+struct ElementIdentifier: Hashable, CustomStringConvertible {
```
kyleve, line 44: Noting for later: Another win we might get is [adopting the `ElementIdentifier` caching](https://github.com/square/Blueprint/pull/398/files#diff-44612c52b487b560a6d4dadecd96df4cd9e7c4a53a7f9c9f5b2b933a9700649f) from the caching branch. It's likely not as valuable here given the measurement caches themselves don't leverage Element ID as heavily, but it likely still provides some benefit.

kyleve (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
watt: > I think this is somewhat duplicated with the impl on `LayoutResultNode`, can we just call into that impl?

Sure, I think at one point it was handy to have both, but we don't need them both now.

watt: I'm going to defer small things like this to a follow-up PR to avoid having to rebase the rest of the stack.

BlueprintUI/Sources/Internal/ElementIdentifier.swift
```diff
@@ -41,24 +41,24 @@
...
-    let elementType: ObjectIdentifier
+    let elementType: Metatype
     let key: AnyHashable?
 
     let count: Int
 
     init(elementType: Element.Type, key: AnyHashable?, count: Int) {
 
-        self.elementType = ObjectIdentifier(elementType)
+        self.elementType = Metatype(elementType)
```
nononoah, line 53: Changed because it was a cheaper or more legible?

BlueprintUI/Sources/Layout/LayoutMode.swift
```diff
@@ -0,0 +1,20 @@
...
+/// Controls the layout system that Blueprint uses to lay out elements.
+///
+/// Blueprint supports multiple layout systems. Each is expected to produce the same result, but
+/// some may have different performance profiles or special requirements.
+///
+/// You can change the layout system used by setting the ``BlueprintView/layoutMode`` property, but
+/// generally you should use the ``default`` option.
+///
+public enum LayoutMode: Equatable {
+    public static let `default`: Self = .legacy
```
nononoah, line 12: Do we expect to swap the default in time?

nononoah (approved):

BlueprintUI/Sources/Internal/ElementIdentifier.swift
watt: > Changed because it was a cheaper or more legible?

The latter I suppose. `Any.Type` is annoying to work with because it's not equatable or hashable, and `ObjectIdentifier` has a useless `description`.

BlueprintUI/Sources/Internal/RenderContext.swift
```diff
@@ -0,0 +1,25 @@
...
+    var layoutMode: LayoutMode
+
+    /// Perform the given block with this as the current render context, restoring the previous
+    /// context before returning.
+    func perform<Result>(block: () throws -> Result) rethrows -> Result {
+        let previous = Self.current
+        defer { Self.current = previous }
+
+        Self.current = self
+
```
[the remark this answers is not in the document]
watt: > In other words, can we begin a perform with one LayoutMode and end with another?

No, though there could hypothetically be a nested call to `perform` that uses a different mode. I hope we don't have to think about that too much.

> my question is why we have RenderContext if the information is available on the environment

It's primarily to support out-of-band operations, e.g. any time something calls `element.content.measure()` directly, and the right environment isn't passed along. This used to be more common thing — thankfully now the main use case is folks doing that inside a `GeometryReader`.

It's also handy place to read from, though.

BlueprintUI/Sources/Layout/LayoutMode.swift
watt: Yep! We also need a way to change this at runtime, so it'll probably become a `var` before the feature branch lands.

BlueprintUI/Sources/Internal/RenderContext.swift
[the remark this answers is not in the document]
robmaceachern: > any time something calls element.content.measure() directly, and the right environment isn't passed along

> thankfully now the main use case is folks doing that inside a GeometryReader

Just so I understand this: Are you saying that `GeometryReader` doesn't pass along the correct environment?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -339,10 +358,21 @@ public final class BlueprintView: UIView {
...
 
-        /// Grab view descriptions
-        let viewNodes = element?
-            .layout(layoutAttributes: LayoutAttributes(frame: rootFrame), environment: environment)
-            .resolve() ?? []
+        let layoutMode = environment.layoutMode
+        let renderContext = RenderContext(layoutMode: layoutMode)
+
+        // Perform layout
+        let layoutResult = renderContext.perform {
```
robmaceachern, line 365: Nit: `self.layoutResult = ...` (and delete `self.layoutResult = layoutResult`)?

robmaceachern (approved):

BlueprintUI/Sources/Internal/RenderContext.swift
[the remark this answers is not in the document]
watt: > Just so I understand this: Are you saying that `GeometryReader` doesn't pass along the correct environment?

`GeometryReader` itself does, but it's not uncommon to see this sort of (bad) thing:
```swift
GeometryReader { geometry in
  if wideElement.content.measure(in: geometry.constraint, environment: .empty).width < myMaxWidth {
    return wideElement
  } else {
    return narrowElement
  }
}
```

watt:
Follow-ups going into #446

PR #442, opened 2023-03-10
3) Caffeinated layout path

watt:
This is PR 3 in a stack targeting `feature/caffeinated-layout`. You can see how it fits into the larger picture by looking at the staging branch in #434.

It creates the new parallel layout API:
- `ContentStorage` gets new methods
- `Layout` is turned into a composition of `LegacyLayout` and `CaffeinatedLayout`
- `SingleChildLayout` is also turned into a composition of `LegacySingleChildLayout` and `CaffeinatedSingleChildLayout`
- some new types like `LayoutSubelement` are stubbed out
- all new API methods have a default implementation that calls `fatalError`

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -16,7 +16,7 @@ public struct ElementContent {
         measure(
             in: constraint,
             environment: environment,
-            cache: CacheFactory.makeCache(name: "ElementContent"),
+            cacheName: "ElementContent",
             layoutMode: RenderContext.current?.layoutMode ?? environment.layoutMode
         )
     }
@@ -24,30 +24,57 @@ public struct ElementContent {
     func measure(
         in constraint: SizeConstraint,
         environment: Environment,
-        cache: CacheTree,
+        cacheName: String,
         layoutMode: LayoutMode
     ) -> CGSize {
-        // TODO: switch on layoutMode
-        storage.measure(in: constraint, environment: environment, cache: cache)
+        switch layoutMode {
+        case .legacy:
+            return measure(
+                in: constraint,
+                environment: environment,
+                cache: CacheFactory.makeCache(name: cacheName)
+            )
+        case .caffeinated:
+            return sizeThatFits(
+                proposal: constraint,
+                context: MeasureContext(environment: environment)
+            )
+        }
     }
 
     func measure(in constraint: SizeConstraint, environment: Environment, cache: CacheTree) -> CGSize {
         storage.measure(in: constraint, environment: environment, cache: cache)
     }
 
+    func sizeThatFits(proposal: SizeConstraint, context: MeasureContext) -> CGSize {
+        storage.sizeThatFits(proposal: proposal, context: context)
+    }
+
     public var childCount: Int {
         storage.childCount
     }
 
-    func performLayout(
+    typealias IdentifiedNode = (identifier: ElementIdentifier, node: LayoutResultNode)
+
+    func performLegacyLayout(
         attributes: LayoutAttributes,
         environment: Environment,
         cache: CacheTree
-    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
-        storage.performLayout(
+    ) -> [IdentifiedNode] {
+        storage.performLegacyLayout(
             attributes: attributes,
             environment: environment,
             cache: cache
+        )
+    }
+
+    func performCaffeinatedLayout(
+        frame: CGRect,
+        context: LayoutContext
+    ) -> [IdentifiedNode] {
+        storage.performCaffeinatedLayout(
+            frame: frame,
+            context: context
         )
     }
 }
@@ -107,7 +134,7 @@ extension ElementContent {
     public init(
         child: Element,
         key: AnyHashable? = nil,
-        layout: SingleChildLayout
+        layout: some SingleChildLayout
     ) {
         self = ElementContent(layout: SingleChildLayoutHost(wrapping: layout)) {
             $0.add(key: key, element: child)
@@ -247,11 +274,13 @@ extension ElementContent {
 
 // All layout is ultimately performed by the `Layout` protocol – this implementations delegates to a wrapped
 // `SingleChildLayout` implementation for use in elements with a single child.
-fileprivate struct SingleChildLayoutHost: Layout {
+fileprivate struct SingleChildLayoutHost<WrappedLayout: SingleChildLayout>: Layout {
 
-    private var wrapped: SingleChildLayout
+    typealias Cache = WrappedLayout.Cache
 
-    init(wrapping layout: SingleChildLayout) {
+    private var wrapped: WrappedLayout
+
+    init(wrapping layout: WrappedLayout) {
         wrapped = layout
     }
 
@@ -265,6 +294,37 @@ fileprivate struct SingleChildLayoutHost: Layout {
         return [
             wrapped.layout(size: size, child: items.map { $0.content }.first!),
         ]
+    }
+
+    func sizeThatFits(
+        proposal: SizeConstraint,
+        subelements: Subelements,
+        cache: inout Cache
+    ) -> CGSize {
+        precondition(subelements.count == 1)
+        return wrapped.sizeThatFits(
+            proposal: proposal,
+            subelement: subelements[0],
+            cache: &cache
+        )
+    }
+
+    func placeSubelements(
+        in size: CGSize,
+        subelements: Subelements,
+        cache: inout Cache
+    ) {
+        precondition(subelements.count == 1)
+        wrapped.placeSubelement(
+            in: size,
+            subelement: subelements[0],
+            cache: &cache
+        )
+    }
+
+    func makeCache(subelements: Subelements) -> Cache {
+        precondition(subelements.count == 1)
+        return wrapped.makeCache(subelement: subelements[0])
     }
 }
 
--- a/BlueprintUI/Sources/Internal/LayoutResultNode.swift
+++ b/BlueprintUI/Sources/Internal/LayoutResultNode.swift
@@ -5,21 +5,58 @@ extension Element {
     /// Build a fully laid out element tree with complete layout attributes
     /// for each element.
     ///
-    /// - Parameter layoutAttributes: The layout attributes to assign to the
-    ///   root element.
     ///
     /// - Returns: A layout result
-    func layout(layoutAttributes: LayoutAttributes, environment: Environment) -> LayoutResultNode {
-        LayoutResultNode(
-            root: self,
+    /// - Parameters:
+    ///   - frame: the root element's frame
+    ///   - environment: the root environment
+    ///   - layoutMode: the mode to use for layout
+    func layout(frame: CGRect, environment: Environment, layoutMode: LayoutMode) -> LayoutResultNode {
+        switch layoutMode {
+        case .legacy:
+            return legacyLayout(
+                layoutAttributes: LayoutAttributes(frame: frame),
+                environment: environment
+            )
+
+        case .caffeinated:
+            return caffeinatedLayout(
+                frame: frame,
+                context: LayoutContext(
+                    environment: environment
+                )
+            )
+        }
+    }
+
+    private func legacyLayout(layoutAttributes: LayoutAttributes, environment: Environment) -> LayoutResultNode {
+        let cache = CacheFactory.makeCache(name: "\(type(of: self))")
+        let children = content.performLegacyLayout(
+            attributes: layoutAttributes,
+            environment: environment,
+            cache: cache
+        )
+
+        return LayoutResultNode(
+            element: self,
             layoutAttributes: layoutAttributes,
-            environment: environment
+            environment: environment,
+            children: children
         )
     }
 
-    func layout(frame: CGRect, environment: Environment, layoutMode: LayoutMode) -> LayoutResultNode {
-        // TODO: switch on layoutMode
-        layout(layoutAttributes: LayoutAttributes(frame: frame), environment: environment)
+    private func caffeinatedLayout(frame: CGRect, context: LayoutContext) -> LayoutResultNode {
+        let children = content.performCaffeinatedLayout(
+            frame: frame,
+            context: context
+        )
+
+        return LayoutResultNode(
+            element: self,
+            layoutAttributes: LayoutAttributes(frame: frame),
+            environment: context.environment,
+            children: children
+        )
     }
 }
 
@@ -48,21 +85,6 @@ struct LayoutResultNode {
         self.environment = environment
         self.children = children
     }
-
-    init(root: Element, layoutAttributes: LayoutAttributes, environment: Environment) {
-        let cache = CacheFactory.makeCache(name: "\(type(of: root))")
-        self.init(
-            element: root,
-            layoutAttributes: layoutAttributes,
-            environment: environment,
-            children: root.content.performLayout(
-                attributes: layoutAttributes,
-                environment: environment,
-                cache: cache
-            )
-        )
-    }
-
 }
 
 
--- a/BlueprintUI/Sources/Layout/Layout.swift
+++ b/BlueprintUI/Sources/Layout/Layout.swift
@@ -1,29 +1,125 @@
-import UIKit
+import CoreGraphics
 
-/// Conforming types can calculate layout attributes for an array of children.
-public protocol Layout {
+/// A type that defines the geometry of a collection of elements.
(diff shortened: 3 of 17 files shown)


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -16,40 +16,67 @@ public struct ElementContent {
...
+    func sizeThatFits(proposal: SizeConstraint, context: MeasureContext) -> CGSize {
+        storage.sizeThatFits(proposal: proposal, context: context)
+    }
+
     public var childCount: Int {
         storage.childCount
     }
 
-    func performLayout(
+    typealias IdentifiedNode = (identifier: ElementIdentifier, node: LayoutResultNode)
```
kyleve, line 57: This is probably something to clean up later; but I found it much nicer (and much better for debugging) in the caching branch by removing this tuple and putting the ID onto `LayoutResultNode` itself directly.

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -1,29 +1,125 @@
-import UIKit
+import CoreGraphics
 
-/// Conforming types can calculate layout attributes for an array of children.
-public protocol Layout {
+/// A type that defines the geometry of a collection of elements.
```
kyleve, line 3: Love these docs; nice work!

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    ///     this method might call the method more than once with different constraints to learn
+    ///     more about the container’s flexibility before choosing a size for placement.
+    ///   - subelements: A collection of proxies that represent the elements that the container
+    ///     arranges. You can use the proxies in the collection to get information about the
+    ///     subelements as you determine how much space the container needs to display them.
+    ///   - cache: Optional storage for calculated data that you can share among the methods of your
+    ///     custom layout container. See ``makeCache(subelements:)-4kmy2`` for details.
+    /// - Returns: A size that indicates how much space the container needs to arrange its
+    ///   subelements.
+    func sizeThatFits(
```
kyleve, line 197: We've discussed a few times exposing the `Environment` here; is it worth adding a `context` parameter to this method and the place method so we can add optional parameters without breaking changes going forward?

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    ///     container's subelements within the region. The size of this region may not match any
+    ///     size that was returned from a call to ``sizeThatFits(proposal:subelements:cache:)``.
+    ///   - subelements: A collection of proxies that represent the elements that the container
+    ///     arranges. Use the proxies in the collection to get information about the subelements and
+    ///     to tell the subelements where to appear.
+    ///   - cache: Optional storage for calculated data that you can share among the methods of your
+    ///     custom layout container. See ``makeCache(subelements:)-4kmy2`` for details.
+    func placeSubelements(
+        in size: CGSize,
+        subelements: Subelements,
```
kyleve, line 229: Note for readers: Later on, the writable properties here are wrapped in a `@Storage` property so you can write to them.

kyleve (approved):

BlueprintUI/Sources/Layout/Layout.swift
kyleve: (I don't think we have to do this here, but it's likely nice to do somewhere in this stack)

BlueprintUI/Sources/Layout/Layout.swift
watt: I should mention that most of these were cribbed from SwiftUI's docs, and then adapted to Blueprint.

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -1,29 +1,125 @@
...
+///
+/// Subelements may have _traits_ that are specific to their container's layout. The traits are of
+/// the ``Layout`` protocol's associated type ``LegacyLayout/Traits``, and each subelement can have
+/// a distinct `Traits` value assigned. You can set this in the `configure` block of
+/// ``ElementContent/init(layout:configure:)``, when you call
+/// ``ElementContent/Builder/add(traits:key:element:)``. If you do not specify a `Traits` type for
+/// your layout, it defaults to the void type, `()`.
+///
+/// Containers can choose to condition their behavior according to the traits of their subelements.
+/// For example, the ``Row`` and ``Column`` types allocate space to their subelements based in part
```
nononoah, line 96:
```suggestion
/// For example, the ``Row`` and ``Column`` types allocate space for their subelements based in part
```

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    ///
+    /// ## Sizing Rules
+    ///
+    /// For performance reasons, the layout engine may deduce the measurement of your container for
+    /// some constraint values without explicitly calling
+    /// ``sizeThatFits(proposal:subelements:cache:)``. To ensure that the deduced value is correct,
+    /// your layout must follow some ground rules:
+    ///
+    /// 1. **Given one fixed constraint axis, the element's growth along the other axis should be
+    ///    _monotonic_.** That is, an element can grow when given a larger constraint, or shrink
```
nononoah, line 172: 🐴 

I think you mean "should increase monotonically," unless of course is is OK to continue to shrink monotonically as you're offered more space.

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    /// 2. If your element has no intrinsic size along an axis, you can represent that in a couple
+    ///     ways:
+    ///     - You can return a fixed value representing the minimum size for your element. In this
+    ///       approach, a containing element is usually responsible for stretching your element to
+    ///       fill desired space.
+    ///     - You can return a size that entirely fills the constraint proposed. In this approach,
+    ///       you **must** return `.infinity` when the constraint is
+    ///       ``SizeConstraint/Axis/unconstrained``. Otherwise, your behavior would be in violation
+    ///       of rule #1.
+    ///
```
nononoah, line 185: This is excellent. It may be accompanied by links to further documentation which show examples of the convention in practice as well as definitions which break the convention and the problems that such cause.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -0,0 +1,8 @@
+import CoreGraphics
+
+
+public typealias LayoutSubelements = [LayoutSubelement]
```
nononoah, line 4: If this is how it stays, I suppose we should ignore my previous note.

BlueprintUI/Sources/Layout/SingleChildLayout.swift
```diff
@@ -21,3 +29,164 @@ public protocol SingleChildLayout {
...
+    ///   - proposal: A size constraint for the container. The container's parent element that calls
+    ///     this method might call the method more than once with different constraints to learn
+    ///     more about the container’s flexibility before choosing a size for placement.
+    ///   - subelement: A proxy that represents the element that the container arranges. You can use
+    ///     the proxy to get information about the subelement as you determine how much space the
+    ///     container needs to display it.
+    ///   - cache: Optional storage for calculated data that you can share among the methods of your
+    ///     custom layout container. See ``makeCache(subelement:)-33y4l`` for details.
+    /// - Returns: A size that indicates how much space the container needs to arrange its
+    ///   subelement.
```
nononoah, line 89: Do we need to copy this again? It might be best to reference the previous block.

BlueprintUI/Tests/Element+Testing.swift
```diff
@@ -9,7 +9,11 @@ extension Element {
     ///
     /// - Returns: A layout result
     func layout(frame: CGRect, environment: Environment = .empty) -> LayoutResultNode {
-        layout(layoutAttributes: LayoutAttributes(frame: frame), environment: environment)
+        layout(
+            frame: frame,
+            environment: environment,
+            layoutMode: RenderContext.current?.layoutMode ?? .default
```
nononoah, line 15: Seems like the first important usage of the singleton here.

I think this means that any raw call with get the default (legacy). I wonder if these tests would be more clearly rephrased to accept a layout mode as a parameter.

The difference is that existing tests may continue to be written as before, but their caffeinated analogs will require a different spelling, like:

```swift
RenderContext(layoutMode: .caffeinated).perform {
  element.layout(...)
}
```

No biggie and no need for change. I imagine this clarifies down the road.

nononoah (approved):

BlueprintUI/Sources/Internal/LayoutResultNode.swift
```diff
@@ -5,21 +5,58 @@ extension Element {
...
-    ///   root element.
     ///
     /// - Returns: A layout result
-    func layout(layoutAttributes: LayoutAttributes, environment: Environment) -> LayoutResultNode {
-        LayoutResultNode(
-            root: self,
+    /// - Parameters:
+    ///   - frame: the root element's frame
+    ///   - environment: the root environment
+    ///   - layoutMode: the mode to use for layout
```
robmaceachern, line 13: Nit: `- Returns:` below `- Parameters:`

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -1,29 +1,125 @@
...
+///
+/// If your layout is specialized for laying out a single subelement, you can use the
+/// ``SingleChildLayout`` protocol instead. It has similar methods, but is strongly typed for a
+/// single subelement instead of a collection.
+///
+/// - Note: During the transition from Blueprint's legacy layout system to Caffeinated Layout, the
+///   ``Layout`` protocol is composed of two sets of layout methods: ``LegacyLayout`` and
+///   ``CaffeinatedLayout``. While this documentation focuses on the newer layout API, you must
+///   currently implement both. Fortunately, the methods are similar, and you may be able to reuse
+///   logic.
```
robmaceachern, line 69: Good callout. Will we also need to advise on how to toggle between the layout systems for testing purposes?

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    ///
+    /// In Blueprint, parents ultimately choose the size of their children, so the actual size that
+    /// this container is laid out in may not be a size that was returned from this method.
+    ///
+    /// ## Sizing Rules
+    ///
+    /// For performance reasons, the layout engine may deduce the measurement of your container for
+    /// some constraint values without explicitly calling
+    /// ``sizeThatFits(proposal:subelements:cache:)``. To ensure that the deduced value is correct,
+    /// your layout must follow some ground rules:
```
robmaceachern, line 169: Just a thought: I wonder if it's possible to implement a general fuzzer that can be run against layout implementations to ensure they adhere to the rules.

robmaceachern (approved):

BlueprintUI/Sources/Layout/Layout.swift
```diff
@@ -33,10 +129,172 @@ public protocol Layout {
...
+    /// sure to either make the type of the cache parameters on your other ``Layout`` protocol
+    /// methods match, or use a type alias to define the ``Cache`` associated type.
+    ///
+    /// - Parameter subelements: A collection of proxy instances that represent the subelements that
+    ///   the container arranges. You can use the proxies in the collection to get information about
+    ///   the subelements as you calculate values to store in the cache.
+    /// - Returns: Storage for calculated data that you share among the methods of your custom
+    ///   layout container.
+    func makeCache(subelements: Subelements) -> Cache
+}
```
n8chur, line 276: All of these docs are fantastic—nice work!

n8chur (approved):

BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -16,40 +16,67 @@ public struct ElementContent {
...
+            return measure(
+                in: constraint,
+                environment: environment,
+                cache: CacheFactory.makeCache(name: cacheName)
+            )
+        case .caffeinated:
+            return sizeThatFits(
+                proposal: constraint,
+                context: MeasureContext(environment: environment)
+            )
```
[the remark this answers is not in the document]
watt: @n8chur is correct

BlueprintUI/Sources/Layout/LayoutSubelement.swift
watt: The `typealias` leaves the door open for us to add a custom type if we ever want to, but I think it's unlikely.

BlueprintUI/Sources/Layout/Layout.swift
watt: You may want to see for yourself: [Layout](https://developer.apple.com/documentation/swiftui/layout)

I hadn't really considered that borrowing too much documentation might be a problem (since we are intentionally trying to mimic the API). A citation is easy enough.

BlueprintUI/Sources/Layout/Layout.swift
watt: > Good callout. Will we also need to advise on how to toggle between the layout systems for testing purposes?

Like, unit and snapshot testing, or just for experimentation? Currently the advice on `LayoutMode` is to stick with the default mode. Normally I wouldn't expect folks who are writing elements to need to explicitly test both modes.

BlueprintUI/Sources/Layout/Layout.swift
watt: Yes! We've got this on the list of potential follow-up work. It would be rather slow, so it would have to be an opt-in thing or a special test harness.

BlueprintUI/Sources/Layout/Layout.swift
watt: Sure, good idea. I'm working on some supplemental documentation with diagrams.

BlueprintUI/Sources/Layout/Layout.swift
watt: Yeah, that makes sense to me, at least for the `Environment`. Not sure how I feel about further muddying the API with a context type though, since we're modeling after SwiftUI. Do you recall any specific things we talked about putting in there?

BlueprintUI/Sources/Layout/SingleChildLayout.swift
watt: > Do we need to copy this again? It might be best to reference the previous block.

Do you mean replacing these docs with a link to the analogous calls in `Layout`? I just didn't want to make folks jump through a link to read docs. I don't feel strongly about it though.

nononoah: I think it may be worth asking them to jump through rather than try to keep this large body of text up to date in multiple locations. That said, I don't feel strongly about it, nor expect it to change much.

BlueprintUI/Sources/Layout/Layout.swift
nononoah: The first few paragraphs look like a pretty direct copy. We should cite the source in a line like "Borrowed with little modification from [X](link)."

BlueprintUI/Sources/Layout/SingleChildLayout.swift
watt: OK, I see what you mean. Kept the summary and removed everything "below the fold" of Xcode's quick help [here](https://github.com/square/Blueprint/pull/446/commits/1d3548120da3fd07b63d39871c136244a88c757c#diff-4bc18050c5802990f199763152c93e481d5eafdc7911ace7a6719cfb8bad11e4).

BlueprintUI/Sources/Layout/Layout.swift
watt: How's [this](https://github.com/square/Blueprint/pull/446/commits/1d3548120da3fd07b63d39871c136244a88c757c#diff-28d36c422d80ceb0a832b0a43375d60f4602ad7f37d4d4199a02bc99dfa4ded9)?

BlueprintUI/Sources/Layout/Layout.swift
kyleve: It's more that: We don't know what we might want in there at some point, so I'd rather not cause a big breaking change to add said things later on. That being said, for the args in the `Layout`, the env is probably fine... But a `context` also feels like a fairly cheap way (a) get us the env, and (b) to avoid a large breaking change to introduce an ultimately optional param only a few layouts need later on

PR #418, opened 2023-01-27
[WIP DNR] Fixing Measurement Only Nodes, Dynamic Nodes

kyleve:

PR #443, opened 2023-03-10
4) ContentStorage implementations

watt:
This is PR 4 in a stack targeting `feature/caffeinated-layout`. You can see how it fits into the larger picture by looking at the staging branch in #434.

This PR provides the implementations for ContentStorage types:
- The `LayoutTreeNode` provides a structure to hold information re-used during the layout pass. It's roughly analogous to the `CacheTree` from legacy layout, but is not a cache itself, just a tree structure.
- `HintingSizeCache` is the new measurement cache. Hinting comes later — here it's just a dumb cache.
- `LayoutSubelement` is a proxy for child elements passed to layout implementations. It's analogous to the `(Traits, Measurable)` tuple from legacy layout, and has an API inspired by SwiftUI's [`LayoutSubview`](https://developer.apple.com/documentation/swiftui/layoutsubview)

### DIFF
--- a/BlueprintUI/Sources/Element/EnvironmentAdaptingStorage.swift
+++ b/BlueprintUI/Sources/Element/EnvironmentAdaptingStorage.swift
@@ -2,6 +2,9 @@ import CoreGraphics
 
 /// Content storage that applies a change to the environment.
 struct EnvironmentAdaptingStorage: ContentStorage {
+
+    typealias IdentifiedNode = ElementContent.IdentifiedNode
+
     let childCount = 1
 
     /// During measurement or layout, the environment adapter will be applied
@@ -10,12 +13,19 @@ struct EnvironmentAdaptingStorage: ContentStorage {
     var adapter: (inout Environment) -> Void
 
     var child: Element
+    var content: ElementContent
+
+    init(adapter: @escaping (inout Environment) -> Void, child: Element) {
+        self.adapter = adapter
+        self.child = child
+        content = child.content
+    }
 
     func performLegacyLayout(
         attributes: LayoutAttributes,
         environment: Environment,
         cache: CacheTree
-    ) -> [(identifier: ElementIdentifier, node: LayoutResultNode)] {
+    ) -> [IdentifiedNode] {
         let environment = adapted(environment: environment)
 
         let childAttributes = LayoutAttributes(size: attributes.bounds.size)
@@ -26,7 +36,7 @@ struct EnvironmentAdaptingStorage: ContentStorage {
             element: child,
             layoutAttributes: childAttributes,
             environment: environment,
-            children: child.content.performLegacyLayout(
+            children: content.performLegacyLayout(
                 attributes: childAttributes,
                 environment: environment,
                 cache: cache.subcache(element: child)
@@ -54,6 +64,44 @@ struct EnvironmentAdaptingStorage: ContentStorage {
                 cache: cache.subcache(element: child)
             )
         }
+    }
+
+    func sizeThatFits(proposal: SizeConstraint, context: MeasureContext) -> CGSize {
+        let environment = adapted(environment: context.environment)
+        let identifier = ElementIdentifier(elementType: type(of: child), key: nil, count: 1)
+        let context = MeasureContext(
+            environment: environment,
+            node: context.node.subnode(key: identifier)
+        )
+        return child.content.sizeThatFits(proposal: proposal, context: context)
+    }
+
+    func performCaffeinatedLayout(
+        frame: CGRect,
+        context: LayoutContext
+    ) -> [IdentifiedNode] {
+        let environment = adapted(environment: context.environment)
+
+        let childAttributes = LayoutAttributes(size: frame.size)
+
+        let identifier = ElementIdentifier(elementType: type(of: child), key: nil, count: 1)
+
+        let context = LayoutContext(
+            environment: environment,
+            node: context.node.subnode(key: identifier)
+        )
+
+        let node = LayoutResultNode(
+            element: child,
+            layoutAttributes: childAttributes,
+            environment: environment,
+            children: content.performCaffeinatedLayout(
+                frame: frame,
+                context: context
+            )
+        )
+
+        return [(identifier, node)]
     }
 
     private func adapted(environment: Environment) -> Environment {
--- a/BlueprintUI/Sources/Internal/LayoutTreeNode.swift
+++ b/BlueprintUI/Sources/Internal/LayoutTreeNode.swift
@@ -1,0 +1,72 @@
+import CoreGraphics
+import Foundation
+
+/// A node in the layout tree.
+///
+/// This tree is created during layout and measurement, and is used to hold information that may be
+/// reused if a node is visited multiple times, such as measurement caches.
+///
+/// In the layout tree, each edge is keyed by an `ElementIdentifier`, and each node corresponds to
+/// an `ElementContent`. An `ElementPath` can uniquely identify a node from the root. Elements that
+/// lazily generate subelements, such as `GeometryReader`, may create subelements with distinct
+/// identifiers across visits during a single render, which will result in spurious nodes being
+/// created that do not correspond to `LayoutResultNode` in the final layout. However, the
+/// `LayoutResultNode` tree will be an isomorphic subgraph of this tree (every `LayoutResultNode`
+/// corresponds to a `LayoutTreeNode`, but not every `LayoutTreeNode` produces a
+/// `LayoutResultNode`).
+///
+final class LayoutTreeNode {
+
+    typealias Subnode = LayoutTreeNode
+    typealias SubnodeKey = ElementIdentifier
+
+    private var subnodes: [SubnodeKey: Subnode] = [:]
+
+    let path: String
+    let sizeCache: HintingSizeCache
+
+    // These commonly used properties are strongly typed. If we need to hang more generalized
+    // things off this type we may want to store them in a heterogeneous dictionary.
+    private var _layoutSubelements: [LayoutSubelement]?
+    private var _associatedCache: Any?
+
+    init(path: String, signpostRef: AnyObject, options: LayoutOptions) {
+        self.path = path
+        sizeCache = HintingSizeCache(path: path, signpostRef: signpostRef, options: options)
+    }
+
+    func subnode(key: SubnodeKey) -> Subnode {
+        if let subnode = subnodes[key] {
+            return subnode
+        }
+        let subnode = Subnode(
+            path: "\(path)/\(key)",
+            signpostRef: sizeCache.signpostRef,
+            options: sizeCache.options
+        )
+        subnodes[key] = subnode
+        return subnode
+    }
+
+    func layoutSubelements(create: () -> [LayoutSubelement]) -> [LayoutSubelement] {
+        if let layoutSubelements = _layoutSubelements {
+            return layoutSubelements
+        }
+        let layoutSubelements = create()
+        _layoutSubelements = layoutSubelements
+        return layoutSubelements
+    }
+
+    func associatedCache<AssociatedCache>(create: () -> AssociatedCache) -> AssociatedCache {
+        if let associatedCache = _associatedCache as? AssociatedCache {
+            return associatedCache
+        }
+        let associatedCache = create()
+        _associatedCache = associatedCache
+        return associatedCache
+    }
+
+    func update<AssociatedCache>(associatedCache: AssociatedCache) {
+        _associatedCache = associatedCache
+    }
+}
--- a/BlueprintUI/Sources/Internal/Logger.swift
+++ b/BlueprintUI/Sources/Internal/Logger.swift
@@ -74,7 +74,7 @@ extension Logger {
 
     static func logSizeThatFitsStart(
         view: BlueprintView,
-        description: String
+        description: @autoclosure () -> String
     ) {
         guard BlueprintLogging.isEnabled else { return }
 
@@ -84,7 +84,7 @@ extension Logger {
             name: "View Sizing",
             signpostID: OSSignpostID(log: .active, object: view),
             "%{public}s",
-            description
+            description()
         )
     }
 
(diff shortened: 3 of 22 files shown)


BlueprintUI/Sources/Element/EnvironmentAdaptingStorage.swift
```diff
@@ -10,12 +13,19 @@ struct EnvironmentAdaptingStorage: ContentStorage {
     var adapter: (inout Environment) -> Void
 
     var child: Element
+    var content: ElementContent
```
kyleve, line 16: I'm not 100% sure we can just grab this and cache it all the time, eg some elements, like `GeometryReader`, can change the end result of the content based on the size, or others, like `EnvironmentReader`, based on `Environment`. In the caching branch [I'd introduced `isDynamicElement`](https://github.com/square/Blueprint/pull/418/files#diff-07c71c5dcdc5da4c2ddf51b560695e562af396d4afafe3908b35656a0907459aR42) (which would then disable caching of `content`) to fix this.

BlueprintUI/Sources/Internal/LayoutTreeNode.swift
```diff
@@ -0,0 +1,72 @@
...
+        if let layoutSubelements = _layoutSubelements {
+            return layoutSubelements
+        }
+        let layoutSubelements = create()
+        _layoutSubelements = layoutSubelements
+        return layoutSubelements
+    }
+
+    func associatedCache<AssociatedCache>(create: () -> AssociatedCache) -> AssociatedCache {
+        if let associatedCache = _associatedCache as? AssociatedCache {
```
kyleve, line 61: Should this be an `as!`, in case the types get mixed up somehow (eg, an internal bug)? Eg, I don't think the `AssociatedCache` type can/should change per element type.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -1,8 +1,189 @@
...
+    typealias SizeCache = HintingSizeCache
+
+    var identifier: ElementIdentifier
+    private var content: ElementContent
+    var measureContext: MeasureContext
+
+    var environment: Environment { measureContext.environment }
+    private var cache: HintingSizeCache { measureContext.node.sizeCache }
+
+    @Storage
```
kyleve, line 31: Probably for later; but I'd be curious on if there's any performance benefit to the property wrappers (which are each their own instance), vs. converting the `LayoutSubelement` itself into a `final class`.

BlueprintUI/Sources/Layout/UnitPoint.swift
```diff
@@ -0,0 +1,84 @@
...
+///
+/// You can create custom unit points with explicit values, like the example above, or you can use
+/// one of the built-in unit points, like ``zero``, ``center``, or ``topTrailing``. The built-in
+/// values correspond to the alignment positions of the similarly named, built-in ``Alignment``
+/// types.
+///
+/// - Note: A unit point with one or more components outside the range `[0, 1]` projects to a point
+///   outside of the element.
+///
+public struct UnitPoint: Hashable {
```
kyleve, line 37: Maybe its just me, but I found this name somewhat confusing, perhaps `FractionPoint` or similar?

kyleve (approved):

BlueprintUI/Sources/Internal/LayoutTreeNode.swift
```diff
@@ -0,0 +1,72 @@
...
+
+/// A node in the layout tree.
+///
+/// This tree is created during layout and measurement, and is used to hold information that may be
+/// reused if a node is visited multiple times, such as measurement caches.
+///
+/// In the layout tree, each edge is keyed by an `ElementIdentifier`, and each node corresponds to
+/// an `ElementContent`. An `ElementPath` can uniquely identify a node from the root. Elements that
+/// lazily generate subelements, such as `GeometryReader`, may create subelements with distinct
+/// identifiers across visits during a single render, which will result in spurious nodes being
```
nononoah, line 12: Spurious layout tree nodes?

BlueprintUI/Sources/Internal/LayoutTreeNode.swift
```diff
@@ -0,0 +1,72 @@
...
+/// created that do not correspond to `LayoutResultNode` in the final layout. However, the
+/// `LayoutResultNode` tree will be an isomorphic subgraph of this tree (every `LayoutResultNode`
```
nononoah, line 14: Are you saying here that a `LayoutTreeNode` and its subgraph will be isomorphic to its `LayoutResultNode` and its subgraph, or does that `LayoutResultNode` only contain the subgraph?

BlueprintUI/Sources/Internal/LayoutTreeNode.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    private var subnodes: [SubnodeKey: Subnode] = [:]
+
+    let path: String
+    let sizeCache: HintingSizeCache
+
+    // These commonly used properties are strongly typed. If we need to hang more generalized
+    // things off this type we may want to store them in a heterogeneous dictionary.
+    private var _layoutSubelements: [LayoutSubelement]?
+    private var _associatedCache: Any?
```
nononoah, line 31: Sorry, is this strongly typed? Even if the underlying access is in practice, I wonder if the `Any?` box storage incurs a bit of a cost.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -1,8 +1,189 @@
...
+    public func place(
+        filling size: CGSize
+    ) {
+        place(at: .zero, size: size)
+    }
+
+    /// Asks the subelement for its size.
+    ///
+    /// In Blueprint, elements are ultimately sized by their parents, but you can use this method to
+    /// determine the size that a subelement would prefer to choose.
```
nononoah, line 116:
```suggestion
    /// determine the size that a subelement would prefer.
```

BlueprintUI/Sources/Layout/UnitPoint.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// The normalized distance from the origin to the point in the horizontal direction.
+    public var x: CGFloat
+    /// The normalized distance from the origin to the point in the vertical direction.
+    public var y: CGFloat
+
+    /// Creates a unit point at the origin.
+    public init() {
+        self.init(x: 0, y: 0)
+    }
```
nononoah, line 47: I'm a bit confused as to why we have this instead of default parameters or omitting it altogether, especially since we don't use it ourselves in the `zero` case.

BlueprintUI/Sources/Layout/UnitPoint.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// The origin of an element.
+    public static let zero: UnitPoint = .init(x: 0, y: 0)
+
+    /// A point that’s centered in an element.
+    public static let center: UnitPoint = .init(x: 0.5, y: 0.5)
+
+    /// A point that’s centered vertically on the leading edge of an element.
+    public static let leading: UnitPoint = .init(x: 0, y: 0.5)
```
nononoah, line 62: 🐴 

I find this read of `leading` to be surprising at first in that it implicitly contains "and centered vertically." I don't think it is  necessarily an intuitive choice. 

We make the effort to apply `top` and `bottom` prefixes to the other cases, so maybe it's worth doing here too.

nononoah (approved):
Apologies for the excess of comments once more. There's nothing really substantive in my feedback, and it is clear that you and Westin have spent significant time getting this right.

BlueprintUI/Sources/Element/EnvironmentAdaptingStorage.swift
```diff
@@ -56,6 +66,44 @@ struct EnvironmentAdaptingStorage: ContentStorage {
...
     }
 
+    func sizeThatFits(proposal: SizeConstraint, context: MeasureContext) -> CGSize {
+        let environment = adapted(environment: context.environment)
+        let identifier = ElementIdentifier(elementType: type(of: child), key: nil, count: 1)
+        let context = MeasureContext(
+            environment: environment,
+            node: context.node.subnode(key: identifier)
+        )
+        return child.content.sizeThatFits(proposal: proposal, context: context)
```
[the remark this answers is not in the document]
robmaceachern: I'd like to understand this too (and probably worth a comment if `content` shouldn't be used in `sizeThatFits`).

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -1,8 +1,189 @@
...
+
+    @Storage
+    private(set) var placement: Placement?
+
+    /// Optional attributes to apply to this subelement, such as opacity and transforms.
+    @Storage
+    public var attributes = Attributes()
+
+    // Once we are able to fully deprecate the old layout API we can remove the `LayoutType.Traits`
+    // and the `traits` parameter, and instead rely on setting `LayoutValueKey`s where needed.
```
n8chur, line 39: Are we still planning on doing this?

BlueprintUI/Sources/Layout/UnitPoint.swift
n8chur: I believe this is inspired by [SwiftUI's UnitPoint](https://developer.apple.com/documentation/swiftui/unitpoint).

BlueprintUI/Sources/Layout/UnitPoint.swift
n8chur: I'm personally fine with it like this sync it matches [SwiftUI's UnitPoint](https://developer.apple.com/documentation/swiftui/unitpoint).

n8chur (approved):

BlueprintUI/Sources/Layout/UnitPoint.swift
robmaceachern, line 1: Is this 1:1 with SwiftUI's [UnitPoint](https://developer.apple.com/documentation/swiftui/unitpoint)? Should we consider using that impl directly?

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -1,8 +1,189 @@
...
+    /// - Parameters:
+    ///   - position: The place where the anchor of the subelement should appear in its container,
+    ///     relative to the container's bounds.
+    ///   - anchor: The unit point on the subelement that appears at `position`. You can use a
+    ///     built-in point, like ``UnitPoint/center``, or you can create a custom ``UnitPoint``.
+    ///   - size: The size of the subelement. In Blueprint, parents choose their children's size.
+    ///     You can determine a good size for a subelement by calling ``sizeThatFits(_:)`` on it.
+    public func place(
+        at position: CGPoint,
+        anchor: UnitPoint = .topLeading,
```
robmaceachern, line 78: Is `UnitPoint` used in the SwiftUI placement APIs?

robmaceachern (approved):

BlueprintUI/Sources/Internal/Logger.swift
```diff
@@ -131,6 +131,60 @@ extension Logger {
         )
     }
 
+    static func logCacheHit(object: AnyObject, description: @autoclosure () -> String, constraint: SizeConstraint) {
```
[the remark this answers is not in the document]
watt: Yeah, for sure, I thought about wiring up all the Logger methods to something besides `os_signpost` in order to collect data for analysis outside of Instruments, but didn't get to it. This can get hit a lot of times very frequently during a render pass, so collecting this data in production might be impractical due to the perf hit.

BlueprintUI/Sources/Internal/Logger.swift
```diff
@@ -131,6 +131,60 @@ extension Logger {
...
+            name: "Cache miss",
+            signpostID: OSSignpostID(log: .active, object: object),
+            "%{public}s in %{public}s×%{public}s",
+            description(),
+            String(format: "%.1f", constraint.width.constrainedValue ?? .infinity),
+            String(format: "%.1f", constraint.height.constrainedValue ?? .infinity)
+        )
+    }
+
+    static func logCacheUnconstrainedSearchMatch(
```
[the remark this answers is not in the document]
watt: Nope, it gets wired up in #445.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
watt: Assuming `LegacyLayout` eventually goes away, we _could_ do this, and I'd like it, but I'm not in a hurry. I can just remove this comment for now.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
```diff
@@ -1,8 +1,189 @@
...
+    /// arranged by the layout. Provide a position within the container’s bounds where the
+    /// subelement should appear, an anchor that indicates which part of the subelement appears at
+    /// that point, and a size.
+    ///
+    /// To learn the subelement's preferred size for a given proposal before calling this method,
+    /// you can call ``sizeThatFits(_:)`` method on the subelement.
+    ///
+    /// If you call this method more than once for a subelement, the last call takes precedence. If
+    /// you don’t call this method for a subelement, the subelement fills the bounds of its
+    /// container.
```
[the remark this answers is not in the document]
watt: You could make that argument, but this is consistent with SwiftUI.

BlueprintUI/Sources/Layout/LayoutSubelement.swift
watt: > Is `UnitPoint` used in the SwiftUI placement APIs?

[Yes](https://developer.apple.com/documentation/swiftui/layoutsubview/place(at:anchor:proposal:))

BlueprintUI/Sources/Layout/UnitPoint.swift
watt: > Is this 1:1 with SwiftUI's [UnitPoint](https://developer.apple.com/documentation/swiftui/unitpoint)? 

That's the intent.

> Should we consider using that impl directly?

We don't depend on SwiftUI today (except optionally for previews), and I'd rather not introduce that dependency if we can avoid it.

BlueprintUI/Sources/Layout/UnitPoint.swift
watt: Yep, taken directly from SwiftUI. I assume the name has to do with the idea that this is a point on [the unit square](https://en.wikipedia.org/wiki/Unit_square).

BlueprintUI/Sources/Layout/UnitPoint.swift
watt: I'll omit.

BlueprintUI/Sources/Element/EnvironmentAdaptingStorage.swift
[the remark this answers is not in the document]
kyleve: So @kdaker has a good demo of how this can fail (can you link to your branch? I can't find it), but it's not so much about grabbing the content of the GeometryReader itself, but what _it_ returns. Again, I'm not 100% sure if this happens here, but I'll describe the issue we saw:

1) A `GeometryReader` which forks behavior based on layout size (you can imagine other similar cases), eg:
```swift
GeometryReader { geometry in
   if geometry.width < 200 {
      return Label(text: "A shorter string")
   } else {
      return Label(text: "A much longer string that is longer")
   }
}
```

2) The cache caches the first fetch of the label & its content under the `ElementIdentifier` `Label.1`. This is usually the `else` case in the above during measurement.

3) Later fetches / accesses do not update the underlying `content`, meaning if the final layout hits the primary `if` case, the wrong content is returned.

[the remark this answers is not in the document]
kyleve: Here's the code he was using: 

```swift
final class PlaygroundViewController: UIViewController {
    let blueprintView = BlueprintView()
    
    lazy var model: [Element] = [geometryElement, geometryElement, geometryElement]
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        self.view = blueprintView
        
        update()
    }
    
    var geometryElement: Element {
        GeometryReader { proxy in
            return Row { row in
                if proxy.constraint.width.maximum >= 60 {
                    row.addFixed(child: Label(text: String("Big label")).box(background: .blue))
                } else {
                    row.addFixed(child: Label(text: String("Small label")).box(background: .green))
                }
                
            }
        }
    }
    
    func update() {
        blueprintView.element = element(for: model)
    }
    
    func element(for model: [Element]) -> Element {
        Row { row in
            for item in self.model {
                let itemElement = Box(
                    backgroundColor: .red,
                    wrapping: item
                ).tappable {
                    self.model.append(self.geometryElement)
                    self.update()
                }
                row.add(child: itemElement)
            }
        }
        .constrainedTo(size: .init(width: 400, height: 300))
        .centered()
    }
}
```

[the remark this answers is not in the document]
watt: I see what you're saying. I don't think the `content` stored here will cause a problem, but we could have a similar issue with the `LayoutTreeNode` that gets passed to the lazily created child, since it can have cached things hanging off it (i.e. `LayoutSubelement`). I'll set up a regression test like this to make sure we're not breaking anything.

PR #444, opened 2023-03-10
5) Element implementations

watt:
This is PR 5 in a stack targeting `feature/caffeinated-layout`. You can see how it fits into the larger picture by looking at the staging branch in #434.

This PR provides the implementations for built-in Elements, removing the default conformances from `Layout` and `SingleChildLayout`. Most of these are trivial; they either re-use the previous implementation or duplicate.

### DIFF
--- a/BlueprintUI/Sources/Layout/Aligned.swift
+++ b/BlueprintUI/Sources/Layout/Aligned.swift
@@ -113,8 +113,70 @@ public struct Aligned: Element {
 
             return attributes
         }
+
+        func sizeThatFits(proposal: SizeConstraint, subelement: LayoutSubelement, cache: inout Cache) -> CGSize {
+            subelement.sizeThatFits(proposal)
+        }
+
+        func placeSubelement(in size: CGSize, subelement: LayoutSubelement, cache: inout ()) {
+            let x: CGFloat
+            let y: CGFloat
+
+            let subelementSize = subelement.sizeThatFits(SizeConstraint(size))
+            let clampedSize = subelementSize.upperBounded(maxWidth: size.width, maxHeight: size.height)
+
+            let width: CGFloat
+            let height: CGFloat
+
+            switch horizontalAlignment {
+            case .leading:
+                x = 0
+                width = clampedSize.width
+            case .center:
+                x = 0.5
+                width = clampedSize.width
+            case .trailing:
+                x = 1
+                width = clampedSize.width
+            case .fill:
+                x = 0
+                width = size.width
+            }
+
+            switch verticalAlignment {
+            case .top:
+                y = 0
+                height = clampedSize.height
+            case .center:
+                y = 0.5
+                height = clampedSize.height
+            case .bottom:
+                y = 1
+                height = clampedSize.height
+            case .fill:
+                y = 0
+                height = size.height
+            }
+
+            let position = CGPoint(x: x * size.width, y: y * size.height)
+
+            subelement.place(
+                at: position,
+                anchor: UnitPoint(x: x, y: y),
+                size: CGSize(width: width, height: height)
+            )
+        }
+    }
+
+
+}
+
+extension CGSize {
+    func upperBounded(maxWidth: CGFloat, maxHeight: CGFloat) -> CGSize {
+        CGSize(width: min(width, maxWidth), height: min(height, maxHeight))
     }
 }
+
 
 extension Element {
     /// Wraps the element in an `Aligned` element with the provided parameters.
--- a/BlueprintUI/Sources/Layout/EqualStack.swift
+++ b/BlueprintUI/Sources/Layout/EqualStack.swift
@@ -153,6 +153,75 @@ extension EqualStack {
             return result
         }
 
+        func sizeThatFits(proposal: SizeConstraint, subelements: Subelements, cache: inout Cache) -> CGSize {
+            guard subelements.count > 0 else { return .zero }
+
+            let totalSpacing = (spacing * CGFloat(subelements.count - 1))
+            let itemProposal: SizeConstraint
+            switch direction {
+            case .horizontal:
+                itemProposal = .init(
+                    width: proposal.width.map { ($0 - totalSpacing) / CGFloat(subelements.count) },
+                    height: proposal.height
+                )
+            case .vertical:
+                itemProposal = .init(
+                    width: proposal.width,
+                    height: proposal.height.map { ($0 - totalSpacing) / CGFloat(subelements.count) }
+                )
+            }
+            let itemSizes = subelements.map { $0.sizeThatFits(itemProposal) }
+
+            let maximumItemWidth = itemSizes.map { $0.width }.max() ?? 0
+            let maximumItemHeight = itemSizes.map { $0.height }.max() ?? 0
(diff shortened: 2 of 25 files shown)


BlueprintUI/Sources/Layout/Aligned.swift
```diff
@@ -113,9 +113,71 @@ public struct Aligned: Element {
...
+                height = clampedSize.height
+            case .bottom:
+                y = 1
+                height = clampedSize.height
+            case .fill:
+                y = 0
+                height = size.height
+            }
+
+            let position = CGPoint(x: x * size.width, y: y * size.height)
```
kyleve, line 161: For my own education / confirmation: This works because of the `UnitPoint`, right? Eg I am thinking through:

- I have an element of width 50
- It's inside an `Aligned` of width 200
- I align `.center`
- `position.x` will then be `25`

I'd actually expect this to be `(200 - 50) / 2 == 75`

But the `UnitPoint` ends up forcing the right positioning?

BlueprintUI/Sources/Layout/EqualStack.swift
```diff
@@ -153,6 +153,75 @@ extension EqualStack {
...
+                    height: size.height
+                )
+            case .vertical:
+                itemSize = CGSize(
+                    width: size.width,
+                    height: (size.height - totalSpacing) / CGFloat(subelements.count)
+                )
+            }
+
+            for (subelement, index) in zip(subelements, 0...) {
```
kyleve, line 213: Noting: As we've seen in the past, might be faster to use `indexedForEach` here.

BlueprintUI/Sources/Layout/Hidden.swift
```diff
@@ -32,6 +32,14 @@ public struct Hidden: Element {
             attributes.isHidden = isHidden
             return attributes
         }
+
+        func sizeThatFits(proposal: SizeConstraint, subelement: LayoutSubelement, cache: inout Cache) -> CGSize {
+            subelement.sizeThatFits(proposal)
+        }
+
+        func placeSubelement(in size: CGSize, subelement: LayoutSubelement, cache: inout ()) {
```
kyleve, line 40: This automatically sets the frame of the child, so we don't need to set it here, right?

BlueprintUI/Sources/Layout/LayoutWriter.swift
```diff
@@ -272,6 +272,19 @@ extension LayoutWriter {
                     .init(frame: child.frame)
                 }
             }
+
+            func sizeThatFits(proposal: SizeConstraint, subelements: Subelements, cache: inout Cache) -> CGSize {
+                builder.sizing.measure(with: builder)
+            }
+
+            func placeSubelements(in size: CGSize, subelements: Subelements, cache: inout ()) {
+                for (child, subelement) in zip(builder.children, subelements) {
```
kyleve, line 281: Same comment re: Perhaps using `indexedForEach` for performance.

kyleve (approved):

BlueprintUI/Sources/Layout/Aligned.swift
```diff
@@ -113,9 +113,71 @@ public struct Aligned: Element {
...
+                size: CGSize(width: width, height: height)
+            )
+        }
     }
+
+
 }
 
+extension CGSize {
+    func upperBounded(maxWidth: CGFloat, maxHeight: CGFloat) -> CGSize {
```
n8chur, line 175: `fileprivate` or move to our other SPI'd conveniences?

BlueprintUI/Sources/Layout/Hidden.swift
n8chur: Yeah the default is to fill the parent's frame.

n8chur (approved):

BlueprintUI/Sources/Layout/Aligned.swift
robmaceachern: I think `position.x` is 100 in your example. Since `x` is 0.5, the horizontal center of the subelement will be positioned at x: 100 (the middle of the `Aligned`), which is correct. At least that's how I understand it.

nononoah (approved):

BlueprintUI/Sources/Layout/GridRow.swift
```diff
@@ -120,6 +120,52 @@ extension GridRow {
...
+            }
+
+            let items = items(subelements: subelements)
+
+            let frames = _frames(in: proposal, items: items)
+
+            // Measure the the row to be as wide as the sum of its children and as tall as its tallest child.
+            let size = frames.reduce(.zero) { size, frame in
+                CGSize(width: frame.maxX, height: max(size.height, frame.height))
+            }
```
robmaceachern, line 145: I was expecting `width: max(size.width, frame.maxX)`.

Does the `_frames` call guarantee that the right-most frame will be the last one?

(aside: I guess we could keep track of the max height within `_frames` and return it since we're already iterating over them there. Then the size calculation avoids iterating: `let size = CGSize(width: value.frames.last!.maxX, height: value.maxHeight)`. Assuming we're comfortable with relying on the last frame for width.)

BlueprintUI/Sources/Layout/Inset.swift
```diff
@@ -158,5 +158,34 @@ extension Inset {
             frame.size.height -= top + bottom
             return LayoutAttributes(frame: frame)
         }
+
+        func sizeThatFits(proposal: SizeConstraint, subelement: LayoutSubelement, cache: inout ()) -> CGSize {
+            let insetProposal = proposal.inset(by: edgeInsets)
+            let childSize = subelement.sizeThatFits(insetProposal)
+            return childSize + CGSize(width: left + right, height: top + bottom)
```
robmaceachern, line 165: Do you think there's any tangible performance gain if we avoid the CGSize arithmetic convenience (since it results in two `CGSize`s being initialized and a function call)?

robmaceachern (approved):

BlueprintUI/Sources/Layout/EqualStack.swift
kyleve: We've seen that any use of zip or enumerated is slower than our indexed helpers, likely due to allocations happening (even if on the stack) and extra internal overhead.

robmaceachern: Interesting! I wasn't seeing any significant differences in my quick profiling (release mode on Mac) of these examples:

<details>

```
let values = repeatElement("Hello", count: 10_000_000_000)

var stored: (String, Int) = ("", -1)

func testZip() {
    for (item, offset) in zip(values, 0...) {
        stored = (item, offset)
    }
}

func testForIn() {
    for item in values {
        stored = (item, -1)
    }
}

func testEnumerated() {
    for (offset, item) in values.enumerated() {
        stored = (item, offset)
    }
}

//testZip() // 22.03s
//testForIn() // 22.04s
testEnumerated() // 22.06s
```

</details>

[enumerated-vs-zip-vs-forIn.trace.zip](https://github.com/square/Blueprint/files/11104102/enumerated-vs-zip-vs-forIn.trace.zip)

PR #447, opened 2023-03-15
Feature: Caffeinated Layout

watt:
This is a feature branch collecting work on Caffeinated Layout.

1. #440 
2. #441 
3. #442 
4. #443 
5. #444 
6. #445 
7. #446 
8. #450 
9. #451 

For a big picture view check out #434 , which contains the first 6 PRs in a single branch.

PR #450, opened 2023-04-11
8) Infinite size elements

watt:
This is PR 8 in a stack targeting `feature/caffeinated-layout`. See #447.

This PR updates existing elements that don't fully adhere to the Caffeinated Layout contract, `Image` and `TextField`.

Additionally, `ScrollView` has been updated to protect against placing elements into infinite bounds. `ScrollView` is uniquely vulnerable to this because it introduces `unconstrained` constraints to its subelements. Most elements don't need to worry about this, but for any that do, a new convenience for replacing infinite values on `CGSize` can help.


n8chur (approved):

BlueprintUI/Sources/Extensions/CGSize+Blueprint.swift
```diff
@@ -0,0 +1,16 @@
+import CoreGraphics
+
+extension CGSize {
+    /// A size with `infinity` in both dimensions.
+    public static let infinity = CGSize(width: CGFloat.infinity, height: .infinity)
```
nononoah, line 5: One has to love it.

nononoah (approved):

BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -137,7 +140,12 @@ extension Image {
                 } else if case .atMost(let height) = constraint.height {
                     mode = .fitHeight(height)
                 } else {
-                    mode = .useImageSize
+                    switch layoutMode {
+                    case .legacy:
+                        mode = .useImageSize
+                    case .caffeinated:
+                        mode = .infinite
```
nononoah, line 147: Could you unpack this a bit for me?

Given an unconstrained measurement space, `Image` now requests infinite layout space. Does that not change the layout behavior of images in these contexts, and therefore any container of these images?

watt: > Given an unconstrained measurement space, `Image` now requests infinite layout space. Does that not change the layout behavior of images in these contexts, and therefore any container of these images?

It does! Practically speaking, there's not that much impact, because the need to measure an element fully unconstrained is somewhat rare.

There's perhaps a semantic change — if you were counting on unconstrained measurement to get you something like an intrinsic size, or SwiftUI's concept of "ideal size", that's no longer possible. But it was never guaranteed to work that way, and would be inconsistent depending on what you measured.

> Or is it that `.infinite` opts them out of caffeinated layout in a way that leads to the restoration of existing behavior?

Nope, there's no way to opt out at the element level.

kyleve: Is this because of the cache hinting optimizations? I guess I would expect for the measurement modes that don't stretch the image, I'd expect to return the size of image, even in an infinite constraint.

watt: > Is this because of the cache hinting optimizations? 

It's required by the new layout contract, yeah.

> I guess I would expect for the measurement modes that don't stretch the image, I'd expect to return the size of image, even in an infinite constraint.

Well, the aspect fit & fill modes already fill space today, even under finite constraints. Given one finite constraint, they'll fill it, no matter how large. So the reasoning is that if the size grows continuously with the constraint, why wouldn't it grow to infinity when the constraint grows to infinity?

We could change the behavior of these modes to return the size of the image without filling space, and that would be compliant too, but it would be a much bigger breaking change to existing layouts.

PR #451, opened 2023-04-11
9) Add Environment to API

watt:
This is PR 9 in a stack targeting `feature/caffeinated-layout`. See #447 

This PR adds an `Environment` parameter to the `CaffeinatedLayout` API methods.

It also removes the `LayoutContext` and `MeasureContext` types that were used internally, flattening them into two parameters in all usages.

### DIFF
--- a/BlueprintUI/Sources/Element/ContentStorage.swift
+++ b/BlueprintUI/Sources/Element/ContentStorage.swift
@@ -22,11 +22,13 @@ protocol LegacyContentStorage {
 protocol CaffeinatedContentStorage {
     func sizeThatFits(
         proposal: SizeConstraint,
-        context: MeasureContext
+        environment: Environment,
+        node: LayoutTreeNode
     ) -> CGSize
 
     func performCaffeinatedLayout(
         frame: CGRect,
-        context: LayoutContext
+        environment: Environment,
+        node: LayoutTreeNode
     ) -> [ElementContent.IdentifiedNode]
 }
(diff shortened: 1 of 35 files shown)


BlueprintUI/Sources/Element/ContentStorage.swift
kyleve, line 1: > also removes the LayoutContext and MeasureContext types that were used internally, flattening them into two parameters in all usages.

Are we 100% sure we want to remove these context objects? The benefit is that we can add more parameters without breaking API changes in the future, because the new parameters would go into the context.

kyleve: (I guess you could argue, the Environment becomes that object that we add parameters to – for some reason that's always felt odd to me, but I suppose it's fine?)

kyleve (approved):

n8chur (approved):

nononoah (approved):

BlueprintUI/Sources/Element/ContentStorage.swift
watt: > Are we 100% sure we want to remove these context objects? The benefit is that we can add more parameters without breaking API changes in the future, because the new parameters would go into the context.

I am sure we want to remove them from the internal `ContentStorage` API where they are now. They're adding some boxing overhead, and it wouldn't be a breaking change to alter that API later.

But I think you're asking whether we might want to add a different context object to the public API. As you've mentioned, the benefit there is future additions wouldn't be breaking. The downside is that it's a worse API today.

My opinion is that I'd rather have the good API today. The upside is too hypothetical: I tried and could not come up anything I would anticipate us adding here any time soon. Even if I could snap my fingers and have dream features like state I'd rather see them go into something like property wrappers anyway.

PR #454, opened 2023-04-26
Fix exact match case of constrained aspect ratios and contract adherence

watt:
There are two changes in this PR, and it may be easier to review by commit.

- A bugfix for the `fitParent` and `fillParent` modes of `ConstrainedAspectRatio`, where the measuring was falling through to the wrong value when the aspect ratio of the container matched the aspect ratio being constrained to.
- Some changes to the behavior when unconstrained to adhere to Caffeinated Layout. The `fillParent` behavior has very limited utility in Caffeinated Layout since it will aggressively fill and overflow any container. I can't think of any practical use case, and I also couldn't find any usage today, so I am proposing we deprecate it.

To validate the bugfix I wound up almost entirely rewriting the test class, so that diff is a mess.

### DIFF
--- a/BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
+++ b/BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
@@ -4,20 +4,32 @@ import UIKit
 public struct ConstrainedAspectRatio: Element {
     /// Represents how the content should size itself relative to its parent.
     public enum ContentMode: Equatable {
-        /// The content will be sized to fill its parent while maintaining the specified aspect
-        /// ratio.
+        /// The content will be sized to fill its parent while
+        /// maintaining the specified aspect ratio.
         ///
-        /// If the parent is unconstrained in all dimensions, the content size will be used for
-        /// measurement and will behave like `fitContent`. If the parent is unconstrained in one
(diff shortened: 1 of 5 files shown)


watt:
> Is there an integration PR (both CL on and off)? How's that looking in terms of breaking changes?

Only with CL off. It went green.

BlueprintUI/Sources/Layout/ConstrainedAspectRatio.swift
```diff
@@ -66,13 +78,14 @@ public struct ConstrainedAspectRatio: Element {
...
                 } else if constraint.width.constrainedValue == nil {
                     return aspectRatio.size(forWidth: constrainedWidth)
                 } else if constraint.height.constrainedValue == nil {
                     return aspectRatio.size(forHeight: constrainedHeight)
                 } else if constrainedWidth < availableWidth {
                     return aspectRatio.size(forHeight: constrainedHeight)
-                } else if constrainedHeight < availableHeight {
+                } else {
                     return aspectRatio.size(forWidth: constrainedWidth)
                 }
```
watt, line 90: When the constraint matched the available size we were falling all the way out of this switch and returning `contentSize` at the bottom of the method. I removed that fallback and made every switch case return something appropriate.

PR #459, opened 2023-05-30
AttributedLabel Link Accessibility

RoyalPineapple:
Updating attributed label's links to more closely mirror the SwiftUI accessibility experience

https://github.com/square/Blueprint/assets/219578/7255cbc0-2c6a-41a5-a0bc-9cd5ec5774c4

This is our current implementation, which doesn't announce links properly. 

https://github.com/square/Blueprint/assets/219578/11c3e109-c632-4542-9e5b-11eddf618eba

Here is the new version, announcing links as they are read. 

https://github.com/square/Blueprint/assets/219578/38aad435-825b-42f4-b65f-a70b5757691a

### DIFF
--- a/BlueprintUI/Sources/Environment/Keys/AccessibilityLinkKey.swift
+++ b/BlueprintUI/Sources/Environment/Keys/AccessibilityLinkKey.swift
@@ -1,0 +1,17 @@
+import UIKit
+
+extension Environment {
+    private enum LinkAccesibilityLabelKey: EnvironmentKey {
+        static var defaultValue: String? {
+            UIImage(systemName: "link")?.accessibilityLabel
+        }
+    }
+
+    /// The localised accessibility label elements should use when handling links.
+    ///
+    /// Defaults to `UIImage(systemName: "link")?.accessibilityLabel`.
+    public var linkAccessibilityLabel: String? {
+        get { self[LinkAccesibilityLabelKey.self] }
+        set { self[LinkAccesibilityLabelKey.self] = newValue }
+    }
+}
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -265,6 +265,11 @@ extension AttributedLabel {
                 if previousAttributedText != attributedText {
                     links = attributedLinks(in: model.attributedText) + detectedDataLinks(in: model.attributedText)
                     accessibilityLinks = accessibilityLinks(for: links, in: model.attributedText)
+                    accessibilityLabel = accessibilityLabel(
+                        with: links,
+                        in: model.attributedText.string,
+                        linkAccessibilityLabel: environment.linkAccessibilityLabel
+                    )
                 }
 
                 if let shadow = model.shadow {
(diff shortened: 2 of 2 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -529,6 +553,17 @@ extension AttributedLabel {
             attributedText = mutableString
         }
 
+        override func accessibilityActivate() -> Bool {
```
RoyalPineapple, line 556: Previously, activating the element would send a touch event to the label's accessibility activation point which defaults to the center of the view. This would occasionally activate a link that happened to overlap that particular position.

We now have the provide behavior:
- No links: Not interactive, no effect.
- Exactly one link: Activate the link.
- More than one link: Ambiguous selection, no effect.
as always we additionally allow link selection and activation through the rotor.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -751,3 +787,11 @@ extension NSTextCheckingResult.CheckingType {
     }
 }
 
+extension AttributedLabel {
```
RoyalPineapple, line 790: This might very well be too clever by half but without adding explicit localization to the project this seems like the most reliable way to get this string. 
We could allow a localized string to be injected into the label but that feels like a slightly different code smell and even so I'd like to offer this as a default to prevent a breaking API change.  

I'm actively looking for thoughts and suggestions on this one.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -264,7 +264,12 @@ extension AttributedLabel {
             if !isMeasuring {
                 if previousAttributedText != attributedText {
                     links = attributedLinks(in: model.attributedText) + detectedDataLinks(in: model.attributedText)
-                    accessibilityLinks = accessibilityLinks(for: links, in: model.attributedText)
+                    accessibilityLinks = accessibilityLinks(
+                        for: links,
+                        in: model.attributedText,
+                        accessibilityValue: environment.linkAccessibility
```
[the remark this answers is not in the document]
RoyalPineapple: it gets set as an `accessibilityValue` on the link element itself, with the link text being set as the `accessibilityLabel`, i wanted to avoid potential confusion around that.

BlueprintUI/Sources/Environment/Keys/AccessibilityLinkKey.swift
```diff
@@ -0,0 +1,17 @@
...
+    private enum AccesibilityLinkKey: EnvironmentKey {
+        static var defaultValue: String? {
+            UIImage(systemName: "link")?.accessibilityLabel
+        }
+    }
+
+    /// The localised accessibility label elements should use when handling links.
+    ///
+    /// Defaults to `UIImage(systemName: "link")?.accessibilityLabel`.
+    public var linkAccessibility: String? {
```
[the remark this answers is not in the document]
RoyalPineapple: It's used as both an `accessibilityLabel` and `accessibilityValue` in different places (inserted into the `accessibilityLabel` for the entire textLabel vs the 'accessibilityValue' on the link element itself) so I wanted to avoid using one or the other explicitly in the naming. 
I thought about `localizedLinkAccessibilityString` but thats a mouthful and imbedding the type in the name seemed like my Objective C accent leaking through.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -310,6 +315,29 @@ extension AttributedLabel {
             }
         }
 
+        private func updateAccessibilityLabel(with links: [Link], localizedLinkLabel: String?) {
```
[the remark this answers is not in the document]
RoyalPineapple: Mostly just separation of concern.`accessibilityLinks(...)` returns an array of link objects and it seemed like a weird side effect to have that func also set the accessibility label of the entire label.

kyleve:
It looks like there's some failing tests:

```
Failing tests:

	AttributedLabelTests.test_linkAttribute()
	AttributedLabelTests.test_textContainerRects()
```

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -514,6 +523,27 @@ extension AttributedLabel {
                         link: link
                     )
                 }
+
+            // Insert the word "link" after each link in the label. This mirrors the VoiceOver utterance of `SwiftUI.Text` when rendering a markdown link.
+            guard var label = accessibilityLabel,
+                  let localizedLinkString = linkAccessibilityLabel,
+                  !links.isEmpty else { return }
+
+            let insertionString = "[\(localizedLinkString)] "
```
kyleve, line 532: Do the `[ ]` have any special meaning here?

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: It would be good to capture this intent in a comment on the method, IMO

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -500,8 +499,18 @@ extension AttributedLabel {
...
         }
 
-        private func accessibilityLinks(for links: [Link], in string: NSAttributedString) -> [LinkAccessibilityElement] {
-            links
+        private func updateAccessibilityLinks(
+            linkAccessibilityLabel: String?
+        ) {
+
+            guard let string = attributedText else {
+                accessibilityLinks = []
```
kyleve, line 507: Do you also need to set `links =` in here?

BlueprintUICommonControls/Sources/AttributedLabel.swift
RoyalPineapple: its really only for the case where the string would be printed. e.g. the accessibility inspector. I wanted to be clear that it's not a part of the displayed text. and that it's explicitly associated with the link string (it doesn't have a preceding space). 

I chose the square brackets because
- Voiceover doesn't read the `[]` characters, but does realize the contained word is distinct from the preceding word. 
- They aren't often used in prose, unlike parenthesis. They're unlikely to be confused with the actual content
- They look like markdown.

RoyalPineapple:
> It looks like there's some failing tests:
> 
> ```
> Failing tests:
> 
> 	AttributedLabelTests.test_linkAttribute()
> 	AttributedLabelTests.test_textContainerRects()
> ```

I had a nasty merge failure and associated cleanup. i bet thats related.

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Makes sense! It might be good to capture that in a comment inline

kyleve (approved):

robmaceachern:
@RoyalPineapple Moving to draft because I think this was addressed elsewhere but feel free to reopen if needed!

meherkasam-square (approved):

PR #461, opened 2023-07-11
fix: [AttributedLabel] Correct typo in enumeration

nononoah:
### Summary

This PR corrects a benign typo in `AttributedLabel`. When tapping a link, `AttributedLabel` attempts to discover alignment of a paragraph style that spans its entire range. By definition, there can only be one such style.

This means that the previous expression would continue enumeration after finding the a style which applies to a subrange, which is unnecessary. It also meant that the stop pointer - erroneously expressed as a continue pointer - did nothing.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -347,6 +347,7 @@ extension AttributedLabel {
 
         private func links(at location: CGPoint) -> [Link] {
             guard let textStorage = makeTextStorage(),
+                  textStorage.string.isEmpty == false,
                   let layoutManager = textStorage.layoutManagers.first,
                   let textContainer = layoutManager.textContainers.first
             else {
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -360,11 +360,11 @@ extension AttributedLabel {
                     .paragraphStyle,
                     in: textStorage.entireRange,
                     options: []
-                ) { style, range, continuePointer in
+                ) { style, range, stop in
                     if let style = style as? NSParagraphStyle, range == textStorage.entireRange {
                         alignment = style.alignment
-                        continuePointer.pointee = false
                     }
+                    stop.pointee = true
```
robmaceachern, line 367: Sorry if I'm being dense, but why are we stopping outside of the conditional?

nononoah: If multiple paragraph styles are set over the label's entire range, we'll never enter the `if` clause, as none of them will be set over the entire range.

This means we can stop after hitting the first paragraph range. If we don't enter the `if` clause, we never will.

watt (approved):

robmaceachern (approved):

PR #463, opened 2023-08-02
Update KeyboardObserver

robmaceachern:
This updates the `KeyboardObserver` to match the more recent implementation we use elsewhere.

Changes include:
- A global `KeyboardObserver`
- Support for the iOS 16.1+ changes related to the screen coordinate space. This can impact reported values when the app isn't full screen in Split View, Slide Over, and Stage Manager.
- The delegate has been changed to report the `UIView.AnimationCurve` instead of `UIView.AnimationOptions`
- `ScrollerWrapperView` changes:
  - The global `KeyboardObserver` is used instead of a per-instance observer. This might actually fix issues if this view is instantiated when the keyboard is already presented.
  - We now use a `UIViewPropertyAnimator` instead of `UIView.animate` when responding to keyboard notifications.

Note: This doesn't add any automatic initialization of the shared observer on launch. It might make sense to do that but I'd like to consider that separate from this impl sync-up.

### DIFF
--- a/BlueprintUICommonControls/Sources/ScrollView.swift
+++ b/BlueprintUICommonControls/Sources/ScrollView.swift
@@ -304,7 +304,7 @@ extension ScrollView {
 fileprivate final class ScrollerWrapperView: UIView {
 
     let scrollView = UIScrollView()
-    let keyboardObserver = KeyboardObserver(center: .default)
+    let keyboardObserver = KeyboardObserver.shared
 
     /// The current `ScrollView` state we represent.
     private var representedElement: ScrollView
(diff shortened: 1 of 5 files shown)


BlueprintUICommonControls/Sources/ScrollView.swift
```diff
@@ -564,11 +564,12 @@ extension ScrollerWrapperView: KeyboardObserverDelegate {
     func keyboardFrameWillChange(
         for observer: KeyboardObserver,
         animationDuration: Double,
-        options: UIView.AnimationOptions
+        animationCurve: UIView.AnimationCurve
     ) {
-        UIView.animate(withDuration: animationDuration, delay: 0.0, options: options, animations: {
+        UIViewPropertyAnimator(duration: animationDuration, curve: animationCurve) {
```
robmaceachern, line 569: We could stick with `UIView.animate` - we'd just need to do a `UIView.AnimationOptions(rawValue: animationCurve << 16)`.

n8chur (approved):

PR #467, opened 2023-09-15
Add `TintAdjustmentMode` to layout attributes

jacksoncheek:
### What changed?
- Added a `TintAdjustmentMode` element and `.tintAdjustmentMode(:)` modifier for finer control of tint color during modal presentations.

### Why?
Whenever a modal is presented, the tint color of `Image` views behind the modal are desaturated/dimmed. Wrapping an element (in this case `Image`) with a `TintAdjustmentMode` gives finer control of that behavior, e.g. if you'd like to prevent that "dimming" from happening.

### DIFF
--- a/BlueprintUI/Sources/Layout/LayoutAttributes.swift
+++ b/BlueprintUI/Sources/Layout/LayoutAttributes.swift
@@ -30,7 +30,7 @@ public struct LayoutAttributes {
     public var isHidden: Bool
 
     /// Corresponds to `UIView.tintAdjustmentMode`.
-    public var tintAdjustmentMode: UIView.TintAdjustmentMode?
+    public var tintAdjustmentMode: UIView.TintAdjustmentMode
 
     public init() {
         self.init(center: .zero, bounds: .zero)
@@ -54,7 +54,7 @@ public struct LayoutAttributes {
         alpha = 1.0
         isUserInteractionEnabled = true
         isHidden = false
-        tintAdjustmentMode = nil
+        tintAdjustmentMode = .automatic
 
         validateBounds()
         validateCenter()
@@ -93,9 +93,7 @@ public struct LayoutAttributes {
         view.alpha = alpha
         view.isUserInteractionEnabled = isUserInteractionEnabled
         view.isHidden = isHidden
-        if let tintAdjustmentMode {
-            view.tintAdjustmentMode = tintAdjustmentMode
-        }
+        view.tintAdjustmentMode = tintAdjustmentMode
     }
 
 
@@ -163,7 +161,15 @@ public struct LayoutAttributes {
         result.alpha = alpha * layoutAttributes.alpha
         result.isUserInteractionEnabled = layoutAttributes.isUserInteractionEnabled && isUserInteractionEnabled
         result.isHidden = layoutAttributes.isHidden || isHidden
-        result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode ?? tintAdjustmentMode
+
+        switch tintAdjustmentMode {
+        case .dimmed, .normal:
+            result.tintAdjustmentMode = tintAdjustmentMode
+        case .automatic:
+            result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode
+        @unknown default:
+            result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode
+        }
 
         return result
     }
--- a/BlueprintUI/Sources/Layout/LayoutAttributes.swift
+++ b/BlueprintUI/Sources/Layout/LayoutAttributes.swift
@@ -29,6 +29,9 @@ public struct LayoutAttributes {
     /// Corresponds to `UIView.isHidden`.
     public var isHidden: Bool
 
+    /// Corresponds to `UIView.tintAdjustmentMode`.
+    public var tintAdjustmentMode: UIView.TintAdjustmentMode?
+
     public init() {
         self.init(center: .zero, bounds: .zero)
     }
@@ -51,6 +54,7 @@ public struct LayoutAttributes {
         alpha = 1.0
         isUserInteractionEnabled = true
         isHidden = false
+        tintAdjustmentMode = nil
 
         validateBounds()
         validateCenter()
@@ -64,6 +68,7 @@ public struct LayoutAttributes {
         alpha = attributes.alpha
         isUserInteractionEnabled = attributes.isUserInteractionEnabled
         isHidden = attributes.isHidden
+        tintAdjustmentMode = attributes.tintAdjustmentMode
     }
 
     public var frame: CGRect {
@@ -88,6 +93,9 @@ public struct LayoutAttributes {
         view.alpha = alpha
         view.isUserInteractionEnabled = isUserInteractionEnabled
         view.isHidden = isHidden
+        if let tintAdjustmentMode {
+            view.tintAdjustmentMode = tintAdjustmentMode
+        }
     }
 
 
@@ -155,6 +163,7 @@ public struct LayoutAttributes {
         result.alpha = alpha * layoutAttributes.alpha
         result.isUserInteractionEnabled = layoutAttributes.isUserInteractionEnabled && isUserInteractionEnabled
         result.isHidden = layoutAttributes.isHidden || isHidden
+        result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode ?? tintAdjustmentMode
 
         return result
     }
@@ -267,6 +276,7 @@ extension LayoutAttributes: Equatable {
             && lhs.alpha == rhs.alpha
             && lhs.isUserInteractionEnabled == rhs.isUserInteractionEnabled
             && lhs.isHidden == rhs.isHidden
+            && lhs.tintAdjustmentMode == rhs.tintAdjustmentMode
     }
 
 }
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -11,12 +11,6 @@ public struct Image: Element {
     /// The tint color.
     public var tintColor: UIColor?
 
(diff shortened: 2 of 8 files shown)


BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -11,6 +11,12 @@ public struct Image: Element {
     /// The tint color.
     public var tintColor: UIColor?
 
+    /// The tint adjustment mode. When this value is `.dimmed`, the view's `tintColor`
+    /// property returns a desaturated, dimmed version of the view's original tint color. When
+    /// this value is `.normal`, the view's `tintColor` property returns the completely
+    /// unmodified tint color.
+    public var tintAdjustmentMode: UIView.TintAdjustmentMode?
```
n8chur, line 18: @square/ui-systems-ios do we think this makes sense to promote to `LayoutAttributes` and add a view modifier for it instead of this being specifically tied to `Image`?

kyleve: I _think_ this is what you're saying, but could we just throw this into the environment and then read it from `ViewDescriptionContext` instead?

I'm not sure this is the right answer, but it does mean we don't need to pipe it through everywhere...

robmaceachern: > do we think this makes sense to promote this to LayoutAttributes and add a view modifier for it instead of this being specifically tied to Image?

I think this makes sense to me if I understand correctly. It'd analogous to `Opacity`/`alpha`, right?

n8chur: Yeah I'm imagining it'd work just like how we handle `Opacity`/`alpha`, `isUserInteractionEnabled`, `isHidden`, etc.

n8chur: I think the type this would be added to in Caffeinated Layout is `LayoutSubelement.Attributes`.

jacksoncheek: Might not fully understand the discussion above, but this is my interpretation of the proposal:
- Add `tintAdjustmentMode: UIView.TintAdjustmentMode` to `LayoutSubelement.Attributes` and `LayoutAttributes`
- Create some new element wrapper like `TintAdjustmentMode` that will conditionally modify the tint adjustment mode of a wrapped element (similar to `Opacity` or `UserInteractionEnabled`)

Then some example usage could be like:
```
let wrappedImage: Element = {
    let image = Image(
        image: config.image,
        contentMode: .aspectFit,
        tintColor: tintColor.color
    )
    .tintAdjustmentMode(.normal) // or .automatic, or .dimmed
    .centered()
}()
```

Yea?

jacksoncheek: I think ⬆️ was the intent so I've put together another un-pushed commit that does this. Happy to go either way on it.

jacksoncheek: Pushed those changes in this commit 7b08f6ea25cbd028ccebd11bc4a10f577a2878e9. We can roll back to the original commit based on further discussion if needed.

BlueprintUI/Sources/Layout/LayoutAttributes.swift
```diff
@@ -88,6 +93,9 @@ public struct LayoutAttributes {
         view.alpha = alpha
         view.isUserInteractionEnabled = isUserInteractionEnabled
         view.isHidden = isHidden
+        if let tintAdjustmentMode {
+            view.tintAdjustmentMode = tintAdjustmentMode
+        }
```
n8chur, line 98: I think we still need to update this if `tintAdjustmentMode` becomes `nil`, but it's unclear what to utilize as the default value, or if that's even appropriate. The docs seem to imply that this will traverse the subview hierarchy to determine a value unless you set this value, but there doesn't appear to be a way to revert it back to the subview traversal behavior. Am I understanding that correctly?

BlueprintUICommonControls/Sources/Image.swift
n8chur: Yup, that's exactly what I was thinking—sorry for the delay!

BlueprintUI/Sources/Layout/LayoutAttributes.swift
n8chur: Or is `automatic` the same as the "default" state?

n8chur: I'm seeing views in the Market Catalog app default to `UIViewTintAdjustmentMode(rawValue: 1)`, where `case normal = 1`, which I wasn't expecting... I don't see us overriding it anywhere.

watt: I've never used this property before, but to me the docs seem to imply that you can set this property to `automatic` to get default behavior, but querying this property will never return a value of `automatic`. So, I think I agree that we'll either need to treat `nil` as `automatic`, or make this attribute non-optional with a default of `automatic`.

BlueprintUI/Sources/Layout/LayoutAttributes.swift
```diff
@@ -155,6 +163,7 @@ public struct LayoutAttributes {
         result.alpha = alpha * layoutAttributes.alpha
         result.isUserInteractionEnabled = layoutAttributes.isUserInteractionEnabled && isUserInteractionEnabled
         result.isHidden = layoutAttributes.isHidden || isHidden
+        result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode ?? tintAdjustmentMode
```
watt, line 166: The cascade behavior can be tricky for these kinds of attributes. Consider this scenario:

```swift
element
  .tintAdjustmentMode(.automatic)
  .tintAdjustmentMode(.dimmed)
```

Would you expect the element's mode to be dimmed, or normal?

BlueprintUI/Sources/Layout/LayoutAttributes.swift
n8chur: That makes sense...

I think I'd lean towards non-optional with a default of `automatic` in that case.

jacksoncheek: Found this in the UIKit docs.
> -tintAdjustmentMode always returns either UIViewTintAdjustmentModeNormal or UIViewTintAdjustmentModeDimmed. The value returned is the first non-default value in the receiver's superview chain (starting with itself).
> If no non-default value is found, UIViewTintAdjustmentModeNormal is returned.
> When tintAdjustmentMode has a value of UIViewTintAdjustmentModeDimmed for a view, the color it returns from tintColor will be modified to give a dimmed appearance.
> When the tintAdjustmentMode of a view changes (either the view's value changing or by one of its superview's values changing), -tintColorDidChange will be called to allow the view to refresh its rendering.

And for `.automatic`: https://developer.apple.com/documentation/uikit/uiview/tintadjustmentmode/automatic
> The tint adjustment mode of the view is the same as its superview’s tint adjustment mode (or UIViewTintAdjustmentModeNormal if the view has no superview).

So I think this means the default value we provide should be `.automatic`. But it's still not clear to me then how the `LayoutAttributes` concatenation should work so we don't mess up the superview's chain of `tintAdjustmentMode` values. IDK. Perhaps we revert this commit and just tie this property specifically to `Image` like the original commit had so the superview chain isn't polluted.

jacksoncheek: `tintAdjustmentMode` can never return `.automatic`

BlueprintUI/Sources/Layout/LayoutAttributes.swift
jacksoncheek: Yeah this is what gives me pause. Lifting this into the `LayoutAttributes` may be dangerous in terms of the cascading affect up the superview chain. https://github.com/square/Blueprint/pull/467#discussion_r1330630096

jacksoncheek: I think I would expect

```
element
  .tintAdjustmentMode(.automatic)
  .tintAdjustmentMode(.dimmed)
```

to have the mode `.dimmed`.

But what about this case? Should the parent fallback to the child's mode if the parent's mode is `.automatic`? I'm not sure. The correct concatenation logic is kind of opaque to me.
```
element
  .tintAdjustmentMode(.normal)
  .tintAdjustmentMode(.automatic)
```

Is it appropriate for this element's mode to be `.normal`? 🤷

watt: > But what about this case? Should the parent fallback to the child's mode if the parent's mode is `.automatic`? I'm not sure. The correct concatenation logic is kind of opaque to me.
> 
> ```
> element
>   .tintAdjustmentMode(.normal)
>   .tintAdjustmentMode(.automatic)
> ```
> 
> Is it appropriate for this element's mode to be `.normal`? 🤷

I think so. I think we can kind of mimic the UIKit behavior here, so that the element inherits the nearest non-default value. So, when both `normal` and `dimmed` are present in the hierarchy, the nearest one wins. And any `automatic` has no real effect.

In that case the cascade would be something like:
```swift
switch tintAdjustmentMode {
  case .normal, .dimmed:
    result.tintAdjustmentMode = tintAdjustmentMode
  case .automatic:
    result.tintAdjustmentMode = layoutAttributes.tintAdjustmentMode
}
```

There's some precedent here in the other properties, I think. You can't reset opacity if a parent element reduces it, for example; you can only further reduce it.

jacksoncheek: Given the precedence, I think this makes sense to me. Updated the concatenation logic and tests. Pushed in commit 4afb1cfc4d52a49e8e09ded64b08aea066f7247a.

n8chur (approved):
Thanks for working through this!

robmaceachern (approved):
LGTM!

PR #468, opened 2023-09-20
Don't cache subelements

watt:
# Summary

In `LayoutStorage` we cache subelements, under the assumption that they will remain stable during a layout pass. This PR removes the caching, because it was a faulty assumption.

# Details

In `LayoutStorage` we currently assume the array of `LayoutSubelement`s for subelements will be stable during a layout pass, and cache it on the `LayoutTreeNode`. This was assumed safe because `LayoutStorage`-backed themselves elements cannot vary during a layout pass, and `LazyStorage`, which can, does not cache its subelement at all.

Turns out there is a case where this assumption is violated, if a `LazyStorage` produces a subtree that varies somewhere in a descendant node. For example, a `GeometryReader` that contains a stack, and varies the number of children in the stack based on space available.

# Performance

This PR is broken into 2 commits:
- First, a fix that is guarded behind a `LayoutMode` option. This option was used to performance test this change against the current behavior as a baseline.
- The second commit removes the option and adopts the fix unconditionally.

I was expecting to incur a hit from removing this caching but benchmarking showed it to be completely negligible, and attempts to more cleverly cache actually performed _worse_. I'll put up a Market PR that shows this benchmark and a regression test.

# Testing
There's a new `GeometryReader` test to validate this case, and regression test will be on the Market side.

See UI-4599.

### DIFF
--- a/BlueprintUI/Sources/Element/LayoutStorage.swift
+++ b/BlueprintUI/Sources/Element/LayoutStorage.swift
@@ -126,40 +126,19 @@ extension LayoutStorage: LegacyContentStorage {
 extension LayoutStorage: CaffeinatedContentStorage {
 
     private func subelements(from node: LayoutTreeNode, environment: Environment) -> LayoutSubelements {
-        if node.sizeCache.options.assumeStableSubelements {
-            // Current behavior
-            return node.layoutSubelements {
-                var identifierFactory = ElementIdentifier.Factory(elementCount: children.count)
-                return children.map { child in
-                    let identifier = identifierFactory.nextIdentifier(
-                        for: type(of: child.element),
-                        key: child.key
-                    )
-                    return LayoutSubelement(
-                        identifier: identifier,
-                        content: child.content,
-                        environment: environment,
-                        node: node.subnode(key: identifier),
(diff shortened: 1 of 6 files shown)


robmaceachern (approved):

kyleve (approved):
Thanks for getting this through!

BlueprintUI/Sources/Element/LayoutStorage.swift
```diff
@@ -126,21 +126,19 @@ extension LayoutStorage: LegacyContentStorage {
...
+                for: type(of: child.element),
+                key: child.key
+            )
+            return LayoutSubelement(
+                identifier: identifier,
+                content: child.content,
+                environment: environment,
+                node: node.subnode(key: identifier),
+                traits: child.traits
+            )
```
watt, line 141: Details of the alternatives I attempted:
- Caching each `LayoutSubelement` by identifier. This didn't work because `child.content` may vary and need to be re-evaluated.
- Making `LayoutSubelement` capture `child.content` as a closure and lazily evaluate it once the first time it is needed. Had no impact on performance.

PR #473, opened 2023-12-05
[wip] Add link attributes for detected links.

RoyalPineapple:
We need to add the `NSLinkAttributeName` to the attributed string for any detected links. This allows voiceover to announce that links are available when reading the label.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -263,7 +263,16 @@ extension AttributedLabel {
 
             if !isMeasuring {
                 if previousAttributedText != attributedText {
-                    links = attributedLinks(in: model.attributedText) + detectedDataLinks(in: model.attributedText)
+                    let attributedLinks = attributedLinks(in: model.attributedText)
+                    let detectedLinks = detectedDataLinks(in: model.attributedText)
+                    if let newText = attributedText {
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -263,7 +263,16 @@ extension AttributedLabel {
 
             if !isMeasuring {
                 if previousAttributedText != attributedText {
-                    links = attributedLinks(in: model.attributedText) + detectedDataLinks(in: model.attributedText)
+                    let attributedLinks = attributedLinks(in: model.attributedText)
+                    let detectedLinks = detectedDataLinks(in: model.attributedText)
+                    if let newText = attributedText {
```
kyleve, line 268: Is there a difference between `attributedText` and `model.attributedText`? IIRC one goes through `displayableAttributedText`.

RoyalPineapple: UGH `displayableAttributedText` is the whole issue here. its removing the link attribute which voiceover uses.

RoyalPineapple: Well, thats frustrating to discover. going back to draft while i recalibrate

RoyalPineapple:
We cant do this because we replace all the `.link` attributes with `.labelLink`attributes. back to the string manipulation ide...

PR #474, opened 2023-12-08
Allow specifying which axes are included in the final measurement when using ElementContent(measuring:)

kyleve:
When using `ElementContent(measuring:)` to create essentially a smart "`Spacer`" based on the size of an element, it's desirable to include in the final size on only one of the measurement dimensions. Thus, expose an `Axes` type you can optionally pass to the initializer to control this.

### DIFF
--- a/BlueprintUI/Sources/Element/ElementContent.swift
+++ b/BlueprintUI/Sources/Element/ElementContent.swift
@@ -279,8 +279,23 @@ extension ElementContent {
     ///
     /// This is useful if you are placing the element in a nested `BlueprintView`, for example (eg
     /// to create a stateful element) and just need this element to be correctly sized.
-    public init(measuring element: Element) {
-        storage = MeasureElementStorage(child: element)
+    ///
+    /// You can optionally pass which axes should be included in the final measurement. Defaults to both axes.
+    public init(
+        measuring element: Element,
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/Element/ElementContent.swift
```diff
@@ -279,8 +279,23 @@ extension ElementContent {
     ///
     /// This is useful if you are placing the element in a nested `BlueprintView`, for example (eg
     /// to create a stateful element) and just need this element to be correctly sized.
-    public init(measuring element: Element) {
-        storage = MeasureElementStorage(child: element)
```
kyleve, line 283: Posting here for posterity: We could expose a method like

```swift
extension Element {
   func asSpacer(along axes: MeasuringAxes = .both) { ... } 
}
```

So that you could use any ol' element as a dynamic spacer – which has been an infrequent but not _uncommon_ ask. I think for now, I'm not going to add this, but this would make it easier for some of these cases. I might put a call out in #market-ios to see if anyone would find this helpful.

kyleve:
@watt pointed out I can wrap this up in another element and then just ignore one of the measurements, which makes a lot more sense! I entirely missed that last week when I had been trying to figure out how to do that.

PR #475, opened 2023-12-13
Remove Legacy Layout

kyleve:
I took a swing at deleting the `Legacy` layout system, since it's no longer used, but is required to be implemented for new `Layout` types. To do this, I:

- Merged the existing legacy and caffeinated layout protocols into the base `Layout` protocol.
- Merged the existing legacy and caffeinated content storage protocols into the base `ContentStorage` protocol.
- Deleted all implementations of the legacy layout system.
- Removed the `legacy` `LayoutMode`.
- Removed tests which existed only for the legacy layout system.
- Removed the caches for the legacy system.
- Updated documentation to point at renamed methods and types (mainly at `Layout` directly).
- To that effect, I've tried to remove the term "caffeinated" from public facing docs, leaving it only as an implementation detail.

### Open Questions
I've left comments on things I'm not 100% sure what to do with, but the biggest open question is do we leave the `LayoutMode` enum around at all. I've opted to leave it in, because we can continue to leverage this to test and roll out any future improvements to Blueprint's layout engine, for example cross-layout caching, or further optimizations to the caffeinated layout contract. 

### Next Steps
After this lands, I'll take care of deleting the now dead code from Market, and POS.


BlueprintUI/Tests/RenderContextTests.swift
```diff
@@ -3,36 +3,39 @@ import XCTest
...
-        }
-
-        view.ensureLayoutPass()
-        XCTAssertEqual(contextualMode, defaultMode)
-
-        view.layoutMode = overrideMode
-        view.ensureLayoutPass()
-        XCTAssertEqual(contextualMode, overrideMode)
-    }
+    // TODO: Decide what to do here. Can't test overrides w/ only one layout mode...
```
kyleve, line 7: - [ ] Need to figure this out before merge

BlueprintUI/Sources/Layout/LayoutOptions.swift
kyleve, line 1: Some tests depend on turning this on/off to count measurements, so I decided to leave it in. Figure we can always remove it in a later iteration.

watt:
Let's talk about the strategy for this in January.

- [ ] Discuss the strategy for removing legacy layout in January

kyleve:
Do you mind enumerating concrete concerns to removing legacy layout? Eg:

- What would you want to keep vs remove? Does that diverge from what is done here?
- What are possible concerns and edge cases from removing legacy layout?

kyleve:
I'm closing this out as it sounds like we want to do the removal different than done here.

PR #476, opened 2023-12-16
Fix link detection for stretched labels

kyleve:
- Fixes link detection for stretched labels. The existing logic for determining the text alignment which is used to determine the alignment multiplier, was somewhat broken. This changes it to do some validation to ensure there's only one alignment in play as well, since otherwise, this won't work (we now assert in these cases).
- Add snapshot tests that visually show acceptable tap areas for label-containing links.
- Fix a vertical alignment bug that came from adding the above tests.


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -362,7 +362,7 @@ extension AttributedLabel {
                     let style = textStorage.attribute(.paragraphStyle, at: 0, effectiveRange: &range)
 
                     guard let style = style as? NSParagraphStyle, range == textStorage.entireRange else {
-                        return .natural
+                        return textAlignment
```
kyleve, line 365: - [ ] Tests

kylebshr:
I'm 99% sure this was just me goofing and not realizing that the labels text alignment was available/a better fallback. Definitely seems like the right thing to do, it's what the code getting the alignment out of the attributes is trying to do.

nsillik (approved):
easiest gorm ever

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Does this actually return an non-default value? Since we are using the attributed string API for UILabel, we don't set this anywhere. I'm not sure it's a good fallback if we don't know where it's coming from.

Do we know why we're seeing this fail in the first place? If it's happening on a single centered line, like on your PR description, we should look into that closer.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -354,19 +354,38 @@ extension AttributedLabel {
...
-                // Otherwise, fall back to `.natural`.
-                let alignment: NSTextAlignment = {
-                    var range: NSRange = NSRange(location: -1, length: 0)
-                    let style = textStorage.attribute(.paragraphStyle, at: 0, effectiveRange: &range)
-
-                    guard let style = style as? NSParagraphStyle, range == textStorage.entireRange else {
-                        return .natural
+            /// The below positioning calculation assumes that there is only one
+            /// alignment within the label, so verify that is the case.
+            func alignment() -> NSTextAlignment {
```
kyleve, line 359: @watt This is what I ended up coming up with to (a) validate we only have one alignment, and (b) use that alignment (vs just `textAlignment`) for link calculation, since we don't actually know `textAlignment`'s heuristic (though its probably basically just what we do here, in practice, I assume).

kyleve: @watt Ping for a review here when you get a chance!

kyleve: @watt Ping on this if you have time this week

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -354,19 +354,38 @@ extension AttributedLabel {
...
+
+                guard let string = attributedText, string.length > 0 else {
+                    return textAlignment
+                }
+
+                var alignments = [NSTextAlignment]()
```
watt, line 365: Could we use a `Set` here? We really care about _unique_ alignments, right? For example, if there are 3 `paragraphStyle` runs that are all left-aligned, that would be fine.

BlueprintUICommonControls/Sources/AttributedLabel.swift
watt: Sounds fine to me. Can you add a bit to this function's doc comment about the `textAlignment` fallback?

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Ahh yeah, good call. Updating!

- [ ] Set

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: 👍

kyleve: - [ ] Update docs

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -388,7 +411,7 @@ extension AttributedLabel {
             let textBoundingBox = layoutManager.usedRect(for: textContainer)
             let textContainerOffset = CGPoint(
                 x: (labelSize.width - textBoundingBox.size.width) * alignmentMultiplier - textBoundingBox.origin.x,
-                y: (labelSize.height - textBoundingBox.size.height) * alignmentMultiplier - textBoundingBox.origin.y
+                y: (labelSize.height - textBoundingBox.size.height) * 0.5 - textBoundingBox.origin.y
```
kyleve, line 414: I added the tap detection snapshot tests, and it found that this has always been broken for any visually stretched label that wasn't _also_ centered. Oops!

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -259,6 +259,95 @@ class AttributedLabelTests: XCTestCase {
         compareSnapshot(of: element)
     }
 
+    func test_linkTapDetection() {
```
kyleve, line 262: @watt Went ahead and added these tap detection snapshot tests

- [ ] Pull down actual correct snapshots from CI

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Done

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
kyleve: - [ ] Pull down actual correct snapshots from CI

watt (approved):

PR #478, opened 2024-02-07
Allow for customization of the preview name

nsillik:
Before, there was no way to explicitly set preview name, leading to things like this in the Xcode canvas view:

image

With no way to make it look prettier.

With this change we either use the Xcode SwiftUI Previews default, or just the name:

image

now shows:

image

much better!


BlueprintUI/Sources/BlueprintView/ElementPreview.swift
```diff
@@ -67,6 +67,7 @@ public struct ElementPreview: View {
     public typealias ElementProvider = () -> Element
 
     private let name: String
+    private let previewName: String?
```
n8chur, line 70: I believe this is only used for previews—I think we might want to just remove the prefix when using `name`.

I think we just want an optional `previewName` override on the Market side.

Thoughts?

nsillik: Okay, here's the result of that suggestion:
image

@kyleve do you have opinions? is keeping the `prefix - name` worthwhile in any cases?

BlueprintUI/Sources/BlueprintView/ElementPreview.swift
```diff
@@ -191,34 +191,28 @@ extension ElementPreview {
...
         ) -> AnyView {
 
-            let formattedName: String = {
-                if name.isEmpty == false {
-                    return " – " + name
-                } else {
-                    return ""
-                }
-            }()
+            let formattedName: String? = name.isEmpty ? nil : name
```
n8chur, line 194: I'm not actually sure, and fine keeping this in, but should we just allow empty string here? We could document that it maps to `.previewDisplayName()` so consumers know it matches that behavior.
What happens if you do provide empty string?

n8chur (approved):

kyleve (approved):

BlueprintUI/Sources/BlueprintView/ElementPreview.swift
nsillik: if you provide an empty string here it defaults to the SwiftUI default. I _didn't_ want to change it to trinary (nil, empty, named), because that'd break the interface.

If you pass in an empty string it should be the same as it was before

nsillik: if you pass in an empty string you get the SwiftUI default, if you pass in a name it shows the name. I could have made this trinary (`String?`) and have `nil` give you the swiftUI default, empty string keep the same behavior ("size that fits" etc.), and a non-empty string override the preview name.

nsillik:
- [ ] changelog this

BlueprintUI/Sources/BlueprintView/ElementPreview.swift
n8chur: Eh, I kinda like the idea of having the `name` match the type of `previewDisplayName` but probably not worth the churn. Fine keeping as is!

PR #481, opened 2024-02-21
Bumping versions to 2.3.0

nsillik:
### Fixed

- Fixed an issue where `AttributedLabel` would not properly handle tapping on links when a label was stretched.

### Added

- `AccessibilityElement` now supports providing arbitrary strings to assistive devices using the `AXCustomContent` protocol. 

### Changed

- The behavior of `name` of `ElementPreview` has been change, affecting the SwiftUI `previewName`. Instead of including device or size information (i.e. `sizeThatFits - \(name)`), it now either defaults to the Xcode default if given an empty string, and shows _only_ the `name` if `name` is non-empty.
- Updated minimum deployment target from iOS 14 to iOS 15.

### Internal
- Updated CI to use M1 machines, Xcode 15.1, and Ruby 3.2.2.
- Added iOS 17 snapshot images.
- Bump Swift version to 5.9.
- Update Ruby gems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+### Added
+
+### Removed
+
+### Changed
+
+### Deprecated
+
+### Security
+
+### Documentation
+
+### Misc
+
+### Internal
+
+# Past Releases
+
+## [2.3.0] - 2024-02-21
+
+### Fixed
+
 - Fixed an issue where `AttributedLabel` would not properly handle tapping on links when a label was stretched.
 
 ### Added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -32,13 +25,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Internal
 
+# Past Releases
+
+## [2.3.0] - 2024-02-21
```
robmaceachern, line 30: I think technically the deployment target bump would necessitate a major version bump but I'm not too concerned about it.

robmaceachern (approved):

CHANGELOG.md
kyleve: +1, I'd do a 3.0 here, since if someone is bound to only 2.0 in their profile and still using iOS 14, they'll get this update and then it'll error... I think

kyleve (approved):

CHANGELOG.md
nsillik: that's a good point, here's the new PR: #482

watt: I don't think I would consider that a major breaking change, but I guess I don't feel too strongly about it. I don't want to get into a pattern of frequent major releases of Blueprint.

PR #482, opened 2024-02-21
Bumping versions to 3.0.0

nsillik:
### Fixed

- Fixed an issue where `AttributedLabel` would not properly handle tapping on links when a label was stretched.

### Added

- `AccessibilityElement` now supports providing arbitrary strings to assistive devices using the `AXCustomContent` protocol. 

### Changed

- The behavior of `name` of `ElementPreview` has been change, affecting the SwiftUI `previewName`. Instead of including device or size information (i.e. `sizeThatFits - \(name)`), it now either defaults to the Xcode default if given an empty string, and shows _only_ the `name` if `name` is non-empty.
- Updated minimum deployment target from iOS 14 to iOS 15.

### Internal
- Updated CI to use M1 machines, Xcode 15.1, and Ruby 3.2.2.
- Added iOS 17 snapshot images.
- Bump Swift version to 5.9.
- Update Ruby gems.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -120,6 +120,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-22
+  arm64-darwin-23
 
 DEPENDENCIES
   cocoapods (~> 1.11)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+### Added
+
+### Removed
+
+### Changed
+
+### Deprecated
+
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -120,6 +120,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-22
+  arm64-darwin-23
```
n8chur, line 123: Was this change intentional? It seems fine, but wanted to double check.

n8chur (approved):

Gemfile.lock
nsillik: This is just Sonoma, figured I'd keep it in otherwise it's just gonna keep happening

PR #483, opened 2024-03-05
AccessibilityBlocker aggressively blocking.

RoyalPineapple:
Fixed a bug where `AccessibilityBlocker` would block accessibility when `isBlocking` is `false`

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
@@ -33,7 +33,7 @@ public struct AccessibilityBlocker: Element {
 
     public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
         UIView.describe { config in
-            config[\.isAccessibilityElement] = !isBlocking
+            config[\.isAccessibilityElement] = false
             config[\.accessibilityElementsHidden] = isBlocking
         }
     }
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
```diff
@@ -33,7 +33,7 @@ public struct AccessibilityBlocker: Element {
 
     public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
         UIView.describe { config in
-            config[\.isAccessibilityElement] = !isBlocking
+            config[\.isAccessibilityElement] = false
```
kyleve, line 36: Ahh, since this is just a container, right?

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
```diff
@@ -33,7 +33,7 @@ public struct AccessibilityBlocker: Element {
 
     public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
         UIView.describe { config in
-            config[\.isAccessibilityElement] = !isBlocking
```
RoyalPineapple, line 36: This UIView should never be an accessibilityElement because it's always wrapping another element. 
if this gets set to true (when isBlocking is false) it also blocks the accessibility because voiceover never asks for its children.

BlueprintUICommonControls/Sources/AccessibilityBlocker.swift
RoyalPineapple: exactly

PR #484, opened 2024-03-12
feat(a11y): Add support for accessibility focus triggers

meherkasam-square:
This change provides a mechanism to force accessibility focus on a Blueprint element at any given time through the use of triggers.

Thanks to @RoyalPineapple for pairing with me on this.

### Demo

https://github.com/square/Blueprint/assets/72885155/c3e83e2d-02e1-4145-ac7f-50f9ec35768c

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityFocus.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityFocus.swift
@@ -1,0 +1,109 @@
+import BlueprintUI
+import UIKit
+
+/// Enables VoiceOver focus to jump to the wrapped element via a trigger that can be manually fired.
+public struct AccessibilityFocus: Element {
+
+    /// The element that will have the focus.
+    public var wrapped: Element
+
+    /// A object that can be held on to by the caller to manually trigger a focus.
+    public var trigger: Trigger
+
+    /// Creates a new `AccessibilityFocus` wrapping the provided element.
+    public init(
+        wrapping wrapped: Element,
+        trigger: Trigger
+    ) {
+        self.wrapped = wrapped
+        self.trigger = trigger
+    }
+
+    // MARK: Element
+
+    public var content: ElementContent {
+        ElementContent(child: wrapped)
+    }
+
+    public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
+        View.describe { config in
+            config.apply { view in
+                view.apply(model: self)
+            }
+        }
+    }
+}
+
+// MARK: Backing view
+
+extension AccessibilityFocus {
+
+
+    private final class View: UIView {
+
+        private var trigger: Trigger?
+
+        func apply(model: AccessibilityFocus) {
+            trigger?.action = nil
+            trigger = model.trigger
+            model.trigger.action = { [weak self] in
+                guard let self = self else { return }
+                UIAccessibility.post(notification: model.trigger.notificationType, argument: self)
+            }
+        }
+
+        override var isAccessibilityElement: Bool {
+            get { false }
+            set {}
+        }
+    }
+}
+
+// MARK: Trigger
+
+extension AccessibilityFocus {
+
+    /// A trigger that can be used to manually fire an accessibility focus.
+    public final class Trigger {
+
+        /// The type of accessibility notification that will be triggered.
+        public var notificationType: UIAccessibility.Notification
+
+        /// An optional identifier for the trigger.
+        public var identifier: AnyHashable?
+
+        /// Creates a new trigger for the purpose of changing accessibility focus.
+        /// - Parameters:
+        ///   - notificationType: Type of accessibility notification to trigger. Defaults to `.layoutChanged`. Limited to `.layoutChanged` or `.screenChanged`.
+        ///   - identifier: An optional identifier for the trigger. Defaults to `nil`.
+        public init(
+            notificationType: UIAccessibility.Notification = .layoutChanged,
+            identifier: AnyHashable? = nil
+        ) {
+            self.notificationType = notificationType
+            self.identifier = identifier
+        }
+
+        fileprivate var action: (() -> Void)?
+
+        /// Manually fire the trigger
+        public func focus() {
+            action?()
+        }
+    }
+}
+
+extension Element {
+
+    /// Enables VoiceOver focus to jump to the wrapped element via the trigger.
+    /// - Parameters:
+    ///   - on: A reference-type trigger object that can be used to trigger accessibility focus via the `focus()` function.
+    public func accessibilityFocus(
+        on trigger: AccessibilityFocus.Trigger
+    ) -> Element {
+        AccessibilityFocus(
+            wrapping: self,
+            trigger: trigger
+        )
+    }
+}
--- a/SampleApp/Sources/AccessibilityViewController.swift
+++ b/SampleApp/Sources/AccessibilityViewController.swift
@@ -1,0 +1,45 @@
+import BlueprintUI
+import BlueprintUICommonControls
+import UIKit
+
+final class AccessibilityViewController: UIViewController {
+
+    private let blueprintView = BlueprintView()
+
+    override func loadView() {
+        view = blueprintView
+        blueprintView.element = element
+    }
+
+    var firstTrigger = AccessibilityFocus.Trigger()
+    var secondTrigger = AccessibilityFocus.Trigger()
+
+    var element: Element {
+
+        Column(alignment: .fill, minimumSpacing: 20) {
+            Row {
+                Label(text: "First")
+                    .accessibilityFocus(on: firstTrigger)
+                Spacer()
+                Label(text: "Second")
+                    .accessibilityFocus(on: secondTrigger)
+            }
+            Button(
+                onTap: {
+                    self.firstTrigger.focus()
+                },
+                wrapping: Label(text: "Focus on First", configure: { label in
+                    label.color = .systemBlue
+                })
+            )
+
+        }
+        .inset(uniform: 20)
+        .centered()
+        .onAppear {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.15) {
+                self.secondTrigger.focus()
+            }
+        }
+    }
+}
--- a/SampleApp/Sources/AccessibilityViewController.swift
+++ b/SampleApp/Sources/AccessibilityViewController.swift
@@ -18,22 +18,61 @@ final class AccessibilityViewController: UIViewController {
 
         Column(alignment: .fill, minimumSpacing: 20) {
             Row {
+                Label(text: "Focus") { label in
+                    label.font = .systemFont(ofSize: 24, weight: .bold)
+                    label.accessibilityTraits = [.header]
+                }
+            }
+
+            Row {
                 Label(text: "First")
                     .accessibilityFocus(on: firstTrigger)
                 Spacer()
                 Label(text: "Second")
                     .accessibilityFocus(on: secondTrigger)
             }
-            Button(
-                onTap: {
-                    self.firstTrigger.focus()
-                },
-                wrapping: Label(text: "Focus on First", configure: { label in
-                    label.color = .systemBlue
-                })
-            )
+
+            Row {
+                Button(
+                    onTap: {
+                        self.firstTrigger.focus()
+                    },
+                    wrapping: Label(text: "Focus on First", configure: { label in
+                        label.color = .systemBlue
+                    })
+                )
+            }
+
+            Row {
+                Label(text: "Blocker") { label in
+                    label.font = .systemFont(ofSize: 24, weight: .bold)
+                    label.accessibilityTraits = [.header]
+                }
+            }
+
+            Row {
+                Label(text: "Blocked label").blockAccessibility()
+            }
+
+            Row {
+                Label(text: "Accessibility Element") { label in
+                    label.font = .systemFont(ofSize: 24, weight: .bold)
+                    label.accessibilityTraits = [.header]
+                }
+            }
+
+            Row {
+                Label(text: "Title")
+
+                Spacer()
+
+                Label(text: "Detail") { label in
+                    label.color = .systemGray
+                }
+            }.accessibilityElement(label: "Title", value: "Detail", traits: [])
 
         }
+        .accessibilityContainer()
         .inset(uniform: 20)
         .centered()
         .onAppear {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where `AccessibilityBlocker` would block accessibility when `isBlocking` is `false`
 
 ### Added
+- Added support for accessibility focus triggers to force VoiceOver to focus on any given element.
 
 ### Removed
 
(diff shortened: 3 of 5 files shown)


BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,115 @@
...
+        fileprivate func fireImmediately() {
+            if firesImmediately && !hasFired {
+                fire()
+            }
+        }
+
+        /// Manually fire the trigger
+        public func fire() {
+            action?()
+            hasFired = true
```
RoyalPineapple, line 81: Maybe we should check if action is nil and only set this if it indeed fired. That would prevent a situation where this method gets called before the view has been hooked up and it doesn't actually fire.

SampleApp/Sources/AccessibilityViewController.swift
```diff
@@ -0,0 +1,39 @@
...
+                Spacer()
+                Label(text: "Second")
+                    .accessibilityFocusOnAppear()
+            }
+            Button(
+                onTap: {
+                    self.trigger.fire()
+                },
+                wrapping: Label(text: "Focus on First", configure: { label in
+                    label.color = .blue
```
RoyalPineapple, line 31: Nit: `.systemBlue`

SampleApp/Sources/AccessibilityViewController.swift
```diff
@@ -0,0 +1,39 @@
+import BlueprintUI
+import BlueprintUICommonControls
+import UIKit
+
+final class AccessibilityViewController: UIViewController {
```
RoyalPineapple, line 5: Not a blocker but we might want to add some more accessibility Examples to this view.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+}
+
+extension Element {
+
+    /// Enables VoiceOver focus to jump to the wrapped element via the trigger.
+    /// - Parameters:
+    ///   - notificationType: Type of accessibility notification to trigger. Defaults to `.layoutChanged`.
```
RoyalPineapple, line 90: We should point out that `.notification` doesn't work here and you're really limited to `.layoutChanged` and `.screenChanged`

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
+import BlueprintUI
+import UIKit
+
+/// Enables VoiceOver focus to jump to the wrapped element
+/// either on appearance or via a trigger that can be manually fired.
+public struct AccessibilityFocus: Element {
```
kyleve, line 6: Missing a public init

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    // MARK: Element
+
+    public var content: ElementContent {
+        ElementContent(child: wrapping)
+    }
+
+    public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
+        View.describe { config in
+            config[\.isAccessibilityElement] = false
```
kyleve, line 25: Should this just be in the `View` itself if it's always false?

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+
+        func apply(model: AccessibilityFocus) {
+            model.trigger.action = { [weak self] in
+                guard let self = self else { return }
+                UIAccessibility.post(notification: model.notificationType, argument: self)
+            }
+
+            /// We need to wait a moment before firing the focus, to ensure that
+            /// we are not competing with UIAccessibility for the focus.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
```
kyleve, line 46: Fwiw this is generally 0.15 in Market. It kinda makes me think we need an a11y queue or something... I'm not sure there is actually one right answer here wrt. "override the in progress actions" vs "wait for them to finish"...

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+}
+
+// MARK: Trigger
+
+extension AccessibilityFocus {
+
+    /// A trigger that can be used to manually fire an accessibility focus.
+    public class Trigger {
```
kyleve, line 58: final

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+                fire()
+            }
+        }
+
+        /// Manually fire the trigger
+        public func fire() {
+            guard let action else { return }
+
+            action()
+            hasFired = true
```
kyleve, line 83: I'd also recommend setting this before `action()` in case that call causes a re-entrant update.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
kyleve: (Unless you only want to allow this to be used via the `Element` builder – in which case, ignore me!)

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+        /// - Parameters:
+        ///   - firesImmediately: If `true`, the trigger will fire immediately when the element appears.
+        public init(firesImmediately: Bool = false) {
+            self.firesImmediately = firesImmediately
+        }
+
+        fileprivate var action: (() -> Void)?
+        private var hasFired = false
+
+        fileprivate func fireImmediately() {
```
kyleve, line 72: I'd call this `fireImmediatelyIfNeeded()` or similar

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+        func apply(model: AccessibilityFocus) {
+            model.trigger.action = { [weak self] in
+                guard let self = self else { return }
+                UIAccessibility.post(notification: model.notificationType, argument: self)
+            }
+
+            /// We need to wait a moment before firing the focus, to ensure that
+            /// we are not competing with UIAccessibility for the focus.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
+                model.trigger.fireImmediately()
```
kyleve, line 47: Fwiw there's no guarantee that this is in a view hierarchy or window hierarchy when this is invoked. Should this be done in `viewDidMoveToWindow` or something? What is the use case here? It feels fragile, especially in list views - where the cells, which are recycled, remain in the list forever – but are not always visible.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+            wrapping: self,
+            notificationType: notificationType,
+            trigger: trigger
+        )
+    }
+
+    /// Enables VoiceOver focus to jump to the wrapped element upon its appearance.
+    /// - Parameters:
+    ///   - notificationType: Type of accessibility notification to trigger. Defaults to `.layoutChanged`. Limited to `.layoutChanged` or `.screenChanged`.
+    public func accessibilityFocusOnAppear(
```
kyleve, line 108: If you want to use "on appear" as the verbiage here, it might be worth hooking into the existing `onAppear` machinery we have in Blueprint – which waits for the entire BlueprintView to be laid out.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
meherkasam-square: Oh good call!

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
meherkasam-square: The goal here was to trigger the accessibility focus on any element that just appeared on the screen or if there's any one element we would like to take automatic accessibility focus when a new screen loads. In the scenario you describe, do we set the model as a property on the `View` class through the `backingViewDescription` and execute this code inside `didMoveToWindow()`?

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
RoyalPineapple: I don't know that a queue even work for this as voiceover will always override our notifications in favor of system announcements and we don't have any insight into wether any announcements happen to be in progress when we post our notification. 

There are a limited number of ways we can gain any insight or control into the announcements, but none work are helpful this context. 

we can observe the `UIAccessibility.announcementDidFinishNotification`, but this only gets sent for `.announcement` notifications, which expect a string rather than an element as an argument and don't alter accessibility focus.

Similarly, we can use the attributed string attribute 
`UIAccessibilitySpeechAttributeQueueAnnouncement` to queue an announcement behind any in progress speech, but again this is only useful for `.announcement` notifications and requires a string rather than an element. 

iOS 17 expands this by adding the concept of  UIAccessibilityPriority which allows us to post an an announcement that is uninterruptible, but again this is only available for attributed strings and `.announcement` notifications

kyleve: Fair enough – I was mostly thinking out loud. It's too bad there's no better way to deal with this. Ah well!

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
kyleve: Something like that... or as mentioned in another comment, hook this into the `onAppear` machinery in Blueprint? Not 100% sure.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+// MARK: Backing view
+
+extension AccessibilityFocus {
+    private final class View: UIView {
+
+        func apply(model: AccessibilityFocus) {
+            model.trigger.action = { [weak self] in
```
watt, line 39: The pattern I'd recommend is to hold a reference to the trigger on `View`, so that you can clear out old triggers when setting this, e.g.:
```swift
var accessibilityFocusTrigger: Trigger {
  didSet {
    oldValue?.action = nil
    accessibilityFocusTrigger.action = { [weak self] in
      // action to perform on trigger
    }
  }
}
```

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
watt: Yeah, Kyle is correct, you can't rely on this, and we don't support it in "normal" focus for that reason. To focus on appear you'd probably want a construction like this:

```swift
myElement
  .accessibilityFocus(trigger: trigger)
  .onAppear {
    trigger.fire()
  }
```

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
watt: also fwiw a common pattern is often to call the parameter `wrapping` but the property `wrapped`, like so:
```swift
struct Foo: Element {
  var wrapped: Element
  init(wrapping wrapped: Element) {
    self.wrapped = wrapped
  }
}

File: BlueprintUICommonControls/Sources/AccessibilityFocus.swift [hunk 14]
```diff
@@ -0,0 +1,117 @@
...
+        private var hasFired = false
+
+        fileprivate func fireImmediately() {
+            if firesImmediately && !hasFired {
+                fire()
+            }
+        }
+
+        /// Manually fire the trigger
+        public func fire() {
```
watt, line 79: I would name this method for the thing it does: `focus()`.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,117 @@
...
+
+extension Element {
+
+    /// Enables VoiceOver focus to jump to the wrapped element via the trigger.
+    /// - Parameters:
+    ///   - notificationType: Type of accessibility notification to trigger. Defaults to `.layoutChanged`. Limited to `.layoutChanged` or `.screenChanged`.
+    ///   - trigger: A reference-type trigger object that can be used to trigger accessibility focus immediately on appearance or at a later point via the `fire()` function.
+    public func accessibilityFocus(
+        notificationType: UIAccessibility.Notification = .layoutChanged,
+        trigger: AccessibilityFocus.Trigger
```
watt, line 96: just a little bikeshedding:
```suggestion
        on trigger: AccessibilityFocus.Trigger
```

watt (commented):
I think the "immediately" behavior is not going to work, otherwise mostly just small suggestions.

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
RoyalPineapple, line 58: Just a thought, perhaps we should add an optional identifier to the trigger to help with managing a bunch of them

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        fileprivate var action: (() -> Void)?
+        private var hasFired = false
+
+        /// Manually fire the trigger
+        public func focus() {
+            guard let action else { return }
+
+            hasFired = true
```
RoyalPineapple, line 87: Do we still need to track this?

RoyalPineapple (approved):

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,112 @@
...
+        ) {
+            self.notificationType = notificationType
+            self.identifier = identifier
+        }
+
+        fileprivate var action: (() -> Void)?
+
+        /// Manually fire the trigger
+        public func focus() {
+            guard let action else { return }
```
RoyalPineapple, line 92: we can replace this guard with just `action?()`

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,110 @@
...
+            action?()
+        }
+    }
+}
+
+extension Element {
+
+    /// Enables VoiceOver focus to jump to the wrapped element via the trigger.
+    /// - Parameters:
+    ///   - trigger: A reference-type trigger object that can be used to trigger accessibility focus immediately on appearance or at a later point via the `focus()` function.
```
RoyalPineapple, line 101: this commment is out of date

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
```diff
@@ -0,0 +1,109 @@
...
+
+extension AccessibilityFocus {
+
+
+    private final class View: UIView {
+
+        private var trigger: Trigger?
+
+        func apply(model: AccessibilityFocus) {
+            trigger?.action = nil
```
kyleve, line 47: Thinking out loud, should we do something like 

```swift
trigger?.action = {
  print("Trigger called after being reset. Set a breakpoint here to debug.")
}
```
Or something?

kyleve (approved):

BlueprintUICommonControls/Sources/AccessibilityFocus.swift
watt: I wouldn't bother, none of the other trigger implementations do this. It's more about cleaning up after ourselves than protecting against some programmer error.

watt (approved):
remember to add this to the changelog

PR #486, opened 2024-03-21
Added `startTimestamp` to `BlueprintViewRenderMetrics`

kcsiegal:
Added `startTimestamp` to `BlueprintViewRenderMetrics`. This will allow us to represent Blueprint renders as spans in our performance logging.

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -429,6 +429,7 @@ public final class BlueprintView: UIView {
             self,
             completedRenderWith: .init(
                 layoutMode: layoutMode,
+                startTimestamp: startTime,
                 totalDuration: viewUpdateEndTime - startTime,
                 layoutDuration: layoutEndTime - startTime,
                 viewUpdateDuration: viewUpdateEndTime - layoutEndTime
@@ -537,6 +538,9 @@ public struct BlueprintViewRenderMetrics {
 
     /// The layout mode used to render the view.
     public var layoutMode: LayoutMode
+
+    /// The mach time in seconds at which the view render started (from `CACurrentMediaTime()`).
+    public var startTimestamp: TimeInterval
 
     /// The total time it took to apply a new element.
     public var totalDuration: TimeInterval
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added support for accessibility focus triggers to force VoiceOver to focus on any given element.
+- Added `startTimestamp` to `BlueprintViewRenderMetrics`. This represents the mach time in seconds at which the render started, from `CACurrentMediaTime()`.
 
 ### Removed
 


kyleve (approved):

watt:
> This will allow us to represent Blueprint renders as spans in our performance logging.

Can I get some context on how this is useful? `CACurrentMediaTime` is not a timestamp, it's basically only useful as a reference to another `CACurrentMediaTime` value.

kcsiegal:
@watt for context, I'm planning to use these values to log `BlueprintView` renders as sub-spans of user interaction time spans. I'm most interested in these values as interval start and end markers relative to other interaction events and sub-spans that occurred on the device, such as touch-up events and Workflow renders, rather than for representing a wall-clock time. We generally use `CACurrentMediaTime()` for "timestamps" elsewhere in our Tracing library for this reason - absolute time is less important than accurately measuring elapsed intervals, as well as accurately representing when different intervals occurred relative to one another. 

So my intention was to keep these as relative, monotonic values until the point where they are logged and sent off device, at which point they're converted to `Date` along with all other intervals else we're logging within the interaction span. 

I'm open to the idea that I'm being overly cautious and these can be safely exposed as `Date` values here even though `Date` is not monotonic, but I also don't necessarily see a downside to using `CACurrentMediaTime`. Would be interested to hear @amorde's thoughts on this decision as well - would `Date` be fine here?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -546,6 +548,12 @@ public struct BlueprintViewRenderMetrics {
 
     /// The time it took to update the on-screen views for the element.
     public var viewUpdateDuration: TimeInterval
+
+    /// The mach time in seconds at which the view render started (from `CACurrentMediaTime()`).
+    public var startTimestamp: TimeInterval
+
+    /// The mach time in seconds at which the view finished updating (from `CACurrentMediaTime()`).
+    public var endTimestamp: TimeInterval
```
robmaceachern, line 556: Do we need the `endTimestamp` given we have the start and total duration?

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
robmaceachern, line 1: (moving discussion here for threading purposes)

@watt:
> Can I get some context on how this is useful? CACurrentMediaTime is not a timestamp, it's basically only useful as a reference to another CACurrentMediaTime value.

@kcsiegal:
> for context, I'm planning to use these values to log BlueprintView renders as sub-spans of user interaction time spans. I'm most interested in these values as interval start and end markers relative to other interaction events and sub-spans that occurred on the device, such as touch-up events and Workflow renders, rather than for representing a wall-clock time. We generally use CACurrentMediaTime() for "timestamps" elsewhere in our Tracing library for this reason - absolute time is less important than accurately measuring elapsed intervals, as well as accurately representing when different intervals occurred relative to one another.

> So my intention was to keep these as relative, monotonic values until the point where they are logged and sent off device, at which point they're converted to Date along with all other intervals else we're logging within the interaction span.

> I'm open to the idea that I'm being overly cautious and these can be safely exposed as Date values here even though Date is not monotonic, but I also don't necessarily see a downside to using CACurrentMediaTime. Would be interested to hear @amorde's thoughts on this decision as well - would Date be fine here?

robmaceachern: The type of the timestamp was my question too. I had done a bit of research around this last year when I was making some updates to `BlueprintViewRenderMetrics` but didn't land on a great solution so I punted it.

@kcsiegal What's the conversion process for going from a `CACurrentMediaTime` to a `Date`? I thought there were cases where `CACurrentMediaTime` basically pauses time in different device states, which I imagine would make it hard to map back to a `Date` reliably (or vice-versa).

amorde: The cases where `CACurrentMediaTime` pauses are likely not relevant to us here as they are when the device goes to sleep. If we start a blueprint render and dont finish it before the device goes to sleep then something else is likely wrong.

Our tracing code is somewhat forced to work with `CACurrentMediaTime`-based timestamps because `CADisplayLink` reports frame timings using that. 

We have a method for converting between the two [here](https://github.com/squareup/ios-register/blob/8f0bac4b223899aec01c94e83b0fd4985f5a3d4f/Frameworks/Tracing/Public/Sources/Clock.swift#L13-L16)

amorde: `Date` has the downside of being affected by the devices clock. So if the system updates its internal clock after querying a remote time server with `ntp` or w/e they use these days, time can jump arbitrarily forward or backward.

robmaceachern (approved):

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kcsiegal: I thought about that as well - I'm happy to remove it.

robmaceachern: I'd probably omit it, or just make it a computed property.

kcsiegal: I removed it!

PR #487, opened 2024-03-26
Prepare 3.1.0 release

kyleve:
### Fixed
- Fixed a bug where `AccessibilityBlocker` would block accessibility when `isBlocking` is `false`.

### Added
- Added support for accessibility focus triggers to force VoiceOver to focus on any given element.
- Added `startTimestamp` to `BlueprintViewRenderMetrics`. This represents the mach time in seconds at which the render started, from `CACurrentMediaTime()`.


nsillik (approved):
🥇

PR #488, opened 2024-03-26
updating custom content

RoyalPineapple:
In integrating this into  market i realized that some internal methods should be public. 

Furthermore I discovered how to use the name default without bothering the compiler, so we should rename this for sake of constancy.


kyleve (approved):
Don't forget to update the Changelog!

PR #489, opened 2024-04-16
Add Flow layout, for laying out elements in a flow-style layout

kyleve:
https://block.atlassian.net/browse/UI-6001

Introduces a `Flow` layout type, which allows for laying out elements in a flow style – as a row is filled with items, the layout flows to a new row, etc. This is pretty much a direct port of the `Verticals/Restaurant/Frameworks/RSTUIComponents/Public/Sources/Flow.swift` layout from SPOS, which I'd helped the RST team write a while back. We now have 4+ copies of this and similar layouts floating around the POS repo, and requests for 2 more use cases on the way, so let's make this a standardized element instead.

Nick has graciously offered to do the POS integration, which is here https://github.com/squareup/ios-register/pull/103940, and allows us to delete these duplicate implementations.


BlueprintUI/Tests/FlowTests.swift
kyleve, line 1: - [ ] Test for sizing

kyleve: - [ ] Note comment that snapshots are in `CommonControls`

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,361 @@
...
+        }
+    }
+}
+
+// MARK: - RowLayout
+
+extension Flow.Layout {
+
+    /// Helper for computing the frames for a row of items.
+    struct RowLayout {
```
kyleve, line 273: fileprivate?

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,361 @@
...
+        struct Item {
+            let size: CGSize
+            let xOffset: CGFloat
+        }
+
+        /// `True` if we can fit an item of the given size in the row.
+        func canFitItem(
+            of size: CGSize
+        ) -> Bool {
+            itemCount * itemSpacing + totalItemWidth + size.width <= maxWidth
```
kyleve, line 312: Add some () to make this math a bit clearer

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,361 @@
+import UIKit
+
+
+/// Element which lays out children horizontally, wrapping to another row when there is not enough space.
```
kyleve, line 4: Include a code sample in the docs

nsillik:
lookin' good on the register side: https://github.com/squareup/ios-register/pull/103940

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,372 @@
...
+
+        /// Shoter items are aligned to the bottom of the row.
+        case bottom
+    }
+
+    /// A child placed within the flow layout.
+    public struct Child: ElementBuilderChild {
+
+        /// The key used to optionally unique the child item.
+        public var key: AnyHashable?
```
kyleve, line 122: Need to add some element extensions for setting this key, like we do for `stackLayoutChild`

nsillik (approved):
gorm

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,372 @@
...
+/// │ │   4   ││  5  ││     6     │       │
+/// │ └───────┘└─────┘└───────────┘       │
+/// └─────────────────────────────────────┘
+/// ```
+public struct Flow: Element {
+
+    /// How to horizontally align items in the row when there is extra space.
+    public var horizontalAlignment: HorizontalAlignment
+
+    /// How to vertically align items which there is extra space.
```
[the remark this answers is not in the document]
kyleve: oops, fixed!

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,372 @@
...
+
+
+/// Element which lays out children horizontally, wrapping to another row when there is not enough space.
+///
+/// You may control the layout of each row within the flow layout, by providing a `HorizontalAlignment`,
+/// which controls the horizontal alignment of content within a row when it is smaller than its container,
+/// and via `RowAlignment`, which controls the vertical alignment of shorter elements within a row.
+///
+/// ```swift
+/// Flow(horizontalSpacing: 10, rowSpacing: 10) {
```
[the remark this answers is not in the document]
kyleve: I thought about this, but it felt sort of ambiguous – so I left it off – it's not hard to pass the value twice IMO. Curious for other's thoughts though!

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,389 @@
...
+                    xOffset: totalItemWidth + itemSpacing * itemCount
+                )
+            )
+            totalItemWidth += size.width
+            height = max(size.height, height)
+        }
+
+        /// Compute frames for the items in the row layout.
+        func itemFrames() -> [CGRect] {
+            let totalSpacing = (itemCount - 1) * itemSpacing
```
kyleve, line 357: Confirm we can't end up with a `0-1` right?

kyleve: Seems like no

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+                // Note: We always want at least one item per row.
+                if !row.canFitItem(of: elementSize) && row.items.isEmpty == false {
+                    frames += row.itemFrames()
+
+                    totalHeight += row.height + rowSpacing
+
+                    row = rowLayout(
+                        origin: .init(x: 0, y: totalHeight),
+                        maxWidth: constraint.maximum.width
+                    )
```
robmaceachern, line 228: Is `origin.x` even read anywhere?

robmaceachern (approved):

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+                }
+
+                row.addItem(of: elementSize)
```
meherkasam-square, line 231: What would happen if the item is larger than the row, so even one wouldn't fit? Does it rescale the item or truncate it?

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+            self.itemSpacing = itemSpacing
+            self.horizontalAlignment = horizontalAlignment
+            self.rowAlignment = rowAlignment
+        }
+
+        private(set) var height: CGFloat = 0
+        private(set) var items: [Item] = []
+        private var totalItemWidth: CGFloat = 0
+
+        private var itemCount: CGFloat {
```
meherkasam-square, line 328: Nit: Feels iffy to have a count be a float. Can we instead directly pull `items.count` and convert to `CGFloat` as needed?

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+/// ┌─────────────────────────────────────┐
+/// │ ┌─────────┐┌─────────┐┌────────────┐│
+/// │ │    1    ││    2    ││     3      ││
+/// │ └─────────┘└─────────┘└────────────┘│
+/// │ ┌───────┐┌─────┐┌───────────┐       │
+/// │ │   4   ││  5  ││     6     │       │
+/// │ └───────┘└─────┘└───────────┘       │
+/// └─────────────────────────────────────┘
+/// ```
+public struct Flow: Element {
```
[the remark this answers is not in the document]
meherkasam-square: Was wondering the same myself if we ever needed to build a vertical flow (similar to Google Image search). We could use the term `Waterfall` for that one ([example](https://github.com/sgr-ksmt/WaterfallLayout)) and stick with `Flow` for the horizontal one.

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+        /// Shorter items are stretched to fill the height of the tallest item.
+        case fill
+
+        /// Shorter items are aligned to the top of the row.
+        case top
+
+        /// Shorter items are vertically aligned within the row.
+        case center
+
+        /// Shoter items are aligned to the bottom of the row.
```
meherkasam-square, line 114:
```suggestion
        /// Shorter items are aligned to the bottom of the row.
```

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+                CGSize(
+                    width: max(current.maxX, reduced.width),
+                    height: max(reduced.height, current.maxY)
+                )
+            }
+        }
+
+        // MARK: Legacy Layout
+
+        func measure(
```
meherkasam-square, line 254: Nit: Consider splitting up the different layout functions into extensions on `Layout`?

BlueprintUI/Sources/Layout/Flow.swift
```diff
@@ -0,0 +1,390 @@
...
+    }
+}
+
+extension Flow {
+
+    /// How to horizontally align items in the row when there is extra space.
+    public enum HorizontalAlignment: CaseIterable {
+
+        /// Items are aligned with the leading edge.
+        case leading
```
meherkasam-square, line 93: Would we also want to support `justify` horizontal alignment?

BlueprintUI/Tests/FlowTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+        let multipleTooWide = Flow(rowSpacing: 5) {
+            ConstrainedSize(width: 110, height: 5)
+            ConstrainedSize(width: 120, height: 5).flowChild(key: "aKey")
+            ConstrainedSize(width: 130, height: 5)
+        }
+
+        XCTAssertEqual(
+            multipleTooWide.content.measure(in: .init(width: 100)),
+            CGSize(width: 130, height: 25)
```
meherkasam-square, line 65: I guess this unit test is related to my comment [here](https://github.com/square/Blueprint/pull/489/files#r1578492578). I'm not sure I fully understand this. It seems like it would cause the container to fit the widest element? If so, wouldn't it violate the size constraint that it's being provided? I'm thinking of scenarios where a pill containing text can grow larger than a developer might expect because of localisation and/or dynamic type.

BlueprintUI/Sources/Layout/Flow.swift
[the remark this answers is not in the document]
kyleve: I'd figure if we ever want this, we could add a `direction` parameter, like we have for `EqualStack` – thoughts?

BlueprintUI/Sources/Layout/Flow.swift
kyleve: - [ ] Check this

BlueprintUI/Sources/Layout/Flow.swift
kyleve: In the name of YAGNI, I decided to leave it off until we have a use case for it – I haven't seen any designs asking for this yet, so, didn't add it!

BlueprintUI/Sources/Layout/Flow.swift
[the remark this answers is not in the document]
meherkasam-square: The implementations of the two orientations and the exposed APIs would be different enough that it might not make a lot of sense to have one struct be used for both? Don't have a strong opinion on this 🤷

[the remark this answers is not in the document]
robmaceachern: Yeah, it seems like working with the configuration properties like `horizontalAlignment`, `horizontalSpacing` etc would be confusing to work with in a vertical `direction`.

If we're thinking this would evolve into supporting vertical direction then it might make sense to use more generic property names, like `UICollectionViewFlowLayout` does (`minimumLineSpacing`, `minimumInteritemSpacing`, etc).

BlueprintUI/Sources/Layout/Flow.swift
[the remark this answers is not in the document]
kyleve: I'm going to leave it off

BlueprintUI/Sources/Layout/Flow.swift
kyleve: Sure!

- [ ] Remove

BlueprintUI/Sources/Layout/Flow.swift
kyleve: - [ ] Fix

BlueprintUI/Tests/FlowTests.swift
kyleve: I flip-flopped on this one – this mirrors behavior of `Column`, which doesn't constrain the measurement, _except_ in `.fill` alignment mode. Eg, this passes:

```swift
XCTAssertEqual(
    Column {
        ConstrainedSize(width: 110, height: 5)
    }.content.measure(in: .init(width: 100)),
    CGSize(width: 110, height: 5)
)
```

But this does not (the measurement in this case is 100):

```swift
XCTAssertEqual(
    Column(alignment: .fill) {
        ConstrainedSize(width: 110, height: 5)
    }.content.measure(in: .init(width: 100)),
    CGSize(width: 110, height: 5)
)
```

I'm good to convert this to constrain the contained children, since yeah, it is a bit odd for things to overflow the container and basically never what someone wants.

- [ ] Convert

meherkasam-square: Even in the Column example, is the expansion the desired behaviour? If a label is longer than anticipated in the non-fill case, would it truncate, go into the next line (assuming `numLines` > 1), or cause the column to expand?

kyleve: This is done

BlueprintUI/Sources/Layout/Flow.swift
kyleve: Nope!

PR #491, opened 2024-04-26
Add `accessibilityElements` to `accessibilityContainer` API [UI-6096]

nsillik:


BlueprintUICommonControls/Sources/AccessibilityContainer.swift
kyleve, line 1: This seems reasonable, but a couple questions:

1) When we're using this in practice, what are the a11y elements here exactly? Are they instances of `UIView`? Instances of `UIAccessibilityElement`? Something else?

2) Are these elements that are already in the view / element hierarchy, or are they additions?

meherkasam-square (approved):

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
nsillik: I'm gonna put this on hold for a minute, i _think_ there might be a cleaner way?

PR #493, opened 2024-05-02
[WIP DNR] Support overriding a11y label on AttributedLabel

kyleve:


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -252,6 +256,7 @@ extension AttributedLabel {
             }
 
             isAccessibilityElement = model.isAccessibilityElement
+            accessibilityLabel = model.accessibilityLabel
```
RoyalPineapple, line 259:
```suggestion
if let label = model.accessibilityLabel {
            accessibilityLabel = label
}
```

PR #496, opened 2024-06-03
Updates Box contentView.layer.cornerMask if it doesn't match the cornerStyle

jhatin94:


kyleve (approved):
I was going to try and nerd snipe you into adding some snapshot tests for this, but I can take an item to do these during my oncall this week as well

PR #501, opened 2024-06-13
Fix string index out of bounds crash.

RoyalPineapple:
Well, this is always embarrassing.

`string.startIndex...string.endIndex` is out of bounds, which seems counter to my understanding of either ranges or string indices but I'm not sure which . 

Regardless, we need a half open range here.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -546,7 +546,7 @@ extension AttributedLabel {
 
         }
 
-        private func accessibilityLabel(with links: [Link], in string: String, linkAccessibilityLabel: String?) -> String {
+        internal func accessibilityLabel(with links: [Link], in string: String, linkAccessibilityLabel: String?) -> String {
             // When reading an attributed string that contains the `.link` attribute VoiceOver will announce "link" when it encounters the applied range. This is important because it informs the user about the context and position of the linked text within the greater string. This can be partocularly important when a string contains multiple links with the same linked text but different link destinations.
 
             // UILabel is extremely insistant about how the `.link` attribute should be styled going so far as to apply its own preferences above any other provided attributes. In order to allow custom link styling we replace any instances of the `.link` attribute with a `labelLink.` attribute (see `NSAttributedString.normalizingForView(with:)`. This allows us to track the location of links while still providing our own custom styling. Unfortunately this means that voiceover doesnt recognize our links as links and consequently they are not announced to the user.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
kyleve, line 1: Tests?

RoyalPineapple: yep

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -357,6 +357,70 @@ class AttributedLabelTests: XCTestCase {
         compareSnapshot(of: element)
     }
 
+    func test_linkAccessibility() {
+        let labelview = AttributedLabel.LabelView()
+
+        do {
+            // Test that link insertion happy path works
```
kyleve, line 364: Can we add a test for a link at the very end of the string too?

kyleve (approved):

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
RoyalPineapple: done

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -565,17 +565,32 @@ extension AttributedLabel {
             // • Square brackets aren't often used in prose, unlike parenthesis. They're unlikely to be confused with the actual content.
             // • They look like markdown.
 
-            let insertionString = "[\(localizedLinkString)] "
+            let insertionString = "[\(localizedLinkString)]"
             // Insert from the end of the string to keep indices stable.
             let reversed = links.sorted { $0.range.location > $1.range.location }
             for link in reversed {
-                let insertionPoint = label.index(label.startIndex, offsetBy: link.range.location + link.range.length)
+                // Extract substring from NSString to allign with NSRange provided by the link.
```
kyleve, line 572: `allign` -> `align`

BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -357,6 +357,109 @@ class AttributedLabelTests: XCTestCase {
...
+                    in: string as String,
+                    linkAccessibilityLabel: "."
+                )
+                XCTAssertEqual(accessibilityLabel, string)
+            }
+        }
+
+        do {
+            // Test with emoji
+            let string = "🇺🇸🇨🇦🇯🇵🇫🇷"
```
kyleve, line 425: These are great, thanks for adding them!

BlueprintUICommonControls/Sources/AttributedLabel.swift
RoyalPineapple: done

robmaceachern:
> `string.startIndex...string.endIndex` is out of bounds, which seems counter to my understanding of either ranges or string indices but I'm not sure which

Ah yeah it's the `endIndex` bit.

> [endIndex](https://developer.apple.com/documentation/swift/string/endindex)
> A string’s “past the end” position—that is, the position one greater than the last valid subscript argument.

Guess they should have called it `pastTheEndIndex` 😛

PR #502, opened 2024-06-17
Replace newlines in a11y labels with spaces

nsillik:
The built in accessibility labels do some mangling of newlines to remove them. Our implementation in `0e9041b4714cac99baa69f4b1b484503868311ba` caused regressions in snapshot and KIF testing.

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -559,7 +559,7 @@ extension AttributedLabel {
                   !links.isEmpty
             else {
                 // We need to replace all newlines with " "
-                return string.components(separatedBy: .newlines).filter { !$0.isEmpty }.joined(separator: " ")
+                return string.removingNewlines
             }
             var label = string
             // Wrap the word in [brackets] to indicate that it is a tag distinct from the content string. This is transparent to voiceover but should be helpful when the accessibility label is printed e.g. in the accessibility inspector.
@@ -601,7 +601,7 @@ extension AttributedLabel {
             }
 
             // We need to replace all newlines with " "
-            return label.components(separatedBy: .newlines).filter { !$0.isEmpty }.joined(separator: " ")
+            return label.removingNewlines
         }
 
         func applyLinkColors(activeLinks: [Link] = []) {
@@ -855,3 +855,8 @@ extension NSTextCheckingResult.CheckingType {
     }
 }
 
+extension String {
+    var removingNewlines: String {
+        components(separatedBy: .newlines).filter { !$0.isEmpty }.joined(separator: " ")
+    }
+}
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -595,7 +595,9 @@ extension AttributedLabel {
                 }
                 label.insert(contentsOf: insertionString, at: insertionPoint)
             }
-            return label
+
+            // We need to replace all newlines with \n
+            return label.components(separatedBy: .newlines).filter { !$0.isEmpty }.joined(separator: " ")
```
kyleve, line 600: Tests please!

RoyalPineapple: To be explicit, theres no prohibition against new lines in accessibility labels. This is a style choice we've made.

RoyalPineapple (approved):

RoyalPineapple:
Thanks for fixing the regression.

BlueprintUICommonControls/Sources/AttributedLabel.swift
nsillik: the tests were a good idea! you'll never believe this: they caught a bug (the change up above that i missed)

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -556,7 +556,11 @@ extension AttributedLabel {
             // Insert the word "link" after each link in the label. This mirrors the VoiceOver behavior when encountering a `.link` attribute.
 
             guard let localizedLinkString = linkAccessibilityLabel,
-                  !links.isEmpty else { return string }
+                  !links.isEmpty
+            else {
+                // We need to replace all newlines with " "
+                return string.components(separatedBy: .newlines).filter { !$0.isEmpty }.joined(separator: " ")
```
RoyalPineapple, line 562: This makes me question where this was happening before...

kyleve: Probably the default value coming from UIKit?

nsillik: Yup, that's what i think too

RoyalPineapple: yea, i tested it and it seems to be happening inside UILabel and only occurs when you don't provide an explicit label.

BlueprintUICommonControls/Sources/AttributedLabel.swift
kyleve: Can we move the line into a private category on string in this file now that we have two copies?

watt (approved):

PR #504, opened 2024-06-19
Label and AttributedLabel now support accessibilityValue

kyleve:
We support hint, but not value. So, adding value too!


meherkasam-square (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -249,13 +255,14 @@ extension AttributedLabel {
 
             if !isMeasuring {
```
kyleve, line 256: I nested these all under `!isMeasuring`, since none of them matter for measuring AFAICT.

PR #508, opened 2024-08-07
Unnest Box.CornerStyle

g-mark:
Since `Box.CornerStyle` is used for other things besides specifying the corners of a `Box`, this PR moves the `CornerStyle` type to the root of the framework.

To maintain compatibility, a typealias is added to `Box`, so consumers can continue to use `Box.CornerRadius`.


CornerStyle.swift
kyleve, line 1: Xcode put it in the wrong spot in the directory structure

g-mark: I hate when that happens

BlueprintUICommonControls/Sources/CornerStyle.swift
kyleve, line 1: I could also see this going in `BlueprintUI` as well, but up to you!

kyleve (approved):

BlueprintUICommonControls/Sources/CornerStyle.swift
g-mark: I suppose the UIBezierPath extension would also move to `BlueprintUI` in that case?

kyleve: I think so, but probably not a big deal to just leave em both here as well!

g-mark: Ok, I'll leave them both where they are. It makes sense to me for them to be colocated, and although I don't think there are any weird side effects of moving them into BlueprintUI (since BlueprintUICommonControls imports it), it's probably best to play it safe for now.

kyleve: SG!

kyleve (approved):

PR #511, opened 2024-09-17
Add passThroughTouches to BlueprintView, PassthroughView

kyleve:
Without this, Blueprint will eat any touches that are meant for views behind it – in particular relevant if you're layering them in the z-index and the one on "top" is largely visually transparent.

image

https://github.com/user-attachments/assets/43c10ec2-0c9b-4b66-ba4e-1134fab32223

### DIFF
--- a/BlueprintUI/Sources/Internal/PassthroughView.swift
+++ b/BlueprintUI/Sources/Internal/PassthroughView.swift
@@ -16,7 +16,6 @@ import UIKit
 
         if passThroughTouches {
             return result == self ? nil : result
-
         } else {
             return result
         }
--- a/BlueprintUI/Sources/Internal/PassthroughView.swift
+++ b/BlueprintUI/Sources/Internal/PassthroughView.swift
@@ -8,10 +8,17 @@ import UIKit
         CATransformLayer.self
     }
 
-    /// Ignore any touches on this view and (pass through) by returning nil if the
-    /// default `hitTest` implementation returns this view.
+    public var passThroughTouches: Bool = true
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Moved `CornerStyle` out of the `Box` namespace, and is now a root type in `BlueprintUICommonControls`. `Box.CornerStyle` is still available as a typealias.
+- `BlueprintView` will now pass through touches to views lower in the view hierarchy if `passThroughTouches` is true.
```
watt, line 19: this is a non-breaking addition, right?

watt (approved):

CHANGELOG.md
kyleve: Ahh yep, good catch. I'd originally just made this the only behavior, but figured that was a bad idea.

- [ ] Fix Changelog

BlueprintUI/Sources/Internal/PassthroughView.swift
```diff
@@ -8,10 +8,17 @@ import UIKit
         CATransformLayer.self
     }
 
-    /// Ignore any touches on this view and (pass through) by returning nil if the
-    /// default `hitTest` implementation returns this view.
+    public var passThroughTouches: Bool = true
```
kyleve, line 11: This should be `false`

kyleve: Wait nope, this isn't `BlueprintView`

ldstreet (approved):

PR #515, opened 2024-10-18
Make `GeometryProxy.environment` public

g-mark:
Makes the `environment` property of `GeometryReader` public.

The immediate need for this is to enable the use of `displayScale` values inside of `GeometryReader` calculations.


kyleve (approved):
I don't think this hurts at all, but you could also wrap the returned element in an `EnvironmentReader { env in ... }` to get to the env as well.

g-mark:
> I don't think this hurts at all, but you could also wrap the returned element in an `EnvironmentReader { env in ... }` to get to the env as well.

Yes, that's right, although it would result in nested blocks to get at both of them. @watt and I did discuss this, and he recommended making the `environment` public.

watt (approved):
- [ ] update the changelog

PR #523, opened 2024-11-16
build: set up tuist

watt:
This PR replaces the CocoaPods-based project generation with [Tuist](https://tuist.io/). In a follow-up PR I'll remove CocoaPods entirely.

Setting up takes a few steps, first to get [Mise](https://mise.jdx.dev/) and then to get Tuist from that.

```sh
# install mise
brew install mise
# add mise activation line to your zshrc
echo 'eval "$(mise activate zsh)"' >> ~/.zshrc
# load mise into your shell
source ~/.zshrc
# install dependencies
mise install

# only necessary for first setup or after changing dependencies
tuist install --path SampleApp
# generates and opens the Xcode project
tuist generate --path SampleApp
```

The bump to `swiftformat` caused some formatting changes, so I recommend reviewing by commit.


README.md
```diff
@@ -113,6 +113,26 @@ target MyTarget do
...
 
+## Local Development
+
+This project uses [Mise](https://mise.jdx.dev/) and [Tuist](https://tuist.io/) to generate a project for local development. Follow the steps below for the recommended setup for zsh.
+
+```sh
+# install mise
+brew install mise
+# add mise activation line to your zshrc
+echo 'eval "$(mise activate zsh)"' >> ~/.zshrc
```
robmaceachern, line 124: It might be nice if we update this command to add it only if it is not already there. If people naively go through the readme of Blueprint/Workflow/Market/etc they'll end up with 3+ entries in their .zshrc.

SampleApp/Project.swift
robmaceachern, line 1: Can we add the changelog as an additional file? ([how we do it in Market](https://github.com/squareup/market/blob/8b0fa88fb3971f670d02963ebf46c9c6d3f3bafd/ios/Project.swift#L530-L534))

.mise.toml
```diff
@@ -0,0 +1,7 @@
+[tools]
+tuist = "4.23.0"
+swiftformat = "0.54.2"
```
robmaceachern, line 3: Nit: It'd be much easier to review this (and compare with other Mise migrations) if we stuck to the existing Swiftformat version and reverted the formatting changes.

Or if we are going to bump it let's go to the latest.

watt: Fair, although I think at the time this _was_ the latest. I've rolled it back to the previously pinned version.

SampleApp/Project.swift
watt: done

README.md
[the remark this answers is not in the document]
watt: Added `mise trust` to the setup instructions.

PR #530, opened 2024-12-04
build: remove CocoaPods

### LINKED ISSUE #215: Add README FAQ about SwiftUI
Blueprint and SwiftUI both tackle a similar kind of problem, so for developers outside Square, it would be helpful to understand how they differ – for example, Blueprint's support for deploying to older versions of iOS.

watt:
Removes podspecs and most traces of CocoaPods. Blueprint will be distributed exclusively via SwiftPM.

Fixes #215


Gemfile.lock
```diff
@@ -125,8 +124,6 @@ PLATFORMS
   x86_64-linux
 
 DEPENDENCIES
-  cocoapods (~> 1.11)
-  cocoapods-generate (~> 1.0)
   jazzy (~> 0.13)
```
watt, line 127: CocoaPods still appears in Gemfile.lock because it's a transitive dependency of jazzy. Might explore switching to docc at some point.

README.md
```diff
@@ -1,23 +1,39 @@
...
+
+### How does it interact with UIKit?
+
+Blueprint is not a replacement for UIKit! From the beginning, Blueprint has been designed as a compliment to all of the powerful tools that come with the platform. You can use Blueprint to manage the display of a single view controller, or of a single view representing a small part of the screen. Likewise, it's straightforward to host standard views and controls *within* a blueprint hierarchy, always leaving you with an escape hatch.
+
+### How does it interact with SwiftUI?
+
+They serve similar purposes, and SwiftUI has heavily influenced Blueprint's API. However, Blueprint predates SwiftUI, and works a bit differently. You can host a `BlueprintView` in SwiftUI to embed Blueprint within SwiftUI, and you can also use the provided `ElementView` to create Xcode previews of Blueprint. Because SwiftUI is hosted by a UIViewController, and Blueprint is hosted in a UIView, we don't provide a way to embed SwiftUI within Blueprint.
```
watt, line 32: I cleaned this file up a lot, but it's mostly moving and deleting. This paragraph is new.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- CocoaPods podspecs removed. Blueprint will only be vended via Swift Package Manager.
```
robmaceachern, line 16: Do you know if there's anything we could/should do with on cocoapods trunk to signal that these pods are inactive/archived? Probably not critical, just curious if there's some final cleanup we should be doing here.

Gemfile
```diff
@@ -1,8 +1,6 @@
 source 'https://rubygems.org'
 
 # Read the current Ruby version from the .ruby-version file
-ruby Bundler.read_file(Bundler.root.join('.ruby-version')).strip
```
robmaceachern, line 4: Why did this change? iirc the `Bundler.root` avoids issues when `bundle install` is invoked in a subdirectory.

If we're going to change it, it'd probably be best to use the more modern `ruby file: ".ruby-version"` [flavor](https://bundler.io/guides/gemfile_ruby.html).

Scripts/release.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+echo "TODO: Update this script to use the new release process."
+echo "Use the manual steps in RELEASING.md for now."
+exit 0
```
robmaceachern, line 5: 😢

robmaceachern (approved):

CHANGELOG.md
watt: I don't think so.

Gemfile
watt: This fixes incompatibility with the VS Code Ruby extension, and still works in subdirectories. I can change to the new thing but I think it'll require a bump to bundler.

Scripts/release.sh
watt: sorry!

Gemfile.lock
```diff
@@ -52,7 +52,6 @@ GEM
       typhoeus (~> 1.0)
     cocoapods-deintegrate (1.0.5)
     cocoapods-downloader (2.1)
-    cocoapods-generate (1.4.1)
     cocoapods-plugins (1.0.0)
       nap
     cocoapods-search (1.0.1)
```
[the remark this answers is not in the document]
watt: CocoaPods still appears in Gemfile.lock because it's a transitive dependency of jazzy

PR #531, opened 2024-12-13
feat: Move view cache to Blueprint hierarchy

robmaceachern:
- **Move view caching to the environment**
- **Update to make cache generic**
- **more**
- **Renamed `MeasurementCache` to `TypeKeyedCache`**
- **Fix leak of AccessibilityViewController demo**
- **Rename `BlueprintView.viewMeasurer` to `viewCache`**
- **Fix leak in TextLinkViewController demo**


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -525,6 +527,13 @@ public final class BlueprintView: UIView {
 
         var environment = inherited.merged(prioritizing: environment)
 
+        /// We're a root `BlueprintView`, so pass in the view measurer for our downstream
+        /// dependencies if we didn't inherit a measurement cache.
+
+        if environment.inheritedViewCache == nil {
+            environment.viewCache = viewCache
+        }
```
robmaceachern, line 535: I'm still a little nervous about cascading view caches through the environment. Won't it still leave open the possibility that nested blueprint view elements leak stuff through long(er)-lived caches?

BlueprintUI/Sources/Environment/Keys/ViewCacheKey.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+extension Environment {
+    private enum ViewCacheKey: EnvironmentKey {
+        static let defaultValue: TypeKeyedCache? = nil
+    }
+
+    private static let fallback = TypeKeyedCache()
```
robmaceachern, line 8: I'm thinking we shouldn't have a static fallback.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
kyleve: It's definitely possible, but largely the root blueprint view lives on the screen level.

BlueprintUI/Sources/Environment/Keys/ViewCacheKey.swift
kyleve: Yeah if we can get away with it. I haven't done an integration in POS yet, and I believe there is older code that just out of band measures elements with an empty environment, so we may need it for now.

BlueprintUI/Sources/Measuring/TypeKeyedCache.swift
```diff
@@ -0,0 +1,34 @@
...
+    // Intentionally internal. Not intended for public instantiation.
+    init() {}
+
+    // Returns the cached value for the Value type if it exists. If it doesn't
+    // exist, it creates a new instance of Value, caches it, and returns it.
+    //
+    // - Parameter create: A closure that creates a new instance of Value. It
+    // is only invoked when a cached value does not exist. The created value is
+    // cached for future usage.
+    public func value<Value: AnyObject>(_ create: () -> Value) -> Value {
```
kyleve, line 17: FWIW, I went with the "access" pattern with an access closure so people aren't tempted to keep a reference to the cached measurement view, kind of like the swift pointer APIs where the pointer shouldn't be stored outside the access closure.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -128,16 +128,22 @@ public struct AttributedLabel: Element, Hashable {
     /// You can check if this value should be false via `NSAttributedString.needsNormalizingForView(...)`
     public var needsTextNormalization: Bool = true
 
-    private static let prototypeLabel = LabelView()
```
robmaceachern, line 131: @watt suggested we consider leaving this one as-is, at least for the initial iteration since its:
- low likelihood of retaining something it shouldn't be (and would be flushed out by other measurements)
- on a relatively hot code path

PR #532, opened 2024-12-16
feat: Introduce CacheClearer to force-clear static caches

robmaceachern:


kyleve (approved):
Do we need to clear out / replace any `static` prototype views as well? (Eg, in `AttributedLabel`?)

BlueprintUI/Sources/Element/UIViewElement.swift
robmaceachern, line 1: (commenting here for threading purposes)

@kyleve:
> Do we need to clear out / replace any static prototype views as well? (Eg, in AttributedLabel?)

Great question! When I was discussing this with @watt last week we both agreed that the AttributedLabel prototype view is [probably ok as-is](https://github.com/square/Blueprint/pull/531/files#r1884636704). It should be updated frequently enough in regular usage that even in the unlikely case it is momentarily retaining something somehow (I guess through a a11y closure?) it'll get freed up pretty quickly.

Happy to deal with it too, though! If it was in the same module I'd just reset the prototype but I can't reach it from BlueprintUI.

If we end up exposing a wrapper in Market for these cache clearing things we could always just measure an empty label there too.

johnnewman-square (approved):
Nice! Providing clients the ability to manually clear the cache feels like a good stop-gap that we can build on top of going forward.

PR #537, opened 2025-01-22
AttributedLabel Accessibility Links

RoyalPineapple:
This PR migrates `AttributedLabel`'s link rotor from a stateful to functional pattern, eliminating the need for two variables and preventing voiceover from getting out of sync. 

I also fixed a bug in `links(at location: CGPoint)` that could cause a crash if the `AttributedString` lacked a specified `NSTextAlignment`.

### DIFF
--- a/BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
+++ b/BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
@@ -10,19 +10,12 @@ extension BidirectionalCollection where Element: Equatable, Element: NSObjectPro
             return UIAccessibilityCustomRotorItemResult(targetElement: first, targetRange: nil)
         }
         let newIndex = (predicate.searchDirection == .next ? index(after: currentIndex) : index(before: currentIndex))
-            .clamped(min: startIndex, max: index(before: endIndex))
+        guard newIndex >= startIndex, newIndex < endIndex else { return nil }
         return .init(targetElement: self[newIndex], targetRange: nil)
     }
 
     /// Returns a UIAccessibilityCustomRotor with the provided system type which cycles through the contained elements.
     public func accessibilityRotor(systemType type: UIAccessibilityCustomRotor.SystemRotorType) -> UIAccessibilityCustomRotor {
         UIAccessibilityCustomRotor(systemType: type) { itemSearch($0) }
-    }
-}
-
-// MARK: Misc. Internal
-extension Comparable {
-    fileprivate func clamped(min: Self, max: Self) -> Self {
-        Swift.max(Swift.min(self, max), min)
     }
 }
--- a/BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
+++ b/BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
@@ -1,0 +1,28 @@
+import UIKit
+
+extension BidirectionalCollection where Element: Equatable, Element: NSObjectProtocol {
+    private func itemSearch(_ predicate: UIAccessibilityCustomRotorSearchPredicate) -> UIAccessibilityCustomRotorItemResult? {
+        guard let first else { return nil }
+        guard let currentItem = predicate.currentItem.targetElement as? Element,
+              let currentIndex = firstIndex(of: currentItem),
+              predicate.searchDirection == .previous || predicate.searchDirection == .next
+        else {
+            return UIAccessibilityCustomRotorItemResult(targetElement: first, targetRange: nil)
+        }
+        let newIndex = (predicate.searchDirection == .next ? index(after: currentIndex) : index(before: currentIndex))
+            .clamped(min: startIndex, max: index(before: endIndex))
+        return .init(targetElement: self[newIndex], targetRange: nil)
+    }
+
+    /// Returns a UIAccessibilityCustomRotor with the provided system type which cycles through the contained elements.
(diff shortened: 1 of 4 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -389,7 +364,7 @@ extension AttributedLabel {
                 }
 
                 assert(
-                    alignments.count == 1,
```
RoyalPineapple, line 392: This was causing a crash in the demo app. 
Someone went through the trouble of writing the nil case below, it'd be a shame never to use it.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -210,35 +210,11 @@ extension AttributedLabel {
             }
         }
 
-        /// The view needs to keep track of the current index for the accessibility rotor.
-        private var accessibilityLinkIndex = -1
-
-        /// These elements need to be retained by the view, and cannot be created inside the
-        /// `accessibilityCustomRotors` getter.
```
RoyalPineapple, line 217: This concern about the elements being retained is still valid. We cannot pass back unretained elements from the rotor search block. 

The trick here is  the  elements are  now retained in an array which is captured by the rotor's `itemSearchBlock`. 

Once VoiceOver releases the rotor these elements all get deallocated and then will be recreated the next time VoiceOver requests the rotor.

BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
```diff
@@ -0,0 +1,21 @@
+import UIKit
+
+extension BidirectionalCollection where Element: Equatable, Element: NSObjectProtocol {
+    private func itemSearch(_ predicate: UIAccessibilityCustomRotorSearchPredicate) -> UIAccessibilityCustomRotorItemResult? {
+        guard let first else { return nil }
+        guard let currentItem = predicate.currentItem.targetElement as? Element,
```
kyleve, line 6: Thinking outloud, im surprised to see both the generic constraint and this check; are both required?

BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
```diff
@@ -0,0 +1,21 @@
+import UIKit
+
+extension BidirectionalCollection where Element: Equatable, Element: NSObjectProtocol {
```
g-mark, line 3: Nit: I think you can also combine the constrains
```suggestion
extension BidirectionalCollection where Element: Equatable & NSObjectProtocol {
```

BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
```diff
@@ -0,0 +1,21 @@
...
+        else {
+            return UIAccessibilityCustomRotorItemResult(targetElement: first, targetRange: nil)
+        }
+        let newIndex = (predicate.searchDirection == .next ? index(after: currentIndex) : index(before: currentIndex))
+        guard newIndex >= startIndex, newIndex < endIndex else { return nil }
+        return .init(targetElement: self[newIndex], targetRange: nil)
+    }
+
+    /// Returns a UIAccessibilityCustomRotor with the provided system type which cycles through the contained elements.
+    public func accessibilityRotor(systemType type: UIAccessibilityCustomRotor.SystemRotorType) -> UIAccessibilityCustomRotor {
```
g-mark, line 18: Nit: Put the public func at the top of the file, and the private extension after?

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -210,35 +210,11 @@ extension AttributedLabel {
...
-        private var accessibilityLinks: [LinkAccessibilityElement] = []
-
         override var accessibilityCustomRotors: [UIAccessibilityCustomRotor]? {
             set { fatalError() }
```
g-mark, line 214: Nit (and outside the scope of the PR): would it be better have this be a no-op instead of a crasher?
```suggestion
            set { _ = newValue }
```

g-mark (approved):
Nice!

BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
RoyalPineapple: The generic constraint is on the array, but the dynamic check is being run on the item that we've been passed in the predicate, which is only guaranteed to be an NSObjectProtocol, not necessarily Equatable. 

We could get away with just checking for Equatable, but I thought it would be nice to ensure that all the types match by using Element instead.

BlueprintUI/Sources/Extensions/BidirectionalCollection+Blueprint.swift
RoyalPineapple: Ah, of course!
 I  initially wrote `Element: Equatable, NSObjectProtocol ` and this was Xcode's suggested fix.

BlueprintUICommonControls/Sources/AttributedLabel.swift
RoyalPineapple: The pattern in the rest of Blueprint is to `fatalError` for "read only" property overrides, so I'd like to match that. But I should add a more descriptive message.

g-mark: That makes sense. I generally try to avoid fatalErrors at all cost, but if this is the pattern for this repo, it makes sense to stick to it. At some point it might be worth discussing making a systemic change - but certainly not now!

kyleve (approved):

PR #545, opened 2025-04-21
feat: Add `UserInterfaceOverridingElement `

nsillik:
Adds a `UserInterfaceStyleOverridingElement` and a `overrideUserInterfaceStyle()` extension on `Element` so that subtrees of views can have their `userInterfaceStyle` (i.e. light/dark) overridden a la SwiftUI's `colorScheme` environment variable.

### DIFF
--- a/BlueprintUICommonControls/Sources/Image.swift
+++ b/BlueprintUICommonControls/Sources/Image.swift
@@ -15,9 +15,6 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
-    /// Whether to override the user interface style, `nil` is the system style.
-    public var overrideUserInterfaceStyle: UIUserInterfaceStyle?
-
     /// iOS 14 added support for Image Descriptions using VoiceOver. This is not always appropriate.
     /// Set this to `true` to prevent VoiceOver from describing the displayed image.
     public var blockAccessibilityDescription: Bool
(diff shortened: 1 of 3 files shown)


BlueprintUICommonControls/Sources/Image.swift
```diff
@@ -15,6 +15,9 @@ public struct Image: Element {
     /// not precisely match the size that the element is assigned.
     public var contentMode: ContentMode
 
+    /// Whether to override the user interface style, `nil` is the system style.
+    public var overrideUserInterfaceStyle: UIUserInterfaceStyle?
```
kyleve, line 19: Interesting! I wonder if exposing this as a view-backed element that sets `overrideUserInterfaceStyle` on itself makes more sense so any views downstream get this? Eg:

```swift
myElement
 .overrideUserInterfaceStyle(.light)
```

nsillik: @kyleve you mean like this? (check the latest changes)

BlueprintUICommonControls/Sources/UserInterfaceOverridingElement.swift
```diff
@@ -0,0 +1,73 @@
...
+///
+/// Example:
+/// ```swift
+/// let content = Label(text: "Hello, World!")
+/// let forcedLightMode = UserInterfaceOverridingElement(
+///     userInterfaceStyle: .light,
+///     wrapping: content
+/// )
+/// ```
+public struct UserInterfaceOverridingElement: Element {
```
kyleve, line 18: Yes, I think this is much more extensible! Couple thoughts:

1) We dropped the "style" from the name here, how about `OverrideUserInterfaceStyle`?

2) Can we make the element in the init be a trailing closure so you can do

```swift
OverrideUserInterfaceStyle(.light) {
  ... some element ...
}
```
?

BlueprintUICommonControls/Sources/UserInterfaceOverridingElement.swift
```diff
@@ -0,0 +1,72 @@
...
+/// An element that allows overriding the user interface style (light/dark mode) for its wrapped content.
+///
+/// Use this element to force a specific appearance for a portion of your UI, regardless of the system-wide
+/// settings. This can be useful when you need to ensure certain UI elements maintain a consistent appearance
+/// across different user interface styles.
+///
+/// Example:
+/// ```swift
+/// let content = Label(text: "Hello, World!")
+/// let forcedLightMode = UserInterfaceOverridingElement(
```
kyleve, line 13: This is out of date

PR #556, opened 2025-05-24
feat!: Remove legacy layout

watt:
Removes the legacy layout mode! May be easier to review by commit.

Following this change, layouts that define a single trait type will be required to explicitly conform to `SingleTraitLayout` to use the existing API for accessing traits (previously this was implicit via `LegacyLayout`).

I've updated layouts to remove legacy implementations, and in some cases refactored to remove the "shim" that allowed a shared implementation. But not all of them; in particular I left the Stack layout as-is, so that we can test if there's any performance impact of eagerly reading traits into an array or if the array allocation itself outweighs that.

Alongside this removal I also removed the `fillParent` mode of `ConstrainedAspectRatio`, which was already deprecated because it's not useful under caffeinated layout.

BREAKING CHANGE: removes `LayoutMode.legacy`, `LegacyLayout`, `ConstrainedAspectRatio.ContentMode.fillParent`, and requires layouts that define a trait type to add the `SingleTraitLayout` conformance explicitly.


BlueprintUI/Sources/Internal/SignpostToken.swift
```diff
@@ -0,0 +1,2 @@
+/// A token reference type that can be used to group associated signpost logs using `OSSignpostID`.
+final class SignpostToken {}
```
watt, line 2: moved from CacheFactory.swift

PR #557, opened 2025-06-06
Fix an issue where resizing scroll views could result in their content offset being changed erroneously

kyleve:
### Bug
https://github.com/user-attachments/assets/7f93f74d-0b54-4131-ac52-8970ec623452

### Notes
🐛  If you have a partial modal (or any other modal that resizes), and you resize the modal while the keyboard is focused and scrolled away from the top, to be taller (eg by adding content), the scroll view within would jump up towards the top.

Westin and I had dug in a while ago and thought it was just a UIKit bug, but it turns out it's an order of operations thing: You need to apply the new `contentSize` to the scroll view before you make its frame taller, otherwise UIKit's automatic `contentOffset` adjustment kicks you back towards the top.

To fix this, I've added a `applyBeforeLayout` addition to `ViewDescription.Config`, which will be applied before the `LayoutAttributes` (except on initial view allocation – we still apply the attributes first so the view's properties are sensical).


BlueprintUI/Sources/ViewDescription/ViewDescription.swift
```diff
@@ -154,9 +169,13 @@ extension ViewDescription {
         /// The default value instantiates the view using `init(frame:)`.
         public var builder: () -> View
 
-        /// An array of update closures.
+        /// An array of update closures that are applied **after** the `LayoutAttributes` is applied to the view.
         public var updates: [Update]
 
+        /// An array of update closures that are applied **before** the `LayoutAttributes` is applied to the view after
```
kyleve, line 175: Grammar is weird here

kyleve:
POS integration passed, surprisingly! https://github.com/squareup/ios-register/pull/127207

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -729,6 +730,8 @@ extension BlueprintView {
                     UIView.performWithoutAnimation {
                         controller = NativeViewController(node: child)
                         child.layoutAttributes.apply(to: controller.view)
+                        // So the view has a reasonable size during creation/allocation, do this afterwards.
+                        child.viewDescription.applyBeforeLayout(to: controller.view)
```
kyleve, line 734: @watt Thoughts on this? This feels a bit better – `init` is all in a `performWithoutAnimation`, and I'd rather the view have at least a sane frame/bounds before we call the apply method here since some views don't generally work well if their frame/bounds has zero size.

PR #558, opened 2025-06-12
Added support for tabbing through links in `AttributedLabel`

meherkasam-square:


BlueprintUICommonControls/Tests/Sources/AttributedLabelTests.swift
```diff
@@ -484,27 +484,6 @@ class AttributedLabelTests: XCTestCase {
         }
     }
 
-
-    func test_linkAccessibility_Rotors() {
-        let labelView = AttributedLabel.LabelView()
```
meherkasam-square, line 489: Removed this due to having refactored how the rotor is built. Additionally, the logic being tested by this deleted test is also being tested by the first part of the below `test_linkAccessibility_Rotors_update` test.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -663,41 +653,108 @@ extension AttributedLabel {
...
+            // In cases where a link overflows from one line to the next,
+            // we will get back more than one rect. If we had used the boundingRect
+            // function instead, it will only return us a bounding rect that can
+            // fully encompass all the sub-rects - this is undesired behavior as
+            // it will suppress focusable items within the label that have an
+            // overlap with the larger bounding box.
+            // Since we only select the first rectangle, if there is indeed an overflow,
+            // only the portion of the link in the preceding line would be highlighted by
+            // VoiceOver or Full Keyboard Access. While this is not ideal behavior, it is still
+            // preferred over the former. The correct fix for this would be to return bezier paths
```
RoyalPineapple, line 678: Can we return bezier paths? at least for the accessibilityPath?

RoyalPineapple (approved):
Looks good, nice work

meherkasam-square:
> Nice! It's cool to see an example of `parentFocusEnvironment` in action. The bounding rect for multiline links is looking good, too.
> 
> It's not a blocker, but I wonder if there are some additional unit tests that we could add for AttributedLabel's `linkElements` or `focusItems(in:)`?

Thanks for the callout, @johnnewman-square. I added a couple of unit tests for the focus items changes.

BlueprintUICommonControls/Sources/AttributedLabel.swift
meherkasam-square: I was thinking I will scope that work into a separate PR so that the main focus of this change is on the keyboard accessibility.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -663,41 +653,108 @@ extension AttributedLabel {
...
+                    forGlyphRange: glyphRange,
+                    withinSelectedGlyphRange: NSMakeRange(NSNotFound, 0),
+                    in: textContainer
+                ) { rect, _ in
+                    glyphRects.append(rect)
+                }
+                return glyphRects
+            }()
+
+            return rects.first ?? .zero
```
soroushsq, line 692: Not sure if this will compile but I think using the second block parameter would make this code simpler
```suggestion
            var rect: CGRect?
            layoutManager.enumerateEnclosingRects(
                forGlyphRange: glyphRange,
                withinSelectedGlyphRange: NSMakeRange(NSNotFound, 0),
                in: textContainer
                ) { rect, stop in
                    glyphRects.append(rect)
                    stop = true
            }
            return rect ?? .zero
```

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -663,41 +653,108 @@ extension AttributedLabel {
     struct Link: Equatable, Hashable {
         var url: URL
         var range: NSRange
+        weak var container: LabelView?
+
+        var boundingRect: CGRect {
```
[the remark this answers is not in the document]
soroushsq: Fun!

soroushsq (approved):

PR #561, opened 2025-06-17
Fix multi-line link accessibility paths

meherkasam-square:
This change is a fast-follow to [this](https://github.com/square/Blueprint/pull/558) PR. The previous PR improved how multi-line links were exposed to VoiceOver and Full Keyboard Access. However, it was flawed in that it would only expose the first fragment. This PR fixes that by encompassing the entire link inside a single bezier path.

### Demo

(For the "Before" view, refer to the description in [this](https://github.com/square/Blueprint/pull/558) PR.)

#### VoiceOver

https://github.com/user-attachments/assets/c8bcff72-5096-4331-9be8-12cdcba3c54d

#### Full keyboard access

https://github.com/user-attachments/assets/a1e3e6f8-3f85-46b5-b486-d8176eafe34d

### DIFF
--- a/BlueprintUICommonControls/Sources/AttributedLabel.swift
+++ b/BlueprintUICommonControls/Sources/AttributedLabel.swift
@@ -654,28 +654,35 @@ extension AttributedLabel {
         var range: NSRange
         weak var container: LabelView?
 
-        var boundingRect: CGRect {
+        struct BoundingShape: Equatable, Hashable {
+            // In case of a link that does not span across multiple lines, the `firstRect` property
+            // is simply the bounding rect for that link. `path` is `nil` in those cases.
+            // Otherwise for links that go longer than one line, `firstRect` encompasses the
+            // first portion of the link up to the end its line.
+            // Using that rectangle as the frame ensures that the item is correctly ordered
+            // by the focus system along with the other links.
+            var firstRect: CGRect
+            var path: UIBezierPath?
+
+            func hash(into hasher: inout Hasher) {
+                hasher.combine(path)
+                hasher.combine(firstRect.origin.x)
+                hasher.combine(firstRect.origin.y)
+                hasher.combine(firstRect.size.width)
+                hasher.combine(firstRect.size.height)
+            }
+        }
+
+        var boundingShape: BoundingShape {
             guard let container = container,
                   let textStorage = container.makeTextStorage(),
                   let layoutManager = textStorage.layoutManagers.first,
                   let textContainer = layoutManager.textContainers.first
             else {
-                return .zero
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -654,28 +654,39 @@ extension AttributedLabel {
         var range: NSRange
         weak var container: LabelView?
 
-        var boundingRect: CGRect {
+        enum BoundingShape: Equatable, Hashable {
```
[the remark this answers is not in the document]
meherkasam-square: We still need to provide a rect because of the `frame` property on `UIFocusItem`. I like Soroush's suggestion, so I updated my PR with it. PTAL now.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -654,28 +654,35 @@ extension AttributedLabel {
...
+            // by the focus system along with the other links.
+            var firstRect: CGRect
+            var path: UIBezierPath?
+
+            func hash(into hasher: inout Hasher) {
+                hasher.combine(path)
+                hasher.combine(firstRect.origin.x)
+                hasher.combine(firstRect.origin.y)
+                hasher.combine(firstRect.size.width)
+                hasher.combine(firstRect.size.height)
```
soroushsq, line 672: So annoying that CGRect isn't Hashable, what's that about

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -688,8 +695,29 @@ extension AttributedLabel {
...
+                let cgPath = CGMutablePath()
+                let cornerRadius: CGFloat = 4.0
+
+                for rect in rects {
+                    cgPath.addRoundedRect(in: rect, cornerWidth: cornerRadius, cornerHeight: cornerRadius)
+                }
+                return .init(cgPath: cgPath)
+            }()
+
+            return .init(firstRect: rects[0], path: path)
```
soroushsq, line 718: Is there any chance `rect` could be empty?

soroushsq (approved):
Looks ok to me, would love for @RoyalPineapple to take a look as well

RoyalPineapple (approved):
This all makes sense, one minor stylistic suggestion. 
I cant wait unitl we can expose these in snapshots tests.

BlueprintUICommonControls/Sources/AttributedLabel.swift
soroushsq: Oh yes that's good

PR #565, opened 2025-07-03
Improve perf of AttributedLabel link focus changes and suppress console messages

meherkasam-square:
This PR mainly fixes two separate things:
1. Performance issues with the focus changes in `AttributedLabel` - Due to the `frame` property being requested each time the focus is changed, the code was doing an expensive recalculation of the bounding shape. This change introduces caching for the bounding shape and only ensures that the cache is invalidated when the label is updated or when `layoutSubviews` is called on the label view. The third commit contains the test code that I used for measuring the performance of this code. In the "before" version, the frame rate had dropped to ~25 FPS. In the "after" version, the frame rate is ~110 FPS.
2. Despite using the [API](https://developer.apple.com/documentation/uikit/uifocusitemcontainer/focusitems(in:)) in the prescribed manner, there would be repeated console output of the form `_TtCV25BlueprintUICommonControls15AttributedLabel9LabelView implements focusItemsInRect: - caching for linear focus movement is limited as long as this view is on screen.` I am able to replicate this on both iPadOS and tvOS (the platform that Apple had [originally](https://devstreaming-cdn.apple.com/videos/wwdc/2018/208piymryv9im6/208/208_whats_new_in_tvos_12.pdf) demonstrated this API on). This change implemented the `_supportsInvalidatingFocusCache()` similar to how Apple does it in a number of their system classes such as `UITableView` and `UICollectionView`, amongst others.

image


BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -644,8 +649,38 @@ extension AttributedLabel {
             trackingLinks = nil
             applyLinkColors()
         }
+
+        override func layoutSubviews() {
+            super.layoutSubviews()
+
+            invalidateLinkBoundingShapeCaches()
+        }
```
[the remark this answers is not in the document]
meherkasam-square: I tried a few scenarios like rotating screen orientation, moving app to background/foreground, and adding new elements into the screen on the fly. They all hit the breakpoint for the invalidation function. Happy to test any other scenarios you can think of.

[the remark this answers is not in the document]
soroushsq: Can you check changing the accessibility scale as well? That will probably also trigger it, but I just want to cover our bases.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -644,8 +690,46 @@ extension AttributedLabel {
...
+
+        override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+            super.traitCollectionDidChange(previousTraitCollection)
+
+            if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
```
soroushsq, line 703: Nice

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -222,6 +225,47 @@ extension AttributedLabel {
...
+            super.init(frame: frame)
+
+            if #available(iOS 17.0, *) {
+                registerForTraitChanges([UITraitPreferredContentSizeCategory.self]) { (
+                    self: Self,
+                    previousTraitCollection: UITraitCollection
+                ) in
+
+                    self.invalidateLinkBoundingShapeCaches()
+                }
```
soroushsq, line 238: Nit, but can we rename `self` to `view` or `label`. Easer to see that it's not a retain cycle that way

soroushsq (approved):

BlueprintUICommonControls/Sources/AttributedLabel.swift
soroushsq: Oh lol

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -644,8 +689,38 @@ extension AttributedLabel {
...
+extension AttributedLabel.LabelView {
+    // Without this, we were seeing console messages like the following:
+    // "LabelView implements focusItemsInRect: - caching for linear focus movement is limited as long as this view is on screen."
+    // It's unclear as to why they are appearing despite using the API in the intended manner.
+    // To suppress the messages, we implemented this function much like Apple did with `UITableView`,
+    // `UICollectionView`, etc.
+    @objc private class func _supportsInvalidatingFocusCache() -> Bool {
+        true
+    }
 }
```
johnnewman-square, line 724: Nice detective work on tracking down these logs. If there isn't a noticeable performance difference when we don't implement this function, my one reservation against implementing it is that it's unclear if returning `true` may have an unintended side effect. But if we'd prefer to reduce the logging, no objections to shipping this and letting it run through QA.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -775,13 +854,7 @@ extension AttributedLabel {
 
         override var accessibilityPath: UIBezierPath? {
             set { assertionFailure("cannot set accessibilityPath") }
-            get {
-                if let path = link.boundingShape.path, let container = link.container {
```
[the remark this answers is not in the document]
johnnewman-square: I like the idea of calculating this on the fly by making a copy of the original path, where the original path is in view coordinates.

If this label were embedded inside a scroll view, it doesn't look like scrolling would trigger a cache invalidation because `layoutSubviews` and `update(model:text:environment:isMeasuring)` wouldn't be called. By calculating the screen path on the fly, we'd guarantee that it's not using stale coordinates after scrolling.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -674,6 +749,10 @@ extension AttributedLabel {
         }
 
         var boundingShape: BoundingShape {
+            container?.boundingShape(for: self) ?? calculateBoundingShape()
```
RoyalPineapple, line 752: I think we should cache in the local coordinate space and we should apply the `UIAccessibility.convertToScreenCoordinates` here rather than below. Similar to `UIAccessibilityElement.accessibilityFrameInContainerSpace`.

BlueprintUICommonControls/Sources/AttributedLabel.swift
```diff
@@ -775,13 +854,7 @@ extension AttributedLabel {
...
         override var accessibilityPath: UIBezierPath? {
             set { assertionFailure("cannot set accessibilityPath") }
-            get {
-                if let path = link.boundingShape.path, let container = link.container {
-                    return UIAccessibility.convertToScreenCoordinates(path, in: container)
-                }
-
-                return nil
-            }
+            get { link.boundingShape.path }
```
RoyalPineapple, line 857: I think we should cache in the local coordinate space and instead apply the `UIAccessibility.convertToScreenCoordinates` here. Similar to `UIAccessibilityEleement.AccessibilityFrameInContainerSpace`.

PR #566, opened 2025-07-10
Fix AccessibilityContainer behaviour when it contains a UICollectionView or UITableView

meherkasam-square:


BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -106,25 +107,57 @@ extension Element {
...
+            }
             set { fatalError("This property is not settable") }
         }
     }
 }
 
 extension UIView {
-    func recursiveAccessibleSubviews() -> [Any] {
-        subviews.flatMap { subview -> [Any] in
+    func accessibilityElements(
```
RoyalPineapple, line 124: We're now duplicating this same logic in three places here, within market, and in the snapshots. 

It would be nice to have this (or an equivalent) as one of the canonical functions available in our forthcoming a11y toolbox. @soroushsq

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -106,25 +107,68 @@ extension Element {
...
+
+            // UITableView is a similar special case as UICollectionView
+            if let tableView = subview as? UITableView {
+                return [tableView]
+            }
+
+            if let accessibilityElements = subview.accessibilityElements {
+                return accessibilityElements.compactMap { element -> [NSObject] in
+                    guard let elementAsView = element as? UIView else {
+                        // If it's not a UIView, assume that there's nothing further to traverse.
```
RoyalPineapple, line 155: why is this assumed?
Why cant a non view contain more elements?

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
soroushsq: In that case, do we expect that Blueprint will depend on the accessibility toolbox?

soroushsq (approved):

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
meherkasam-square: Discussed with Alex and updated the logic the account for a nested NSObject-based hierarchy.

BlueprintUICommonControls/Sources/AccessibilityContainer.swift
meherkasam-square: Given the more recent development of modifying the sort function to be more specific to how Blueprint behaves, we might need to maintain at two separate versions of `frameSort`.

PR #568, opened 2025-07-16
Add equivalency protocols and implementations (1/3)

cae7ffb:
1/3 for caching changes.
Big-picture PR here: https://github.com/square/Blueprint/pull/587
Design doc: https://docs.google.com/document/d/1CLhFbzZGbEKgvZTUwLuGbFY-rfZNJVTi6f3PkoZT9qI/edit?usp=sharing

This PR adds a `ContextualEquivalency` protocols and conformances. This is basically like a lightweight `Equatable`, which adds the ability to interrogate objects with the notion of a context.


BlueprintUI/Sources/Environment/Environment.swift
```diff
@@ -71,8 +81,58 @@ public struct Environment {
...
+    public func isEquivalent(to other: Self?, in context: EquivalencyContext) -> Bool {
+        guard let other else { return false }
+        let keys = Set(values.keys).union(other.values.keys)
+        for key in keys {
+            guard key.isEquivalent(self[key], other[key], context) else {
+                return false
+            }
+        }
+        return true
+    }
```
johnnewman-square, line 99: Small question: how many keys are generally stored in `values`? Maybe we could introduce a short circuit by comparing the count of each array before creating the sets and iterating over keys. Just a thought!

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,88 @@ public protocol EnvironmentKey {
...
     static var defaultValue: Self.Value { get }
+
+
+    /// Compares two environment values without direct conformance of the values.
+    /// - Parameters:
+    ///   - lhs: The left hand side value being compared.
+    ///   - rhs: The right hand side value being compared.
+    ///   - context: The context to evaluate the equivalency.
+    /// - Returns: Whether or not the two values are equivalent in the specified context.
+    static func isEquivalent(lhs: Value, rhs: Value, in context: EquivalencyContext) -> Bool
```
watt, line 37: If this is required I think there should be a default implementation returning false.

- [ ] default impl on isEquivalent

watt (approved):
A few minor things, left you tasks for the important bits.

BlueprintUI/Sources/Internal/Equivalency.swift
```diff
@@ -0,0 +1,72 @@
...
+    case all
+
+    // More fine-grained contexts:
+
+    /// The two values are equivalent in all aspects that would affect the size of the element.
+    /// - Warning:Non-obvious things may affect element-sizing – for example, setting a time zone may seem like something that would only affect date calculations, but can result in different text being displayed, and therefore affect sizing. Consider carefully whether you are truly affecting sizing or not.
+    case elementSizing
+}
+
+public protocol ContextuallyEquivalent {
```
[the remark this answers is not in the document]
maxg-square: It's slightly more broad than element-specific (includes env values too) and I think comparable has a connotation of being sortable given the stl usage of it – maybe something like `CacheEquivalent`? Or could just go the other way and be less concretely tied to that specific method it requires and be something more tied to purpose like `CrossLayoutCacheable`?

BlueprintUI/Sources/Environment/EnvironmentKey.swift
maxg-square: Discused offline earlier but feel pretty strongly we shouldn't have that – there's default implementations when values are equatable and also convenience implementations like this that are one-liners https://github.com/square/Blueprint/pull/568/files#diff-e664d2d763ebe3ca35c985ddc39c338eff03d90a5dd5f6d90091b7612d02ce4aR55 but IMO new environmentkeys should be forced to consider this.

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,88 @@ public protocol EnvironmentKey {
...
     static var defaultValue: Self.Value { get }
+
+
+    /// Compares two environment values without direct conformance of the values.
+    /// - Parameters:
+    ///   - lhs: The left hand side value being compared.
+    ///   - rhs: The right hand side value being compared.
+    ///   - context: The context to evaluate the equivalency.
+    /// - Returns: Whether or not the two values are equivalent in the specified context.
+    static func isEquivalent(lhs: Value, rhs: Value, in context: CrossLayoutCacheableContext) -> Bool
```
watt: I still feel like it's a mistake to not have a default implementation here, but it's not a blocker.

PR #569, opened 2025-07-16
CacheStorage (2/3)

cae7ffb:
2/3 for caching changes.
Big-picture PR here: https://github.com/square/Blueprint/pull/587
Design doc: https://docs.google.com/document/d/1CLhFbzZGbEKgvZTUwLuGbFY-rfZNJVTi6f3PkoZT9qI/edit?usp=sharing

- Adds `HostingViewContext`, an object which hangs off of `Environment`. `HostingViewContext` has a similar API to `Environment`, with consumers declaring their types and default values by declaring a `CrossLayoutCacheKey`.
- Adds internal values to `Environment` which do not participate in `ContextuallyEquivalent` checks.
- Adds `EnvironmentAccessList`, which is a snapshot of an `Environment` that holds a subset of its storage, and is used by `ValidatingCache`.
- Adds `ValidatingCache`, which allows a key to be set with a value that's associated with a validation function.
- Adds `EnvironmentValidatingCache` and `EnvironmentAndValueValidatingCache`, which are simplified interfaces to `ValidatingCache` which provide additional comparison of `Environment` contexts.
- Logging and tests.

### DIFF
--- a/BlueprintUI/Sources/Environment/Environment.swift
+++ b/BlueprintUI/Sources/Environment/Environment.swift
@@ -188,7 +188,7 @@ extension Environment: CrossLayoutCacheable {
 
 extension HostingViewContext {
 
-    fileprivate struct EnvironmentFingerprintCache {
+    fileprivate final class EnvironmentFingerprintCache {
 
         struct Key: Hashable {
             let lhs: CacheComparisonFingerprint.Value


BlueprintUI/Sources/CrossLayoutCaching/Cache/HostingViewContext.swift
```diff
@@ -0,0 +1,85 @@
...
+            "HostingViewContext"
+        }
+        return "\(debugName): \(storage.count) entries"
+    }
+
+}
+
+extension Environment {
+
+    struct HostingViewContextKey: InternalEnvironmentKey {
```
watt: Plucking this question out of an old thread:

Can we avoid the need for `InternalEnvironmentKey` by making this key use a reference comparison of `HostingViewContext`, or hardcoded `true`?

watt (approved):
Names and overall structure look great, thanks! Just one open question.

PR #574, opened 2025-07-17
[DNR] All the caching

kyleve:


BlueprintUI/Sources/Element/MeasurableStorage.swift
```diff
@@ -17,7 +28,19 @@ extension MeasurableStorage: CaffeinatedContentStorage {
         environment: Environment,
         node: LayoutTreeNode
     ) -> CGSize {
-        measurer(proposal, environment)
+        guard environment.layoutMode.options.measureableStorageCache, let validationKey else {
+            return measurer(proposal, environment)
+        }
+
+        let key = MeasurableSizeKey(path: node.path, max: proposal.maximum)
+        return environment.cacheStorage.measurableStorageCache.retrieveOrCreate(
```
kyleve, line 36: Thought: Could we stick making the key and looking up the value on `LayoutTreeNode` (in a lazy way) so that we don't need to do this key creation and lookup dance outside of a single spot / in every `Storage` that will need it?

BlueprintUI/Sources/Element/MeasurableStorage.swift
```diff
@@ -17,7 +28,19 @@ extension MeasurableStorage: CaffeinatedContentStorage {
         environment: Environment,
         node: LayoutTreeNode
     ) -> CGSize {
-        measurer(proposal, environment)
+        guard environment.layoutMode.options.measureableStorageCache, let validationKey else {
+            return measurer(proposal, environment)
+        }
+
+        let key = MeasurableSizeKey(path: node.path, max: proposal.maximum)
```
kyleve, line 35: Unrelated to your PR, but TIL we store the path as a string, I thought we'd keep the fully qualified `ElementPath` around

BlueprintUI/Sources/Environment/Cache/CacheKey.swift
```diff
@@ -0,0 +1,27 @@
...
+///
+/// You can write a small extension on `CacheStorage` to make it easier to use your key.
+///
+///     extension CacheStorage {
+///         var widgetCounts: [WidgetID: Int] {
+///             get { self[WidgetCountsKey.self] }
+///             set { self[WidgetCountsKey.self] = newValue }
+///         }
+///     }
+public protocol CacheKey {
```
kyleve, line 24: Idle thought: I think we can nest protocols in other types now, maybe this could be `CacheStorage.Key`?

BlueprintUI/Sources/Environment/EnvironmentKey.swift
```diff
@@ -26,4 +26,88 @@ public protocol EnvironmentKey {
...
+    /// - Parameters:
+    ///   - lhs: The left hand side value being compared.
+    ///   - rhs: The right hand side value being compared.
+    ///   - context: The context to evaluate the equivalency.
+    /// - Returns: Whether or not the two values are equivalent in the specified context.
+    static func isEquivalent(lhs: Value, rhs: Value, in context: EquivalencyContext) -> Bool
+
+}
+
+extension EnvironmentKey where Value: Equatable {
```
kyleve, line 41: Would be good to add something for NSObject and AnyObject too?

BlueprintUI/Sources/Internal/Equivalency.swift
```diff
@@ -0,0 +1,72 @@
+import Foundation
+
+// A context in which to evaluate whether or not two values are equivalent.
+public enum EquivalencyContext: Hashable, Sendable, CaseIterable {
```
kyleve, line 4: I am wondering if we'd need this at all if we returned to tracking / comparing only the environment values that are read by the element in question, vs. comparing everything

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -40,6 +40,8 @@ public final class BlueprintView: UIView {
 
     private var sizesThatFit: [SizeConstraint: CGSize] = [:]
 
+    private var cacheStorage = Environment.CacheStorageEnvironmentKey.defaultValue
```
kyleve, line 43: I think we need to do some sort of garbage collection on this, right? Eg, evict paths that are no longer present after a layout update?

PR #577, opened 2025-07-22
AccessibilityContainer allow custom elements

RoyalPineapple:
now allowing child elements to be specified explicitly, allowing for elements that are not backed by uiviews

more for discussion really, is this something we should support?

### DIFF
--- a/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
+++ b/BlueprintUICommonControls/Sources/AccessibilityContainer.swift
@@ -30,6 +30,9 @@ public struct AccessibilityContainer: Element {
     /// An optional `accessibilityValue` to give the container. Defaults to `nil`.
     public var value: String?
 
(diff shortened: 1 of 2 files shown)


BlueprintUICommonControls/Sources/AccessibilityContainer.swift
```diff
@@ -30,6 +30,9 @@ public struct AccessibilityContainer: Element {
     /// An optional `accessibilityValue` to give the container. Defaults to `nil`.
     public var value: String?
 
+    /// An optional array of accessibility elements, to override the wrapped children.
+    public var elements: [NSObjectProtocol]?
```
RoyalPineapple, line 34: theres an argument to be made for this to be `[Any]?`

soroushsq:
> more for discussion really, is this something we should support?

Let's discuss during our 1:1, curious to know what use cases you see for this

PR #581, opened 2025-07-28
Support accessibility large content viewer

meherkasam-square:


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -333,6 +336,12 @@ public final class BlueprintView: UIView {
 
     public override func didMoveToWindow() {
         super.didMoveToWindow()
+        if supportsAccessibilityLargeContentViewer {
```
RoyalPineapple, line 339: Should we also check if the window is nil here? this method gets called when we remove a view from the hierarchy.

https://developer.apple.com/documentation/uikit/uiview/didmovetowindow()

meherkasam-square:
- [ ] Update CHANGELOG

meherkasam-square:
- [ ] Namespace Accessibility (from @RoyalPineapple)

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -333,6 +336,12 @@ public final class BlueprintView: UIView {
 
     public override func didMoveToWindow() {
         super.didMoveToWindow()
+        if window != nil, supportsAccessibilityLargeContentViewer {
+            let largeContentViewerInteraction = UILargeContentViewerInteraction(delegate: self)
+            addInteraction(largeContentViewerInteraction)
```
kyleve, line 341: Do we need this on the Blueprint view at all? I'd expect / think it might be simpler to do this on the backing view / element itself – or can there only be one of these per screen or view hierarchy or something?

(We currently put this on the tab view in Market, vs some higher level container: https://github.com/squareup/market/blob/main/ios/MarketUI/Sources/TabBar/MarketTabBar.swift#L265)

meherkasam-square: The interaction could conceivably be placed even the level of each individual view. The tradeoff is that we lose the update behaviour when you finger-swipe between elements that don't share an interaction. We could manually add the interaction at the items grid level to address the issue for the Checkout applet. But we would still lose out on potential finger-swipe-update behaviour for components that would logically belong together but don't share an interaction.

RoyalPineapple: If thats the only reason we're applying this at the blueprint level, should we have a largeInteractionContentContainer that maintains the interaction and can be inserted in the blueprint hierarchy more strategically?

BlueprintUICommonControls/Sources/AccessibilityLargeContentViewer.swift
```diff
@@ -0,0 +1,236 @@
...
+
+            switch largeContentViewerConfiguration.display {
+            case .title(let title, let image):
+                largeContentTitle = title
+                largeContentImage = image
+            case .image(let image):
+                largeContentTitle = nil
+                largeContentImage = image
+            case .none:
+                largeContentTitle = nil
```
RoyalPineapple, line 167: this case with nil nil. shouldn't be allowed. it brings up a grey square that blocks the UI with no content. its a bad experience.

BlueprintUICommonControls/Sources/AccessibilityLargeContentViewer.swift
```diff
@@ -0,0 +1,236 @@
...
+
+        public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription? {
+            LargeContentViewerInteractionContainerView.describe { _ in }
+        }
+    }
+}
+
+extension Accessibility {
+
+    private final class LargeContentViewerInteractionContainerView: UIView, UILargeContentViewerInteractionDelegate {
```
RoyalPineapple, line 210: Thanks, i like this a lot better than adding cruft to BlueprintView

BlueprintUICommonControls/Sources/AccessibilityLargeContentViewer.swift
meherkasam-square: Okay, I moved the line from the top of the function into each case. I also added a button with a `.none` `display` to the demo screen.

BlueprintUICommonControls/Sources/AccessibilityLargeContentViewer.swift
```diff
@@ -0,0 +1,242 @@
...
+}
+
+extension Accessibility {
+    public struct LargeContentViewerConfiguration {
+
+        public enum Display: Equatable {
+            case title(String, UIImage?)
+            case image(UIImage)
+            case none
+        }
```
soroushsq, line 85: Not a big enough deal to go back and revise, but if you end up making some changes here, I'm not sure the enum provides you any value over a struct with two optional members

meherkasam-square: I wanted to capture the specific edge case of a `nil` title and `nil` image - by default when you supply those params, a blank HUD appears. By having a `.none` case, we instead force the caller to be explicit that they want to hide the HUD and remove the blank HUD edge case completely.

soroushsq: In that case you can just check that both values are nil (or put a little helper boolean property on the type)

meherkasam-square: That was an option I had considered, but I preferred this approach as the `.none` was more explicit about not displaying the HUD whereas the two nils approach was more implicit. Don't have a strong opinion here though.

BlueprintUICommonControls/Sources/AccessibilityLargeContentViewer.swift
```diff
@@ -0,0 +1,242 @@
...
+// MARK: - Large content viewer container
+
+extension Element {
+
+    /// Adds a large content viewer interaction container to the element.
+    /// This is used to wrap elements that need to be able to show a large content viewer.
+    /// Use this in conjunction with accessibilityShowsLargeContentViewer() on elements that need to show a large content viewer.
+    /// Elements that are wrapped in this container will be able to show a large content viewer and allow a user to swipe through them with one finger
+    /// and have the HUD update in real time.
+    public func accessibilityLargeContentViewerInteractionContainer() -> Element {
```
kyleve, line 193: Was just thinking about this this morning, and came to suggest this same thing, nice!

PR #614, opened 2026-05-05
Declare tappable views as accessibility buttons

RoyalPineapple:
🤖 Summary:
- Mark `TappableView` as an accessibility element with the `.button` trait.
- Add tests for the backing view trait and for button trait inheritance through `accessibilityCombine`.

Verification:
- `xcodebuild build -scheme BlueprintUICommonControls -destination 'platform=iOS Simulator,id=EF731F66-6F9B-40A7-9F42-0C6CF3CBEFAB'`


BlueprintUICommonControls/Sources/Tappable.swift
```diff
@@ -9,7 +9,7 @@ public struct Tappable: Element {
     public var onTap: () -> Void
 
     public init(onTap: @escaping () -> Void, wrapping element: Element) {
-        wrappedElement = element
+        wrappedElement = element.accessibilityTraits(add: [.button])
```
robmaceachern, line 12: Should we be doing this conditionally? For example if the element already has a `.link` trait does it make sense to also apply `.button`?

robmaceachern, line 12: Does the `onTap` need to be wired up to actually be activated by voiceover?

RoyalPineapple: I considered doing it explicitly but we don't need to.  

Voiceover first calls -accessibilityActivate which we could override, but if that returns `false` (or falls through to the base implementation in NSObject which does) it instead issues a tap, which in this case works just fine.

RoyalPineapple: My initial reaction is no, but .link is one good counter example.  Let me think on that as i work on the integration testing.

johnnewman-square (approved):

robmaceachern (approved):

BlueprintUICommonControls/Sources/Tappable.swift
RoyalPineapple: So we don't have a simple way to query what the element's traits are at any given time.  We generally only use links in AttributedLabel, which strips the trait anyway. I think the risk of duplication (which, while confusing is not completely misleading or anything) is acceptable.

PR #618, opened 2026-06-08
fix: defer reentrant BlueprintView updates

robmaceachern:
## Summary

- defer BlueprintView hierarchy invalidations that occur while Blueprint is already updating its view tree
- coalesce the deferred retry so UIKit safe-area/window callbacks cannot synchronously re-enter `updateViewHierarchyIfNeeded()`
- add a regression test that mutates `additionalSafeAreaInsets` during backing-view attach/layout, matching the production crash shape
- add the missing `BlueprintUICommonControls` dependency for the package test target so the focused package test builds

## Context

This came from an ios-register 7.10 crash investigation where `BlueprintView.safeAreaInsetsDidChange()` re-entered `BlueprintView.updateViewHierarchyIfNeeded()` during a Market keyboard/header layout update. A MarketUI harness reproduced the exact precondition crash locally before this Blueprint patch; with this patch copied into Market's Blueprint checkout, the same harness passes.

## Testing

- `xcodebuild test -scheme BlueprintUI-Package -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=26.5' -only-testing:BlueprintUITests/BlueprintViewTests/test_safeAreaChangeDuringUpdateDefersNestedUpdate -quiet`
- Market validation with this Blueprint source copied into Market's local checkout: `xcodebuild test -workspace ios/MarketDevelopment.xcworkspace -scheme UnitTests -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=26.5' -only-testing:MarketUI-Tests/MarketKeyboardReaderReentrantLayoutTests/test_safeAreaChangeDuringKeyboardReaderUpdateDoesNotReenterBlueprint -quiet`

### DIFF
--- a/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
+++ b/BlueprintUI/Sources/BlueprintView/BlueprintView.swift
@@ -384,6 +384,14 @@ public final class BlueprintView: UIView {
         }
     }
 
+    /// Schedules a coalesced view hierarchy update on a following main run loop turn.
+    ///
+    /// Used when an invalidation arrives while an update is already in progress — for example,
+    /// UIKit keyboard and safe-area callbacks can fire synchronously mid-update and force a
+    /// layout pass. Re-arming with `setNeedsLayout()` immediately would let a forcing
+    /// `layoutIfNeeded()` re-dirty this view while the in-flight update still holds
+    /// `isInsideUpdate`, preventing that forced layout from ever terminating, so the re-arm
+    /// is deferred until the current update has unwound.
     @MainActor
     private func scheduleDeferredViewHierarchyUpdate() {
         guard hasScheduledDeferredViewHierarchyUpdate == false else { return }
@@ -415,6 +423,9 @@ public final class BlueprintView: UIView {
         guard needsViewHierarchyUpdate || bounds != lastViewHierarchyUpdateBounds else { return }
 
         if isInsideUpdate {
+            // A reentrant update was triggered from within the in-flight update, e.g. by a
+            // safe-area or keyboard change synchronously forcing a layout pass. Defer it so
+            // the current update can finish; the deferred pass picks up the latest state.
             setNeedsViewHierarchyUpdate()
             return
         }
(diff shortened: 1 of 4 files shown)


BlueprintUI/Sources/BlueprintView/BlueprintView.swift
```diff
@@ -386,10 +422,13 @@ public final class BlueprintView: UIView {
...
-            !isInsideUpdate,
-            "Reentrant updates are not supported in BlueprintView. Ensure that view events from within the hierarchy are not synchronously triggering additional updates."
-        )
+        if isInsideUpdate {
+            // A reentrant update was triggered from within the in-flight update, e.g. by a
+            // safe-area or keyboard change synchronously forcing a layout pass. Defer it so
+            // the current update can finish; the deferred pass picks up the latest state.
+            setNeedsViewHierarchyUpdate()
+            return
+        }
```
johnnewman-square, line 469: Logging this case could be useful, if we'd like to track this going forward.

johnnewman-square (approved):
This is the best fix proposal so far. This should address all call sites, more than the isolated Market fix.

I'm in favor of trying this out until we find a tighter repro.

BlueprintUI/Sources/BlueprintView/BlueprintView.swift
johnnewman-square: To avoid infinite async layout churn, an agent suggested this:

> I'd suggest a debug-only diagnostic — e.g., count consecutive deferred re-arms with no intervening quiet turn and assertionFailure/os_log past a threshold (say 10). That preserves the developer signal the precondition used to provide while keeping production safe.

I think there's value in logging and in debug asserts for cases like this.

robmaceachern: How does this look? https://github.com/square/Blueprint/pull/618/commits/bd980ae9f52eb9b3c061b2dc94bdcf37bf8dd783

